PR #12, opened 2019-02-10
Add test harness for running all of Julia's tests

timholy:
This adds an enormous test suite for JuliaInterpreter: all of Julia's own tests. It is presumably impractical to run these on Travis, but local devs can do so. The idea here is to use this to see how robust (or not) JuliaInterpreter really is. (If anyone has ideas about how to get them to run on PRs without bogging things down too much, I'm all ears.)

The tests do not pass yet, but this allows people to work towards that goal. Meanwhile I'll get started on an issue that flags failures and some of the reasons why. Help (even just triaging files that pass and those that do not) would be very welcome!

### DIFF
--- a/test/juliatests.jl
+++ b/test/juliatests.jl
@@ -1,0 +1,59 @@
+using JuliaInterpreter
+using Test
+
+if !isdefined(Main, :read_and_parse)
+    include("utils.jl")
+end
+
+const juliadir = dirname(dirname(Sys.BINDIR))
+const testdir = joinpath(juliadir, "test")
+if isdir(testdir)
+    include(joinpath(testdir, "choosetests.jl"))
+else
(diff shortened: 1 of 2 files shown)


vchuravy:
We might be able to run them through Gitlab on our own Hardware and in stages

timholy:
OK. My mission for the morning is to rewrite this to:
- support resource-limiting: e.g., each lowered block gets at most `n` statements that can be executed before bailing (thus preventing, e.g., infinite loops or really slow tests from blocking everything)
- aggregate test results
- write out results in Markdown table format for pasting into #13 as status updates
- parallelize

Would be nice to have an overall summary but then click on individual test files to see which lines pass and which do not. Hopefully once that's done, maybe it will be easier to "parallelize development" and have more folks digging into specific failures.

PR #22, opened 2019-02-13
rethrow interrupt exceptions when testing julia

KristofferC:
Makes it a bit easier to stop the execution when running the julia tests.


KristofferC:
Absolete with #21

Issue #13, opened 2019-02-10
Julia tests don't pass


timholy:
The test script `test/juliatests.jl` can be used to run Julia's own tests (presumably, you need a source build of Julia to do that).

Here's the start of a status list. Check means the test passes, no check means it's failing. Help is very much wanted! Here are good ways to help out:

- [ ] (easy) complete the list below (one line per test)
- (moderate) investigate failures and file an issue with a MWE
- (hard) fix the bugs

Without further ado, here's the list:

- [ ] ambiguous: `rethrow() not allowed outside a catch block`
- [x] string/search
- [ ] string/types: seems to take "forever"

KristofferC:
For + `string/basic` we get stuck at

```
ex = quote
            let b, n
                for T = (UInt8, Int8, UInt16, Int16, UInt32, Int32, UInt64, Int64, UInt128, Int128, BigInt), b = 2:62, _ = 1:10
                    n = if T != BigInt
                            rand(T)
                        else
                            BigInt(rand(Int128))
                        end
                end
            end
        end
end
```

I think we will have some problem finishing tests that do this type of combinatorial testing.

timholy:
Thanks! Agreed this is an issue. Plans here: https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/12#issuecomment-463122146

timholy:
The numbers seem substantially better on a relatively recent build of Julia's master branch. I found that julia 1.0.x doesn't even finish, at least on an earlier version the script (though I didn't investigate why).

I also ran with a much higher `nstmts` threshold, so this is much more comprehensive than what I first posted. And it wasn't even that slow, maybe 10 minutes on an 8-worker run.

timholy:
Now that #24 is merged, OP updated to provide helpful links about how to debug problems.

KristofferC:
One worker for me seems to get stuck at something and consume a pretty hefty amount of memory 
![mem](https://user-images.githubusercontent.com/1282691/52798795-ef450f00-3078-11e9-9954-7df7da11e6c9.PNG)

KristofferC:
With the stdlibs now running, it seems the REPL tests deadlock. Perhaps this is something to do with the tasks in the REPL tests (which are notorious for deadlocking if something goes wrong). In this case the `nstmts` don't save us so perhaps another option is to give a timer to abort a test if it has been running too long?

KristofferC:
<details>
[...]

PR #24, opened 2019-02-13
Run Julia tests with resource limits, with worker process, and collate the results

timholy:
This makes it much more feasible to run "all" of Julia's tests. The key advance is the ability to run a specified maximum number of statements in the interpreter, and abort early when that number is reached. Choosing a number of statements, rather than execution time, should make it easier to compare results on machines with different performance. On my laptop, the default settings run all of Julia's tests in < 5 minutes. This is *not* a sign that the interpreter is really efficient (it's not), just that it aborts a lot of slow tests.

Sample output is now visible in #13.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ failed.out
 src/builtins.jl
 deps/build.log
 docs/build/
+test/results.md
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -3,6 +3,10 @@
 [[Base64]]
 uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
+[[Distributed]]
+deps = ["Random", "Serialization", "Sockets"]
+uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+
 [[InteractiveUtils]]
 deps = ["Markdown"]
 uuid = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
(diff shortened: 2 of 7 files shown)


KristofferC:
Would bolding the text in the table for those files that somehow fail make sense?  Might make it easier to find them in the table and get an overview of the "density" of failures.

timholy:
In the long run, I agree, but even if we discount the "aborted" column, a sizable majority of tests have some problem that Base itself does not. So most lines would be bolded. (The only tests that may not need attention are those with zeros in the "Fails" and "Errors" columns.)

timholy:
OK to merge? Given that I've advertised #13 I'd like to make this runnable by others.

timholy:
I now have the suspicion that anonymous functions are the overwhelming reason for the tests with `X`s in #13. This is tricky territory; check out the results of `Meta.lower(Main, :(map(x->x^2, [1,2,3])))`. Because you're defining new structs there are things that have to be done at top-level and as a consequence you can get into world-age issues depending on how you handle this. While this PR tries to solve that problem, I suspect it's not really correct yet. I'm not immediately sure about the best fix.

KristofferC:
Feel free to merge this and future PRs at will :)

timholy:
I was just wondering how strongly you felt about the bold text suggestion.

KristofferC:
``` 
weakly [======|                                ] strongly
```

PR #29, opened 2019-02-14
slightly better line numbers

pfitzseb:
This gives "correct" line numbers for method code (at least as correct as Julia allows it).

### DIFF
--- a/test/interpret.jl
+++ b/test/interpret.jl
@@ -196,3 +196,18 @@ JuliaInterpreter.finish_and_return!(JuliaStackFrame[], frame, true)
 val = @interpret(BigInt())
 @test isa(val, BigInt) && val == 0
 @test isa(@interpret(Base.GMP.version()), VersionNumber)
+
+# "correct" line numbers
+defline = @__LINE__() + 1
+function f(x)
+    x = 2x
+    # comment
+    # comment
+    x = 2x
+    # comment
+    return x*x
+end
+frame = JuliaInterpreter.enter_call(f, 3)
+@test JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(1)) == defline + 1
+@test JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(3)) == defline + 4
+@test JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(5)) == defline + 6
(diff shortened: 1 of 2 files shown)


test/interpret.jl
```diff
@@ -156,3 +156,19 @@ ex = quote
...
+function f(x)
+    x = 2x
+    # comment
+    # comment
+    x = 2x
+    # comment
+    return x*x
+end
+frame = JuliaInterpreter.enter_call_expr(:($(f)(3)))
+JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(1)) == defline + 1
```
timholy:
```suggestion
@test JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(1)) == defline + 1
```

test/interpret.jl
```diff
@@ -156,3 +156,19 @@ ex = quote
...
+JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(3)) == defline + 4
```
timholy:
```suggestion
@test JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(3)) == defline + 4
```

test/interpret.jl
```diff
@@ -156,3 +156,19 @@ ex = quote
...
+JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(5)) == defline + 6
```
timholy:
```suggestion
@test JuliaInterpreter.location(frame, JuliaInterpreter.JuliaProgramCounter(5)) == defline + 6
```

test/interpret.jl
```diff
@@ -156,3 +156,19 @@ ex = quote
...
+defline = @__LINE__() + 1
+function f(x)
+    x = 2x
+    # comment
+    # comment
+    x = 2x
+    # comment
+    return x*x
+end
+frame = JuliaInterpreter.enter_call_expr(:($(f)(3)))
```
timholy: Not important to change, but FYI you can also use `enter_call(f, 3)`.

pfitzseb: Neat, changed to that.

test/interpret.jl
pfitzseb: Not entirely sure how I managed to forget the `@test`s there... Fixed now.

timholy: One of my main motivations for writing Revise was to be able to do the following without needing to restart Julia:
- commit the new test(s)
- `git stash` the fix so the code should still be broken
- `include("runtests.jl")` and verify that they fail in the way I expected
- `git stash pop`
- `include("runtests.jl")` and verify that they pass
- amend the commit to include the fix (or just add a new one)

Otherwise I'd be doing stuff like that all the time.

PR #34, opened 2019-02-15
Nicer display of JuliaStackFrames

timholy:
I doubt this should be merged as-is, since it duplicates some functionality of Debugger.jl. More likely we should move some stuff. But some of the functionality didn't seem to be used, like the `undef_callback` arg of `DebuggingFramework.print_var`, so I thought it might be better to discuss.

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -143,6 +143,53 @@ include("localmethtable.jl")
 include("interpret.jl")
 include("builtins.jl")
 
+function Base.show(io::IO, framecode::JuliaFrameCode)
+    show_scope(io, framecode.scope)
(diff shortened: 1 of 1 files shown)


pfitzseb:
One thing I'd really like is to cleanly separate printing and general display logic, e.g. having a `locals` function that returns a vector of `(name, value)` tuples that can then be printed by `print_var`. 
Otherwise we end up having the same code in all frontends (namely [Juno](https://github.com/JunoLab/Atom.jl/blob/6b3e4b658d04ebf27673fdfcc15387dde1985d66/src/debugger/stepper.jl#L224-L239) :smile:).

timholy:
Superseded by #124

PR #47, opened 2019-02-18
Make it easier to create useful bug reports from Julia tests

timholy:
See the new instructions in #13.


timholy:
Doesn't change any code that runs in the "conventional" tests (those run by `runtests`) so I'll merge.

PR #50, opened 2019-02-20
rename location to linenumber

KristofferC:
Calling this function `location` is promising a bit too much.


timholy:
I've already been bit by this once, thanks for fixing this.

And from the Travis run I see you're getting the full dose of joys of cross-package development!

KristofferC:
Yeah, will fix Debugger as well.

PR #54, opened 2019-02-21
add a function to extract local variables from a frame

KristofferC:
This provides a way to extract the local variables from a frame. The output is in a structured format to make it easier for different frontends to present it in their chosen way (ref https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/34#issuecomment-464000838).

Happy with inputs regarding the API chosen. Is a Vector the correct thing to return, perhaps a `Set` is more correct?
Creating a new struct `LocalVariable` has a bit of cognitive overhad over e.g. a tuple but might make it easier to change the behavior in the future.

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -159,6 +159,51 @@ JuliaStackFrame(frame::JuliaStackFrame, pc::JuliaProgramCounter; kwargs...) =
     JuliaStackFrame(frame.code, frame, pc; kwargs...)
 
 """
+`Variable` is a struct representing a variable with an asigned value.
+By calling the function `locals`[@ref] on a `JuliaStackFrame`[@ref] a
+`Vector` of `Variable`'s is returned.
+
+Important fields:
+- `value::Any`: the value of the local variable
+- `name::Symbol`: the name of the variable as given in the source code
+- `isparam::Bool`: if the variable is a type parameter, for example `T` in `f(x::T) where {T} = x` .
+"""
+struct Variable
+    value::Any
+    name::Symbol
+    isparam::Bool
+end
+Base.show(io::IO, var::Variable) = print(io, var.name, " = ", var.value)
+Base.isequal(var1::Variable, var2::Variable) = 
+    var1.value == var2.value && var1.name == var2.name && var1.isparam == var2.isparam
+
+"""
+    local_variables = locals(frame::JuliaStackFrame)::Vector{Variable}
+
+Return the local variables as a vector of `Variable`[@ref].
+"""
+function locals(frame::JuliaStackFrame)
+    vars = Variable[]
+    syms = Set{Symbol}()
+    for i = length(frame.locals):-1:1
+        if !isa(frame.locals[i], Nothing)
+            sym = frame.code.code.slotnames[i]
+            sym in syms && continue
+            push!(vars, Variable(something(frame.locals[i]), sym, false))
+            push!(syms, sym)
+        end
+    end
+    reverse!(vars)
+    if frame.code.scope isa Method
+        syms = sparam_syms(frame.code.scope)
+        for i in 1:length(syms)
+            push!(vars, Variable(frame.sparams[i], syms[i], true))
+        end
+    end
+    return vars
+end
+
+"""
 `framedict[method]` returns the `JuliaFrameCode` for `method`. For `@generated` methods,
 see [`genframedict`](@ref).
 """
--- a/docs/src/dev_reference.md
+++ b/docs/src/dev_reference.md
@@ -73,4 +73,6 @@ JuliaInterpreter.iswrappercall
 JuliaInterpreter.isdocexpr
 JuliaInterpreter.isglobalref
 JuliaInterpreter.statementnumber
+JuliaInterpreter.Variable
+JuliaInterpreter.locals
 ```
--- a/test/interpret.jl
+++ b/test/interpret.jl
@@ -274,3 +274,39 @@ end
 # Some expression can appear nontrivial but lower to nothing
 @test isa(JuliaInterpreter.prepare_thunk(Main, :(@static if ccall(:jl_get_UNAME, Any, ()) == :NoOS 1+1 end)), Nothing)
 @test isa(JuliaInterpreter.prepare_thunk(Main, :(Base.BaseDocs.@kw_str "using")), Nothing)
+
+@testset "locals" begin
+    f_locals(x::Int64, y::T, z::Vararg{Symbol}) where {T} = x
+    frame = JuliaInterpreter.enter_call(f_locals, 1, 2.0, :a, :b)
+    locals = JuliaInterpreter.locals(frame)
+    @test JuliaInterpreter.Variable(1, :x, false) in locals
+    @test JuliaInterpreter.Variable(2.0, :y, false) in locals
+    @test JuliaInterpreter.Variable((:a, :b), :z, false) in locals
+    @test JuliaInterpreter.Variable(Float64, :T, true) in locals
+
+    function f_multi(x)
+        c = x
+        x = 2
+        x = 3
+        x = 4
+        return x
+    end
+    frame = JuliaInterpreter.enter_call(f_multi, 1)
+    stack = [frame]
+    locals = JuliaInterpreter.locals(frame)
+    @test length(locals) == 2
+    @test JuliaInterpreter.Variable(1, :x, false) in locals
+    JuliaInterpreter.step_expr!(stack, frame)
+    JuliaInterpreter.step_expr!(stack, frame)
+    locals = JuliaInterpreter.locals(frame)
+    @test length(locals) == 3
+    @test JuliaInterpreter.Variable(1, :c, false) in locals
+    JuliaInterpreter.step_expr!(stack, frame)
+    locals = JuliaInterpreter.locals(frame)
+    @test length(locals) == 3
+    @test JuliaInterpreter.Variable(2, :x, false) in locals
+    JuliaInterpreter.step_expr!(stack, frame)
+    locals = JuliaInterpreter.locals(frame)
+    @test length(locals) == 3
+    @test JuliaInterpreter.Variable(3, :x, false) in locals
+end


KristofferC:
Is this a use-case for named tuples? The custom printing is pretty nice to have though.

src/JuliaInterpreter.jl
```diff
@@ -114,6 +114,50 @@ end
...
     JuliaStackFrame(frame.code, frame, pc; kwargs...)
 
+"""
+`LocalVariable` represents a local variable inside a frame.
+By calling the function `locals`[@ref] on a `JuliaStackFrame`[@ref] a
+`Vector` of `LocalVariable`'s is returned.
+
+Important fields:
+- `value`: the value of the local variable
+- `symbol`: the symbol of the variable as was
```
pfitzseb: as was?

KristofferC: Thanks, did a github edit to fix. yolo

pfitzseb:
The `show` method should maybe incorporate some of the logic in [`Debugger.print_var`](https://github.com/JuliaDebug/Debugger.jl/blob/49b21224365047b813698b4208f93231bd257293/src/printing.jl#L7-L24), I think.

And having a custom type is perfectly fine here, imho.

KristofferC:
We could pull in https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/34 here

timholy:
> Is a Vector the correct thing to return, perhaps a `Set` is more correct?

If we don't care about ordering, how about `Dict{Symbol,LocalVar}` where we drop the `name` from `LocalVar`. But maybe we care about ordering.

Another issue is how to handle `x@_2` vs `x@_4` variables. Is this intended to be literal or focus on "parent symbols" that appear in user code?

src/JuliaInterpreter.jl
```diff
@@ -115,6 +115,50 @@ end
...
+`LocalVariable` represents a local variable inside a frame.
+By calling the function `locals`[@ref] on a `JuliaStackFrame`[@ref] a
+`Vector` of `LocalVariable`'s is returned.
+
+Important fields:
+- `value::Any`: the value of the local variable
+- `name::Symbol`: the name of the variable as given in the source code
+- `isparam::Bool`: if the variable is a type parameter, for example `T` in `f(x::T) where {T} = x` .
+"""
+struct LocalVariable
```
KristofferC: Perhaps this should just be called `Variable`, is there anything specific about it being `local`. If we want to show all globals, we would presumably use the same struct.

KristofferC:
> The `show` method should maybe incorporate some of the logic in [`Debugger.print_var`](https://github.com/JuliaDebug/Debugger.jl/blob/49b21224365047b813698b4208f93231bd257293/src/printing.jl#L7-L24). I think e.g. big arrays will cause trouble otherwise (I guess you could pipe through the correct `IOContext` as well).

True but the point isn't really to use these `show` methods in "frontend" applications. They should just write their own `print_var` methods. It is just to have a decent default. They already `show` compactly I think:

```
> JuliaInterpreter.Variable(rand(100), :foo, false)
foo = [0.607186, 0.82001, 0.0310669, 0.0289033, 0.905648, 0.640141, 0.244837, 0.374682, 0.840797, 0.901677  …  0.185747, 0.0583803, 0.253379, 0.438763,
0.629232, 0.210685, 0.762034, 0.0228114, 0.769191, 0.792854]
```

KristofferC:
I updated the debugger printing based on this (https://github.com/JuliaDebug/Debugger.jl/pull/23).

src/JuliaInterpreter.jl
```diff
@@ -115,6 +115,50 @@ end
...
+        if !isa(frame.locals[i], Nothing)
+            # Remove #self# here in cases where it is just the method itself?
+            push!(vars, Variable(something(frame.locals[i]), frame.code.code.slotnames[i], false))
+        end
+    end
+    if frame.code.scope isa Method
+        sig = frame.code.scope.sig
+        for i in 1:length(frame.sparams)
+            @assert sig isa UnionAll 
+            # Only insert this if we haven't shadowed it with a local variable?
```
KristofferC: Actually, this is a syntax error.

timholy:
This seems fine as it stands, but I want to bring up https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/54#issuecomment-466116100 again. Let me illustrate the issue:
```julia
julia> m = @which maximum(rand(5,5); dims=1)
(::getfield(Base, Symbol("#kw##maximum")))(::Any, ::typeof(maximum), a::AbstractArray) in Base

julia> ci = Base.uncompressed_ast(m)
CodeInfo(
1 ─ %1  = (Base.haskey)(#temp#@_2, :dims)
└──       goto #3 if not %1
2 ─       #temp#@_6 = (Base.getindex)(#temp#@_2, :dims)
└──       goto #4
3 ─       #temp#@_6 = Base.:(:)
4 ┄       dims = #temp#@_6
│   %7  = (:dims,)
│   %8  = (Core.apply_type)(Core.NamedTuple, %7)
│   %9  = (Base.structdiff)(#temp#@_2, %8)
│   %10 = (Base.pairs)(%9)
│   %11 = (Base.isempty)(%10)
└──       goto #6 if not %11
5 ─       goto #7
6 ─       (Base.kwerr)(#temp#@_2, , a)
7 ┄ %15 = (Base.:(#maximum#554))(dims, , a)
└──       return %15
)

julia> ci.slotnames
6-element Array{Any,1}:
 Symbol("#unused#")
 Symbol("#temp#")  
 Symbol("")        
 :a                
 :dims             
 Symbol("#temp#")  
```
There are two slots named `#temp#`: one gets set at entry and then it gets set to a new value on either line 3 or 6. Is this something worth worrying about, or not?

pfitzseb:
IMHO it'd be fine to ignore all "internal" locals, but maybe there's a usecase for wanting to get *all* slots?

KristofferC:
Is there a difference with that example and something like:

```
julia> function f(x)
           x = 2
           x = 3
           x = 4
       end

julia>  m = @which f(1)
f(x) in Main at REPL[33]:2

julia>  ci = Base.uncompressed_ast(m)
CodeInfo(
1 ─     x@_3 = x@_2
│       x@_3 = 2
│       x@_3 = 3
│       x@_3 = 4
└──     return 4
)
```

In this case, just giving `x = ` whatever value it has depending on the pc seems reasonable?

FWIW, it also seems Debugger.jl has a bug here:

```
> @enter f(2)
In f(x) at REPL[5]:2
2            x = 2
3            x = 3
4            x = 4
5        end

About to run: return 4
1|debug> fr
[1] f(x) at REPL[5]:2
  | x::Int64 = 2
  | x::Int64 = 4
1|julia> x
> 4
```

timholy:
> just giving x = whatever value it has depending on the pc seems reasonable?

That's where I was going with this, :+1:

KristofferC:
Pushed a commit to only keep the latest variable with a given name,

src/JuliaInterpreter.jl
```diff
@@ -129,6 +129,49 @@ end
...
+    end
+    if frame.code.scope isa Method
+        sig = frame.code.scope.sig
+        for i in 1:length(frame.sparams)
+            @assert sig isa UnionAll 
+            push!(vars, Variable(frame.sparams[i], Symbol(sig.var.name), true))
+            sig = sig.body
+        end
+    end
+    return unique(v -> v.name, Iterators.reverse(vars))
```
KristofferC: Maybe this should have a `reverse` on it?

timholy: It occurs to me that this is what `frame.last_reference` seems to be for. But I'm not sure if these can ever "go backwards." @Keno, any insight?

Keno: Maybe? I'm not entirely sure what's being asked though ;).

timholy: Fair enough. This is mostly in reference to phenomena like https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/54#issuecomment-467489347. Questions:
- what's the maximum number of slots that will share the same slotname? (2?)
- do these exhibit SSA-like properties? That is, once `x@_5` gets assigned do we have to worry about `x@_2` ever holding the most recent value? In other words, do we really need that `Dict` or can we just scan `locals` backwards?

Keno: Ah, I don't really know the details - this is an artifact of lowering, so you may have to ask @JeffBezanson

src/JuliaInterpreter.jl
```diff
@@ -129,6 +129,54 @@ end
...
+
+"""
+    local_variables = locals(frame::JuliaStackFrame)::Vector{Variable}
+
+Return the local variables as a vector of `Variable`[@ref].
+"""
+function locals(frame::JuliaStackFrame)
+    vars = Variable[]
+    syms = Set{Symbol}()
+    for i = length(frame.locals):-1:1
```
timholy: As long as these truly are SSA-like then this is fine. Let's wait for confirmation from @JeffBezanson. If not, #81 includes a fix for `last_reference` and you can use that to look up values.

KristofferC: Sounds good!

timholy:
Since we're in a bit of limbo here, I'd support merging and fixing any problems we discover. Alternatively you can use the dict now. (But I kind of like the idea of keeping it this way and finding out if it works :wink:)

src/JuliaInterpreter.jl
```diff
@@ -158,6 +158,54 @@ end
...
+        if !isa(frame.locals[i], Nothing)
+            # Remove #self# here in cases where it is just the method itself?
+            sym = frame.code.code.slotnames[i]
+            sym in syms && continue
+            push!(vars, Variable(something(frame.locals[i]), sym, false))
+            push!(syms, sym)
+        end
+    end
+    reverse!(vars)
+    if frame.code.scope isa Method
```
KristofferC: This is just the inlined `sparams_syms`. Will switch to using the method instead.

KristofferC:
Let's try this then.

src/JuliaInterpreter.jl
timholy: Actually, Jeff's example inhttps://github.com/JuliaLang/julia/issues/31255#issuecomment-469823809 makes me think we definitely need the Dict (good choice in #94, @KristofferC). 

```julia
julia> function f(x)
           x = let x = x
               x = let x = x
                   x = let x = x
                       x + 1
                   end
                   x + 1
               end
               x + 1
           end
           x
       end

julia> @code_lowered f(1)
CodeInfo(
1 ─      x@_6 = x@_2
│   %2 = x@_6
│        x@_3 = %2
│   %4 = x@_3
│        x@_4 = %4
│   %6 = x@_4
│        x@_5 = %6
│        x@_4 = x@_5 + 1
│        x@_3 = x@_4 + 1
│        x@_6 = x@_3 + 1
└──      return x@_6
)
```

Keno: Part of the problem is that in lowered IR, the scoping for variables is no longer obvious (and the mapping from surface syntax locations where it is to lowered IR where it is not).

KristofferC: > the scoping for variables is no longer obvious

Yep https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/94#issuecomment-469656569

PR #55, opened 2019-02-21
add location info for stackframes

KristofferC:
Moving over some more useful stuff from Debugger.jl

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -7,6 +7,16 @@ using Core: CodeInfo, SSAValue, SlotNumber, TypeMapEntry, SimpleVector, LineInfo
 
 export @enter, @make_stack, @interpret, Compiled, JuliaStackFrame
 
+
+const SEARCH_PATH = []
+
+function __init__()
+    append!(SEARCH_PATH,[joinpath(Sys.BINDIR,"../share/julia/base/"),
+                         joinpath(Sys.BINDIR,"../include/")])
+    return nothing
+end
+
+
 """
 `Compiled` is a trait indicating that any `:call` expressions should be evaluated
 using Julia's normal compiled-code evaluation. The alternative is to pass `stack=JuliaStackFrame[]`,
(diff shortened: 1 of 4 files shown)


test/interpret.jl
```diff
@@ -214,3 +214,21 @@ frame = JuliaInterpreter.enter_call(f, 3)
...
+
+@testset "location info" begin
+    frame = JuliaInterpreter.enter_call(f_locationinfo, 1, 2)
+    JuliaInterpreter.step_expr!([frame], frame)
+    locinfo = JuliaInterpreter.location(frame)
+    @test locinfo.infile == true
+    @test locinfo.data == nothing
+    @test locinfo.filepath == @__FILE__
+    @test locinfo.line == methoddef + 2 # at d = 1 + c
+    @test locinfo.defline == methoddef + 1 # Why + 1 here??
```
pfitzseb: Because Julia gives the line of the first expression as the function definition line number in some/many cases:
grafik

KristofferC: Without thinking much about it, that seems like a bad thing...

src/JuliaInterpreter.jl
```diff
@@ -114,6 +124,65 @@ end
 JuliaStackFrame(frame::JuliaStackFrame, pc::JuliaProgramCounter; kwargs...) =
     JuliaStackFrame(frame.code, frame, pc; kwargs...)
 
+"""
+`LocationInfo` contains information about the location of the source code for a `JuliaStackFrame`[@ref].
+
+It is constructed by calling `location(frame::JuliaStackFrame)`[@ref].
+
+Important fields:
+- `file`: Whether the method was defined in a file (`isfile == true`) or the REPL.
```
timholy: Several typos, you want `infile` I think

src/JuliaInterpreter.jl
```diff
@@ -114,6 +124,65 @@ end
...
+- `filepath`: Path to the file of the method if it was defined in a file, otherwise `nothing`.
```
timholy: Might be good to keep the same order as the fields appear?

src/JuliaInterpreter.jl
```diff
@@ -114,6 +124,65 @@ end
...
+- `line`: The current line of the expression about to be executed.
+- `defline`: The line where the method is defined.
+"""
+struct LocationInfo
+    infile::Bool
+    data::Union{Nothing, String}
+    filepath::Union{Nothing, String}
+    line::Int
+    defline::Int
+end
```
timholy: Perhaps relevant: CodeTracking is ready to go live. I'm leaving [this PR](https://github.com/timholy/CodeTracking.jl/pull/2) open for a bit for comments, but once that merges I will register it.

Once that happens, I propose we use CodeTracking in this package. That will give us Revise integration, once I push a new version of Revise.

KristofferC: Sure.

KristofferC:
Let's move to CodeTracking instead.

PR #60, opened 2019-02-22
Reduce startup latency

### LINKED ISSUE #57: Latency on first call unacceptable
This package loads quite quickly:
```julia
julia> @time(using JuliaInterpreter)
  0.146759 seconds (178.45 k allocations: 11.704 MiB, 3.25% gc time)
```
but the first usage is slow:
```julia
julia> @time @interpret 1+1
  3.875351 seconds (1.34 M allocations: 65.670 MiB, 1.16% gc time)
2
```
Especially if Revise is going to be transitioning to using JuliaInterpreter for signature-extraction, this needs improvement.

### LINKED ISSUE #59: Caching of builtins.jl might be problematic
Two different versions of julia using the same version of ASTInterpreter2 will use the same package folder. However, since the julia versions are different, they might require different `builtins.jl`. Perhaps the `builtins.jl` file should be namespaced based on the julia version.

timholy:
This is addressed at mitigating #57. The main thing here is that `maybe_evaluate_builtin` is now a much shorter function; overall latency is halved and the runtime performance hit is only ~3% (not even really certain it's measurable). I think that's acceptable, and to say this fixes #57.

Also fixes #59. Now for anyone `dev`ing this package you have to `]build JuliaInterpreter` manually on each Julia version. Can also delete the file `builtins.jl`.


timholy:
Don't merge as-is, I see this is built on top of #58

Issue #63, opened 2019-02-23
`linenumber` function assumes all statements have location info


Keno:
However, that's not necessarily the case. It is legal for the index to be `0` corresponding to unknown location info. E.g.

```
diff --git a/src/interpret.jl b/src/interpret.jl
index 1a22e66..f7c3830 100644
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -616,6 +616,7 @@ isgotonode(node) = isa(node, GotoNode) || isexpr(node, :gotoifnot)
 linenumber(frame) = linenumber(frame, frame.pc[])
 function linenumber(frame, pc)
     codeloc = frame.code.code.codelocs[pc.next_stmt]
+    codeloc == 0 && return nothing
     return frame.code.scope isa Method ?
         frame.code.code.linetable[codeloc].line :
         codeloc
```

but obviously the callers would have to be adjusted to deal with that, so there may be a better way.

PR #65, opened 2019-02-24
handle `codeloc==0` in `linenumber`

pfitzseb:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/63.


KristofferC:
Would `return nothing` here be more "julian" in some way?

pfitzseb:
Yup, unless you know how to make the printing prettier? :)

KristofferC:
🤷‍♂️ why not, let's try that. Feels a bit less C-y :P

pfitzseb:
Done.

Issue #62, opened 2019-02-23
Use `Base.copy(::CodeInfo)` rather than custom copy_codeinfo


Keno:
I was about to suggest the following diff to prevent the method at:

https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/ca07ee9acada7257ab80c964955e5a31e4091f18/src/JuliaInterpreter.jl#L550

```
diff --git a/src/JuliaInterpreter.jl b/src/JuliaInterpreter.jl
index 1caa2e0..f528b47 100644
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -552,7 +552,7 @@ function copy_codeinfo(code::CodeInfo)
     for (i, name) in enumerate(fieldnames(CodeInfo))
         if isdefined(code, name)
             val = getfield(code, name)
-            ccall(:jl_set_nth_field, Cvoid, (Any, Csize_t, Any), newcode, i-1, val===nothing || isa(val, Type) ? val : copy(val))
+            ccall(:jl_set_nth_field, Cvoid, (Any, Csize_t, Any), newcode, i-1, val===nothing || isa(val, Union{Type, Method}) ? val : copy(val))
         end
     end
     return newcode
```

from trying to copy a `Method`, but then I remembered that base has an implementation of `copy(::CodeInfo)` that should just be used instead.

PR #69, opened 2019-02-26
use copy instead of copy_codeinfo

KristofferC:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/62


timholy:
Talk about simultaneous!

src/precompile.jl
```diff
@@ -33,7 +33,6 @@ function _precompile_()
     precompile(Tuple{typeof(extract_args), Module, Expr})
     precompile(Tuple{typeof(enter_call), Int, Int})
     precompile(Tuple{typeof(enter_call_expr), Expr})
-    precompile(Tuple{typeof(copy_codeinfo), Core.CodeInfo})
```
KristofferC: I don't think this is needed

```
> @time precompile(Tuple{typeof(copy), Core.CodeInfo})
0.000011 seconds (9 allocations: 448 bytes)
```

KristofferC:
Looks like this only exists in 1.1. Unfortunate.

PR #70, opened 2019-02-26
Delete copy_codeinfo (just use copy). Fixes #62

### LINKED ISSUE #62: Use `Base.copy(::CodeInfo)` rather than custom copy_codeinfo

timholy:


timholy:
Superseded by #69

Issue #28, opened 2019-02-14
MWE of char crash


GunnarFarneback:
This reproduces the crash in the `char` tests.
```
using JuliaInterpreter
ex = quote
    a = ['0']
    b = ['a']
    c = [a; b]
end
frame = JuliaInterpreter.prepare_toplevel(Main, ex)
JuliaInterpreter.finish_and_return!(JuliaStackFrame[], frame, true)
```
with the result
```
Unreachable reached at 0x7f9c2cce8ab4

signal (4): Illegal instruction
[...]
```

KristofferC:
Seems to work locally on Windows on both 1.1 and master.

```
> JuliaInterpreter.finish_and_return!(JuliaStackFrame[], frame, true)
> 2-element Array{Char,1}:
 '0'
 'a'

> versioninfo()
Julia Version 1.1.0
Commit 80516ca202 (2019-01-21 21:24 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

GunnarFarneback:
```
julia> versioninfo()
Julia Version 1.2.0-DEV.320
Commit 38d6247 (2019-02-13 21:23 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
Environment:
  IPYTHON_JL_SETUP_PYJULIA = no
```

timholy:
I can replicate with both 1.1 and master.
```
julia> versioninfo()
Julia Version 1.1.1-pre.0
Commit a84cf6f56c* (2019-01-22 04:33 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
Environment:
  JULIAFUNCDIR = /home/tim/juliafunc
  JULIA_CPU_THREADS = 2
[...]

PR #72, opened 2019-02-26
More like #28

timholy:
I missed a couple of ccalls to memcpy in #68 (more issue #28).


timholy:
This seems to fix the remaining :skull_and_crossbones: in Julia's tests (I've not run them all together, but they work individually).

Also note this throws in a convenience method for writing the results of the juliatests to an arbitrary file.

PR #81, opened 2019-02-27
Add support for breakpoints (!)

timholy:
Would be great to see what people think of the API. Currently the main API is:

- `breakpoint(f, sig, condition)`: add a breakpoint on entry to the method given by `which(f, sig)`
- `breakpoint(f, condition)`: add break-on-entry for all methods of `f`

Notably missing is `breakpoint(file, lineno, condition)`; I'll add that once I work on CodeTracking.jl integration. There are also `Breakpoints.enable`, `Breakpoints.disable`, and `Breakpoints.remove`, which can either take a specific `bp` to operate on or can be used as `Breakpoints.remove()` to get rid of all breakpoints.

There is also the non-exported all-powerful `breakpoint!(framecode, pc, condition)` that allows you set them wherever you want; this could easily be used when stepping through code to set a breakpoint at the current `pc`. (The tests illustrate this to set a breakpoint in the interior of a loop, but for now it's a purely manual operation.)

Internally, the way this works is to compile a function, call it `slotfunction`, that takes a `frame` as its only input, `slotfunction(frame)::Bool`. When this returns `true`, you've satisfied the condition and should stop. Implementing `slotfunction` steals bits from Debugger (portions of `eval_code`), and I copied over `sparam_syms` wholesale. If this gets merged I will submit a PR to delete it from Debugger and have it use this package's implementation.

Happy to modify aspects of this to make it easier for Debugger, Juno, and/or Rebugger to use.

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -41,8 +41,10 @@ isless(x::JuliaProgramCounter, y::Integer) = isless(x.next_stmt, y)
 
 Base.show(io::IO, pc::JuliaProgramCounter) = print(io, "JuliaProgramCounter(", pc.next_stmt, ')')
 
+# Breakpoint support
 truecondition(frame) = true
 falsecondition(frame) = false
+const break_on_error = Ref(false)
 
 """
     BreakpointState(isactive=true, condition=JuliaInterpreter.truecondition)
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -1119,8 +1119,6 @@ macro interpret(arg)
         if frame === nothing
             return eval(Expr(:call, map(QuoteNode, theargs)...))
         end
-        empty!(framedict)  # start fresh each time; kind of like bumping the world age at the REPL prompt
-        empty!(genframedict)
         finish_and_return!(stack, frame)
     end
 end
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -11,7 +11,7 @@ using UUIDs
 using Random.DSFMT
 using InteractiveUtils
 
-export @enter, @make_stack, @interpret, Compiled, JuliaStackFrame
+export @enter, @make_stack, @interpret, Compiled, JuliaStackFrame, Breakpoints, breakpoint
 
 module CompiledCalls
 # This module is for handling intrinsics that must be compiled (llvmcall)
@@ -47,7 +47,7 @@ end
 
 """
 `JuliaFrameCode` holds static information about a method or toplevel code.
-One `JuliaFrameCode` can be shared by many `JuliaFrameState` calling frames.
+One `JuliaFrameCode` can be shared by many `JuliaStackFrame` calling frames.
 
 Important fields:
 - `scope`: the `Method` or `Module` in which this frame is to be evaluated
@@ -61,6 +61,7 @@ struct JuliaFrameCode
     scope::Union{Method,Module}
     code::CodeInfo
     methodtables::Vector{Union{Compiled,TypeMapEntry}} # line-by-line method tables for generic-function :call Exprs
+    breakpoints::Vector{Tuple{Bool,Function}}          # (isactive, condition)
     used::BitSet
     wrapper::Bool
     generator::Bool
@@ -69,7 +70,7 @@ struct JuliaFrameCode
 end
 
 function JuliaFrameCode(frame::JuliaFrameCode; wrapper = frame.wrapper, generator=frame.generator, fullpath=frame.fullpath)
-    JuliaFrameCode(frame.scope, frame.code, frame.methodtables, frame.used,
+    JuliaFrameCode(frame.scope, frame.code, frame.methodtables, frame.breakpoints, frame.used,
                    wrapper, generator, fullpath)
 end
 
@@ -80,8 +81,9 @@ function JuliaFrameCode(scope, code::CodeInfo; wrapper=false, generator=false, f
         code = copy_codeinfo(code)
         methodtables = Vector{Union{Compiled,TypeMapEntry}}(undef, length(code.code))
     end
+    breakpoints = Vector{Tuple{Bool,Function}}(undef, length(code.code))
     used = find_used(code)
-    return JuliaFrameCode(scope, code, methodtables, used, wrapper, generator, fullpath)
+    return JuliaFrameCode(scope, code, methodtables, breakpoints, used, wrapper, generator, fullpath)
 end
 
 """
@@ -211,6 +213,16 @@ function is_generated(meth)
     isdefined(meth, :generator)
 end
 
+function sparam_syms(meth::Method)
+    s = Symbol[]
+    sig = meth.sig
+    while sig isa UnionAll
+        push!(s, Symbol(sig.var.name))
+        sig = sig.body
+    end
+    return s
+end
+
 function namedtuple(kwargs)
     names, types, vals = Symbol[], [], []
     for pr in kwargs
@@ -267,6 +279,48 @@ function prepare_args(@nospecialize(f), allargs, kwargs)
         allargs = Base.append_any((allargs[2],), allargs[3:end]...)
     end
     return f, allargs
+end
+
+function prepare_framecode(method::Method, argtypes; enter_generated=false)
+    sig = method.sig
+    isa(method, TypeMapEntry) && (method = method.func)
+    if method.module == Core.Compiler || method.module == Base.Threads || method ∈ compiled_methods
+        return Compiled()
+    end
+    # Get static parameters
+    (ti, lenv::SimpleVector) = ccall(:jl_type_intersection_with_env, Any, (Any, Any),
+                        argtypes, sig)::SimpleVector
+    enter_generated &= is_generated(method)
+    if is_generated(method) && !enter_generated
+        framecode = get(genframedict, (method, argtypes), nothing)
+    else
+        framecode = get(framedict, method, nothing)
+    end
+    if framecode === nothing
+        if is_generated(method) && !enter_generated
+            # If we're stepping into a staged function, we need to use
+            # the specialization, rather than stepping through the
+            # unspecialized method.
+            code = Core.Compiler.get_staged(Core.Compiler.code_for_method(method, argtypes, lenv, typemax(UInt), false))
+            code === nothing && return nothing
+            generator = false
+        else
+            if is_generated(method)
+                code = get_source(method.generator)
+                generator = true
+            else
+                code = get_source(method)
+                generator = false
+            end
+        end
+        framecode = JuliaFrameCode(method, code; generator=generator)
+        if is_generated(method) && !enter_generated
+            genframedict[(method, argtypes)] = framecode
+        else
+            framedict[method] = framecode
+        end
+    end
+    return framecode, lenv
 end
 
 function whichtt(tt)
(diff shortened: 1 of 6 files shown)


src/breakpoints.jl
```diff
@@ -0,0 +1,186 @@
...
+    framecode::JuliaFrameCode
+    stmtidx::Int
+end
+
+function Base.show(io::IO, bp::Breakpoint)
+    lineno = linenumber(bp.framecode, bp.stmtidx)
+    print(io, "breakpoint(", bp.framecode.scope, ", ", lineno, ')')
+end
+
+const _breakpoints = Set{Breakpoint}()
```
KristofferC: I think this should be a vector and the API in the REPL debugger for disabling or removing a breakpoint would give the index into that vector. Also when showing it is nice to do it in the order that they where added.

src/JuliaInterpreter.jl
```diff
@@ -61,6 +61,7 @@ struct JuliaFrameCode
     scope::Union{Method,Module}
     code::CodeInfo
     methodtables::Vector{Union{Compiled,TypeMapEntry}} # line-by-line method tables for generic-function :call Exprs
+    breakpoints::Vector{Tuple{Bool,Function}}          # (isactive, condition)
```
KristofferC: Could potentially make this a named tuple `(isactive = true, f = ...)` for a bit nicer usage later `bp.isactive` over `bp[1]`.

src/breakpoints.jl
```diff
@@ -0,0 +1,186 @@
...
+
+"""
+    breakpoint(f)
+    breakpoint(f, condition)
+
+Break-on-entry to all methods of `f`.
+"""
+function breakpoint(f, condition::Union{Bool,Expr}=true)
+    for method in methods(f)
+        breakpoint(method, condition)
```
KristofferC: Im a bit worried about the number of breakpoints this might add in some cases. From a users point of view, if you add a breakpoint to `push!` you likely consider that one breakpoint that you handle and not `n` individual. 

Alternatively, it is up to the frontend to collapse all these breakpoints into one.

KristofferC: I guess my question is if there should be "higher" breakpoint structs like

```
struct FunctionBreakpoint
    f::Function
    bps::Vector{JuliaFrameCode, Int}
end
```

KristofferC: Im fine with the answer being that this is the low level end of things and that frontends should just wrap things and provide higher level abstractions themselves

src/breakpoints.jl
KristofferC: https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/81/files#r261159550 applies here as well

pfitzseb: I think it'd also be useful to have an API for accessing this (probably readonly).

KristofferC:
Is it intended that this will not work with `@interpret` (due to it emptying the `framedict`) or should the breakpoints be re-added for every `@interpret` call?

timholy:
The emptying occurs with the `frame`, not the `framecode`; one of the nice things about the split between `JuliaFrameCode` and `JuliaStackFrame` is that the same `framedict` gets (re)used by every frame. This is also true for recursive calls, where you may have the same `framedict` appearing in several of the frames on the stack. You want breakpoints associated with code, not particular stackframes.

src/breakpoints.jl
timholy: What's wrong with many breakpoints? Just that you have to prepare so many framecodes?

```julia
julia> using JuliaInterpreter

julia> foo(x) = x

julia> breakpoint(foo)  # compile necessary methods

julia> @time breakpoint(push!)
  0.045100 seconds (15.96 k allocations: 977.291 KiB)

julia> length(Breakpoints._breakpoints)
22
```

src/breakpoints.jl
timholy: Is the main purpose to list the current breakpoints? Or are there other uses intended?

Unless I'm missing something, making it readonly seems very tricky to me. I'd favor a `print_breakpoints` function that just dumps textual information.

timholy: Oh, now I understand: you want to disable a specific breakpoint by providing its index. Gotcha.

KristofferC: Like in gdb you do `info b` and then you can disable the breakpoint based on an index.

src/breakpoints.jl
KristofferC: > From a users point of view, if you add a breakpoint to `push!` you likely consider that one breakpoint that you handle and not `n` individual.
> Alternatively, it is up to the frontend to collapse all these breakpoints into one.

So a user would likely want to be able to (something like)

```
bp = breakpoint(push!)
disable(bp)
```

in other words, handle the breakpoint on `push!` as it was a single breakpoint. Anyway, it is easy to workaround from a frontends pov so dont have to change anything here.

timholy:
One thing I forgot to ask about was `enable` and friends. It would make sense to export `enable(bp)`, but `enable()` is too vague to export. Consequently this currently requires module-scoping, `Breakpoints.enable()`. But currently this also forces `Breakpoints.enable(bp)`, which seems dumb.

It occurs to me, though, that we could make `enable(bp)` an exported method of JuliaInterpreter, and then have `Breakpoints.enable` be a different function that would call `JuliaInterpreter.enable`. It seems slightly weird, but perhaps from the user standpoint this makes sense. Thoughts?

KristofferC:
>  It would make sense to export `enable(bp)`, but `enable()` is too vague to export.

I don't really understand this, you export a function and `enable(bp)` and `enable()` is the same function. Even if we make `enable()` a different function it won't help against any name collisions from other packages exporting `enable`.

> But currently this also forces `Breakpoints.enable(bp)`, which seems dumb

Doesn't seems too dumb to me. You can always import it.

src/breakpoints.jl
pfitzseb: I meant "readonly" in the sense that mutating the vector returned by `get_breakpoints` (or whatever) isn't supported/encouraged by the API. Instead we'd recommend using the `breakpoint()`/`remove()` API.

src/breakpoints.jl
pfitzseb: Regardless of whether it's one meta-breakpoint or a vector of normal breakpoints, it would be useful to them here, instead of `nothing`.

src/interpret.jl
```diff
@@ -673,6 +735,7 @@ function next_line!(stack, frame, dbstack = nothing)
             pc == nothing && return nothing
         end
         frame.pc[] = pc
+        shouldbreak(frame, pc) && break
```
KristofferC: Is something off here?

```julia
> function f(x)
      g(x)
  end

> function g(x)
      print("hello")
      return 2
  end

> breakpoint(g)

> frame = JuliaInterpreter.enter_call(f, 2)
JuliaStackFrame(JuliaInterpreter.JuliaFrameCode(f(x) in Main at REPL[18]:2, CodeInfo(
1 ─ %1 = ($(QuoteNode(g)))(x)
└──      return %1
), Union{Compiled, TypeMapEntry}[#undef, #undef], Tuple{Bool,Function}[#undef, #undef], BitSet([1]), false, false, true), Union{Nothing, Some{Any}}[Some(f), Some(2)], Any[#undef, #undef], Any[], Int64[], Base.RefValue{Any}(nothing), Base.RefValue{JuliaInterpreter.JuliaProgramCounter}(JuliaProgramCounter(1)), Dict(Symbol("#self#")=>1,:x=>2), Any[])

> JuliaInterpreter.next_line!([frame], frame)
ERROR: MethodError: Cannot `convert` an object of type JuliaInterpreter.Breakpoints.Breakpoint to an object of type JuliaInterpreter.JuliaProgramCounter
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:154
  JuliaInterpreter.JuliaProgramCounter(::Any) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\JuliaInterpreter.jl:34
Stacktrace:
 [1] setproperty!(::Base.RefValue{JuliaInterpreter.JuliaProgramCounter}, ::Symbol, ::JuliaInterpreter.Breakpoints.Breakpoint) at .\sysimg.jl:19
 [2] setindex!(::Base.RefValue{JuliaInterpreter.JuliaProgramCounter}, ::JuliaInterpreter.Breakpoints.Breakpoint) at .\refvalue.jl:33
 [3] next_line!(::Array{JuliaStackFrame,1}, ::JuliaStackFrame, ::Nothing) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:744
 [4] next_line!(::Array{JuliaStackFrame,1}, ::JuliaStackFrame) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:708
 [5] top-level scope at none:0
```

src/breakpoints.jl
KristofferC: > it would be useful to them here

Did you accidentally a word here?

src/interpret.jl
KristofferC:
```diff
diff --git a/src/interpret.jl b/src/interpret.jl
index 354c9ee..ec4b013 100644
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -710,7 +710,7 @@ function next_line!(stack, frame, dbstack = nothing)
         # If this is a goto node, step it and reevaluate
         if isgotonode(expr)
             pc = _step_expr!(stack, frame, pc)
-            pc == nothing && return nothing
+            (pc == nothing || isa(pc, Breakpoint)) && return pc
         elseif dbstack !== nothing && iswrappercall(expr)
             # With splatting it can happen that we do something like ssa = tuple(#self#), _apply(ssa), which
             # confuses the logic here, just step into the first call that's not a builtin
@@ -727,15 +727,15 @@ function next_line!(stack, frame, dbstack = nothing)
                     break
                 else
                     pc = _step_expr!(stack, frame, pc)
-                    pc == nothing && return nothing
+                    (pc == nothing || isa(pc, Breakpoint)) && return pc
                 end
             end
         else
             pc = _step_expr!(stack, frame, pc)
-            pc == nothing && return nothing
+            (pc == nothing || isa(pc, Breakpoint)) && return pc
         end
         frame.pc[] = pc
-        shouldbreak(frame, pc) && break
+        shouldbreak(frame, pc) && return Breakpoint(frame.code, pc)
     end
     maybe_next_call!(stack, frame, pc)
 end
```

maybe?

timholy: Yes, that looks good. Feel free to push that here.

src/breakpoints.jl
timholy: Prob. "return them here"

src/interpret.jl
KristofferC: Done

timholy:
OK, I've addressed comments and reworked a bunch of types. I've renamed the reference-type `BreakpointRef` and endowed it with `Ref`-like properties, at least as far as setting `isactive` goes. See if you like this.

I've also started working on a macro `@breakpoint`. I'd like to support this kind of thing:
```julia
remove()
@breakpoint radius2 y>x
runsimple()
remove()
@breakpoint radius2(0,0) y>x
runsimple()
local_identity(x) = identity(x)
remove()
@breakpoint radius2(0,0) y>local_identity(x)
runsimple()
```

The scoping here is quite nontrivial, and I'm not currently certain that there is a solution that doesn't require global names. If anyone has good ideas...

timholy:
OK, I think this is as good as it's going to get for now. Scope problem is detailed in #85.

timholy:
Worth noting that `@breakpoint` uses different semantics than Gallium's old `@conditional`: with `@conditional` you still provided the function in terms of signature types

```julia
@conditional breakpoint(gcd,Tuple{Int,Int}) (a==5)
```
whereas here it would be
```julia
@breakpoint gcd(10, 20) (a==5)
```

One can get the Gallium-like version with
```julia
breakpoint(gcd, Tuple{Int,Int}, :(a==5))
```
so the only difference is the need to explicitly create the Expr. Does this seem reasonable?

timholy:
One worker failed and I restarted it; if you click on "details" you'll see all tests pass despite the :x: above.

timholy:
OK to merge?

KristofferC:
Heroic stuff!

src/interpret.jl
```diff
@@ -474,11 +481,16 @@ by normal dispatch, whereas a vector of `JuliaStackFrame`s will use recursive in
...
 function step_expr!(stack, frame, istoplevel::Bool=false)
     pc = _step_expr!(stack, frame, frame.pc[], istoplevel)
-    pc === nothing && return nothing
+    (pc === nothing || isa(pc, BreakpointRef)) && return pc
     frame.pc[] = pc
 end
 
-function handle_err(frame, err)
+function handle_err(stack, frame, pc, err)
+    if break_on_error[]
```
KristofferC: Hm, should we really break here even if the error end up getting caught?

timholy: Interesting question. Should we support both? `try/catch` often make debugging hard, breaking on the initial error can help figure out what's going wrong. But you're also right that when this is used as semi-normal control flow, the current setting is problematic.

If you like the idea of supporting both, any suggestions for names? How about `:none` (equivalent to `false` now), `:uncaught`, `:caught` (corresponding to `true` now)?

pfitzseb: Having both options sounds great, and the names are fine imho.

PR #86, opened 2019-03-02
fixes for stack inversion in Debugger.jl, also remove some unused functions

KristofferC:
In https://github.com/JuliaDebug/Debugger.jl/pull/31 (in preparation for breakpoint support) I changed the way the stack grows to correspond to JuliaInterpret (`push!` instead of `pushfirst!`). Debugging this took a bit longer than needed because there are functions in JuliaInterpreter only being used in Debugger.jl that assumes that the stack grew the other way. I moved over a few of them to Debugger.jl.

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -11,7 +11,7 @@ using UUIDs
 using Random.DSFMT
 using InteractiveUtils
 
-export @enter, @make_stack, @interpret, Compiled, JuliaStackFrame
+export @enter, @interpret, Compiled, JuliaStackFrame
 
 module CompiledCalls
 # This module is for handling intrinsics that must be compiled (llvmcall)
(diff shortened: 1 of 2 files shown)


src/interpret.jl
```diff
@@ -653,13 +653,13 @@ function next_line!(stack, frame, dbstack = nothing)
             # With splatting it can happen that we do something like ssa = tuple(#self#), _apply(ssa), which
             # confuses the logic here, just step into the first call that's not a builtin
             while true
-                dbstack[1] = JuliaStackFrame(JuliaFrameCode(frame.code; wrapper = true), frame, pc)
```
KristofferC: When Debugger shares the stack with JuliaInterpereter I expect this whole block to go away.

src/JuliaInterpreter.jl
```diff
@@ -11,7 +11,7 @@ using UUIDs
 using Random.DSFMT
 using InteractiveUtils
 
-export @enter, @make_stack, @interpret, Compiled, JuliaStackFrame
+export @enter, @interpret, Compiled, JuliaStackFrame
```
timholy: Hasn't `@enter` already moved to Debugger?

KristofferC: Good catch, removing it here as well.

PR #88, opened 2019-03-03
Integrate with CodeTracking and support line numbers when setting breakpoints

timholy:
If the user is running an appropriate version of Revise, reported location will update as code moves around due to edits. Because methods can move files as well as from line-to-line, `linenumber` should not be used except internally. (Use `whereis` instead.)

If the user is not running Revise, CodeTracking should be harmless. Of course this makes this package sensitive to bugs in CodeTracking, but because it's a small package hopefully there won't be many of those before it becomes an asset.

Here's a fun demo (requires https://github.com/timholy/CodeTracking.jl/pull/10 to work with methods defined at the REPL):

```julia
julia> using JuliaInterpreter

julia> function mysum(A)
           s = zero(eltype(A))
           for a in A
               s += a
           end
           return s
       end
mysum (generic function with 1 method)

julia> first(methods(mysum))  # note the first line is "2"
mysum(A) in Main at REPL[2]:2

julia> @breakpoint mysum(1:5) 4 s >= 10
breakpoint(mysum(A) in Main at REPL[2]:2, 4)

julia> stack, bp = @interpret mysum(1:10);

julia> bp
breakpoint(mysum(A) in Main at REPL[2]:2, 4)

julia> stack[1].locals
5-element Array{Union{Nothing, Some{Any}},1}:
 Some(mysum) 
 Some(1:10)  
 Some(10)    
 Some((5, 5))
 Some(5)     

julia> stack[1].code.code.slotnames
5-element Array{Any,1}:
 Symbol("#self#")
 :A              
 :s              
 Symbol("#temp#")
 :a              
```

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -3,21 +3,44 @@
 [[Base64]]
 uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
-[[Dates]]
-deps = ["Printf"]
-uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
+[[CodeTracking]]
+deps = ["Test", "UUIDs"]
+git-tree-sha1 = "591b73b37c92ed7d55d3a14e266829c21aa3a7eb"
+uuid = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
+version = "0.3.0"
+
+[[Distributed]]
+deps = ["Random", "Serialization", "Sockets"]
+uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 
 [[InteractiveUtils]]
 deps = ["Markdown"]
 uuid = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 
+[[Logging]]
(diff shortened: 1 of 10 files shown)


timholy:
All except the doctests passed, so I'm tempted to merge and fix those after cleanups.

There's one more step to go: support for `breakpoint(filename, line)`. With my current plan, that's going to require https://github.com/timholy/CodeTracking.jl/pull/11, which is no problem from the package standpoint, but for it to actually work at runtime it will require that the user runs Revise. Do folks view that as a problem? (In all other cases, there's a graceful fallback without Revise, but this is a hard one to get right without it.)

Juno presumably has its own approaches for finding a method given a file/lineno, though given my experience in LoweredCodeUtils I'd be surprised if it works for keyword-arg functions and optionally-`@generated` functions. The current version of Revise (not even https://github.com/timholy/Revise.jl/pull/245) won't get those either, but the new one (https://github.com/timholy/Revise.jl/pull/243) built on JuliaInterpreter/LoweredCodeUtils will.

PR #89, opened 2019-03-04
only break if we are throwing an uncaught exception

KristofferC:
This seems more in line with how I think a user would except break on error to work. Discussion welcome :)

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -486,6 +486,11 @@ function step_expr!(stack, frame, istoplevel::Bool=false)
 end
 
 function handle_err(stack, frame, pc, err)
+    if !isempty(frame.exception_frames)
+        # Exception caught
+        frame.last_exception[] = err
+        return JuliaProgramCounter(frame.exception_frames[end])
+    end
     if break_on_error[]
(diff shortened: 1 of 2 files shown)


timholy:
I think I'll go for [this](https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/81#discussion_r262155818) eventually, but if you don't want to implement it now, this is fine.

KristofferC:
Alright, let's do this first which I think is perhaps a better default and then we add the caught exception flag as well.

KristofferC:
Actually, this doesn't work when catching things across functions:

```jl
julia> function f_exc_inner()
      error()
  end

julia> function f_exc_outer()
      try
          f_exc_inner()
      catch err
          return err
      end
  end

julia> stack, bp = @interpret f_exc_outer();

julia> bp
breakpoint(error(s::Vararg{Any,N}) where N in Base at error.jl:41, 42, ErrorException(""))

julia> f_exc_outer()
ErrorException("")
```

We would try to completely undwind the stack, looking for a catch, and only if we bubble up to the top, we return the breakpoint.

PR #94, opened 2019-03-05
use last_reference instead for locals

KristofferC:
To make me sleep easier at night

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -184,14 +184,8 @@ Return the local variables as a vector of `Variable`[@ref].
 """
 function locals(frame::JuliaStackFrame)
     vars = Variable[]
-    syms = Set{Symbol}()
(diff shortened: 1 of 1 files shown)


KristofferC:
This is still a bit buggy since it doesn't handle scopes correctly:

```jl
julia> function f()
      let i = 2
         print(i)
      end
      return
  end

julia> frame = JuliaInterpreter.enter_call(f)

julia> pc = JuliaInterpreter.next_line!(stack, frame)
JuliaProgramCounter(2)

julia> pc = JuliaInterpreter.next_line!(stack, frame)
2julia> JuliaProgramCounter(3)

julia> JuliaInterpreter.locals(frame)
2-element Array{JuliaInterpreter.Variable,1}:
 #self# = f
 i = 2
```` 

`i = 2` has gotten out of scope at this `pc` so should probably not be listed.

KristofferC:
It isnt worse than what we have right now though.

PR #96, opened 2019-03-05
Support `breakpoint(filename, lineno)`

timholy:
This implements the last remaining way of setting breakpoints. Note that it depends on `signatures_at`, which in turn depends on Revise. This is the only truly robust solution I know of: for example, keyword-arg methods will need the breakpoint set in a gensymmed body method that cannot be easily discovered (https://github.com/JuliaLang/julia/issues/30908). I guess the question is whether we want a fallback that sometimes works.

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -5,7 +5,7 @@ using JuliaInterpreter: JuliaFrameCode, JuliaStackFrame, BreakpointState,
                         truecondition, falsecondition, prepare_framecode, get_framecode,
                         sparam_syms, linenumber, statementnumber
 using Base.Meta: isexpr
-using InteractiveUtils
+using CodeTracking, InteractiveUtils
 
 export @breakpoint, breakpoint, enable, disable, remove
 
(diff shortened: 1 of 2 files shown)


timholy:
OK, so I think I've worked out a way to do this that doesn't require Revise. (Much as I like Revise, I hate forcing people to use something they haven't explicitly chosen, especially if it changes behavior.) Strategy:
- parse the file, split expressions
- look for the block that contains the line number
- lower that block
- look for the last 3-arg :method Expr that starts at a line at or above the requested line number. The reason we do the last one is to avoid https://github.com/JuliaLang/julia/issues/30908, since the last method seems to reliably have the same name as a compiled method.
- move some code from `LoweredCodeUtils` here and compute the signature of that method
- call `which` to get that method
- *don't* set the breakpoint here: steal back Debugger's `maybe_step_through_wrapper`. The issue is that we might be in a wrapper that doesn't even have that line number.

Hmm, the last can't be done literally, because we don't have the call arguments to the method. But we could presumably steal the analysis.

Thoughts? Is this worth it? Or should we leave this as a TODO for later?

timholy:
Seems like folks aren't worried about the Revise requirement. We can come back to that later as needed.

PR #97, opened 2019-03-05
Try harder to find the line number

timholy:
Does this seem OK, @pfitzseb? Methods annotated with `@inline` and the like have `codeloc = 0` for `pc==1`, but there doesn't seem to be any particularly good reason to have this fail.


pfitzseb:
Yup, this seems like a strict improvement. 👍

Issue #102, opened 2019-03-05
More breakpoint automatic modes


timholy:
Currently `break_on_error[]` can take values `true` or `false`; if true, it breaks on any uncaught exception. Some other things that might be cool:

- break on caught exceptions (can help figure out tricky `try/catch`-induced issues)
- break on test failure
- break on warn

I haven't investigated the feasibility of these, so I have no idea what implementation would look like.

KristofferC:
I don't really think we need more of this than what we currently have.

timholy:
Break on test failure could be nice, but I agree it's not a priority and probably not useful to keep this issue open.

PR #109, opened 2019-03-06
Revert "only break if we are throwing an uncaught exception (#89)"

KristofferC:
This reverts commit 28af3b350bb2f6ab427e55e8cadbe539a5ef5a5e.

This behavior actually doesn't make sense since it will only not break for things that are caught in the same frame which is a really weird special case.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -488,11 +488,6 @@ function step_expr!(stack, frame, istoplevel::Bool=false)
 end
 
 function handle_err(stack, frame, pc, err)
-    if !isempty(frame.exception_frames)
-        # Exception caught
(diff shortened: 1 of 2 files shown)


timholy:
Sorry, I should have read that PR more carefully. We still should implement the behavior in the title, along with #102.

My intuition suggests that if `break_on_error[]` is true (or any of the more complex options), the best implementation would be to return the breakpointref, leaving the stack intact, and then if a catch is encountered you discard the breakpointref, pop the stack up to the current level, handle the exception and continue execution. This might be a good argument for the doubly-linked list, since that makes it much easier to truncate the callees, whereas that might require a search through the list with the current implementation.

Issue #110, opened 2019-03-06
Stack is not empty after returning from top function with a try catch


KristofferC:
```jl
julia> function f_exc_outer()
      try
          f_exc_inner()
      catch err
          return err
      end
  end

julia> function f_exc_inner()
      error()
  end

julia> stack = JuliaStackFrame[];

julia> frame = JuliaInterpreter.enter_call(f_exc_outer);

julia> JuliaInterpreter.finish_and_return!(stack, frame)
ErrorException("")

julia> length(stack)
2
```

It seems like we have some frames left in the stack even though we have exited from the top level function.

KristofferC:
Fixed by https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/111

PR #111, opened 2019-03-06
pop top frame from stack if we throw evaluating it

KristofferC:
Unless each `push!` is guaranteed to have a corresponding `pop!`, we might end up with frames in the stack even when the function exits (#110). Currently, if we throw in `exec!` the `pop!` will not be executed, so when the stack unwinds after an error, we are left with frames in the stack.

This should fix that.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -212,7 +212,12 @@ function evaluate_call!(stack, frame::JuliaStackFrame, call_expr::Expr, pc; exec
         push!(stack, newframe)
         return BreakpointRef(newframe.code, newframe.pc[])
     end
-    ret = exec!(stack, newframe)
+    ret = try
+        exec!(stack, newframe)
+    catch e
+        pop!(stack)
+        rethrow(e)
+    end
     isa(ret, BreakpointRef) && return ret
     pop!(stack)
     push!(junk, newframe)  # rather than going through GC, just re-use it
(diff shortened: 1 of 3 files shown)


KristofferC:
Yeah, I was thinking about this while out on a walk and I think I convinced myself that this would be problematic in the presence of `break_on_error[]`.

KristofferC:
Actually, is checking `break_on_error[]` needed? If `break_on_error[] = true` then we won't enter the `catch` because the `exec!` call just returns the breakpoint instead of throwing. So if `break_on_error[] = true` there should be no functional change here I think...

KristofferC:
I pushed another commit with some more tests including `break_on_error`

timholy:
I think you're right, this is fine if we want "break on any error" behavior (like I implemented initially). For "break on uncaught error" we'll need to insert a check in `_step_expr!` (probably at [this line](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/390bb66ba8fe06d431a78e3147a2a1c14b1e563c/src/interpret.jl#L462)) and intercept if `bp.err !== nothing` and `!isempty(frame.exception_frames)`.

KristofferC:
I rebased this on master (without implementing your `_step_expr!` check).

Do you have an example of a something that would fail with this version of the PR.

timholy:
I think given how you handled #115, this works fine.

PR #121, opened 2019-03-07
WIP: Cleanup & linked-list implementation

timholy:
A super-early preview to see if this organization, naming, and design makes sense. Perhaps pay special attention to `src/types.jl`.


timholy:
I think I did this wrong. My proposal:

- @KristofferC and anyone else chimes in about whether this file organization makes sense
- I close this and start from a fresh master
- I move code around without renaming or reimplementing anything
- Check to make sure tests still pass
- Submit and quickly merge the PR.

Then merge conflicts will not be a problem except over a very short window. I can tackle the linked-list implementation as a separate, slower PR.

KristofferC:
Seems like a nice structure to me! The only small question I have is whether the `show` methods are better off just after where the type is defined or not. But it is a minor thing.

timholy:
Great. I wonder about that too. We might have some more extensive show methods (eventually I'll bring in #34 and some code from Debugger), but I'll explore leaving them in place for now.

Also, anyone have any big pending PRs? If so better get 'em in. If I haven't heard by 13:15UTC I will begin the reorganization.

PR #124, opened 2019-03-09
Linked-list implementation

timholy:
Here's a preview of the new API. It's feeling a lot cleaner, partly thanks to the linked-list suggestion but also just because it's good to go through the whole codebase everyone once in a while and do some housekeeping.

I got interrupted partway through; I think the conversion is mostly done (not yet touched: `commands.jl`), but I haven't yet tried to compile the package or run tests. ~~I suspect we want a `truncate!` utility and have `handle_err` call that.~~

I'm going to be traveling and visiting family today and tomorrow, so I'm posting this to give folks a chance to look it over and comment. Anyone should feel free to push to this branch, if they are feeling ambitious. But please submit any changes by this evening, as I may pick it up again early tomorrow morning.


timholy:
OK, this is almost over the finish line. Almost every test would pass were it not for the `@assert is_leaf(frame)` I added to `step_expr!`. It's catching some logical error in the handling of exceptions. But this may be the only bug that needs fixing before tests will pass.

Still to do:
- [ ] fix the bug
- [ ] update the docstrings and get doc tests passing
- [ ] play with it a bit to see if we like it.

Bringing over more functionality from Debugger.jl can wait for the next PR.

timholy:
There is a bit of a performance regression: on the `summer` example it has gone from ~15us to ~25us per iteration.

src/commands.jl
```diff
@@ -0,0 +1,183 @@
...
+        pc = frame.pc
+        if isassign(frame, pc)
+            lhs = getlhs(pc)
+            do_assignment!(frame, lhs, ret)
+        end
+        pc += 1
+        frame.pc = pc
+        shouldbreak(frame) && return BreakpointRef(frame.framecode, pc)
+    end
+    return frame
```
KristofferC: Should this return `ret`?

src/commands.jl
```diff
@@ -0,0 +1,183 @@
...
+    return pc
+end
+next_until!(predicate, frame::Frame, istoplevel::Bool=false) =
+    next_until!(predicate, finish_and_return!, frame, istoplevel)
+
+next_call!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false) =
+    next_until!(is_call_or_return, recurse, frame, istoplevel)
+next_call!(frame::Frame, istoplevel::Bool=false) = next_call!(finish_and_return!, frame, istoplevel)
+
+function maybe_next_call!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false)
```
KristofferC:
```suggestion
"""
    maybe_next_call!(predicate, frame, istoplevel=false)

Return the current statement of `frame` if it is a `:return` or `:call` expression. Otherwise, step through the statements of `frame` until the next `:return` or `:call` expression.
"""
function maybe_next_call!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false)
```

src/interpret.jl
```diff
@@ -417,42 +419,42 @@ function _step_expr!(stack, frame, @nospecialize(node), pc::JuliaProgramCounter,
...
+                    if isa(recurse, Compiled)
+                        finish!(recurse, newframe, true)
                     else
-                        push!(stack, frame)
-                        finish!(stack, newframe, true)
-                        pop!(stack)
-                        push!(junk, newframe)  # rather than going through GC, just re-use it
+                        newframe.caller = frame
+                        frame.callee = newframe
+                        finish!(recurse, newframe, true)
```
KristofferC: Is there a need to set `frame.callee` here in case `finish!` throws?

src/commands.jl
```diff
@@ -0,0 +1,183 @@
...
+        pc += 1
+        frame.pc = pc
+        shouldbreak(frame) && return BreakpointRef(frame.framecode, pc)
+    end
+    return frame
+end
+finish_stack!(frame::Frame, istoplevel::Bool=false) = finish_stack!(finish_and_return!, frame, istoplevel)
+
+"""
+    next_until!(predicate, recurse, frame, istoplevel=false)
```
KristofferC: Is it worth saying that `pc` is returned here (so you know e.g. that this can return a breakpoint).

src/types.jl
```diff
@@ -165,9 +225,13 @@ struct Unassigned end
 
 A reference to a breakpoint at a particular statement index `stmtidx` in `framecode`.
 If the break was due to an error, supply that as well.
+
+Commands that execute complex control-flow may also return a `BreakpointRef` to indicate
```
KristofferC: Perhaps an example of this would help.

src/interpret.jl
timholy: It is set on the line above that

timholy:
Tests pass and the performance is back!

src/commands.jl
```diff
@@ -0,0 +1,190 @@
...
+
+Run `frame` until execution terminates. `ret` is either `nothing` (if execution terminates
+when it hits a `return` statement) or a reference to a breakpoint.
+In the latter case, `leaf(frame)` returns the frame in which it hit the breakpoint.
+
+`recurse` controls call evaluation; `recurse = Compiled()` evaluates :call expressions
+by normal dispatch, whereas the default `recurse = finish_and_return!` uses recursive interpretation.
+"""
+function finish!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false)
+    local pc
```
KristofferC: Is this line needed?

src/interpret.jl
KristofferC: I meant a need to set it to `nothing`. But perhaps you made `handle_err` handle that now.

src/utils.jl
```diff
@@ -1,63 +1,74 @@
...
-    reverse!(vars)
-    if frame.code.scope isa Method
-        syms = sparam_syms(frame.code.scope)
-        for i in 1:length(syms)
-            push!(vars, Variable(frame.sparams[i], syms[i], true))
-        end
-    end
-    return vars
-end
+# Note: to avoid dynamic dispatch, many of these are coded as a single method using isa statements
```
KristofferC: Won't the methods be specialized on anyway? So e.g. `scopeof` would be specialized and dynamically dispatched to so that the method `scopeof(::Module) = true` is generated after constant folding. So, is `@nospecialize` needed to prevent dynamic dispatch?

src/utils.jl
```diff
@@ -94,3 +123,171 @@ function is_doc_expr(ex)
...
+
+"""
+    local_variables = locals(frame::Frame)::Vector{Variable}
+
+Return the local variables as a vector of `Variable`[@ref].
+"""
+function locals(frame::Frame)
+    vars = Variable[]
+    data, code = frame.framedata, frame.framecode
+    for (sym, idx) in data.last_reference
```
KristofferC: Not related to the PR but one thing I thought about here is that since `data.last_reference` is a `Dict` the order is unspecified. So perhaps we should iterate through the slot symbols, use a Set to keep track of what symbol we have added, but look up the value of that symbol in `data.last_reference`.

src/interpret.jl
timholy: I see. I *think* it's taken care of in `handle_err`. I think we should leave that `@assert` in place in `step_expr!`, though, just to be sure.

src/utils.jl
timholy: If it can resolve it by union-splitting then dynamic dispatch won't harm anything. It's cases where it doesn't use union-splitting that we have to worry about. But you're right, it might be better to mark these anyway.

src/utils.jl
timholy: I'll fix it here.

I'm also adding an `O(N^2)` performance note to `handle_err`; I propose we not worry about it until we detect a problem in practice.

src/utils.jl
KristofferC: Sure but in the union splitting case, writing the methods out in a dispatch fashion (not using `isa`) would be equivalent, no? So my point was, it is only in the case of non-specialization that the comment on this line is true?

timholy:
OK, since @KristofferC and @pfitzseb both approve, will merge as soon as tests pass. Then:
- move `maybe_step_through_wrapper`, a variant of `execute_command`, and possibly other functions from Debugger to here
- register the package

KristofferC:
I think what would be most useful to have here from Debugger.jl is the functionality to step into a call https://github.com/JuliaDebug/Debugger.jl/blob/4d9c0fc99b09ffdcf8ae1b4df8e6612ca0924884/src/commands.jl#L72-L131 and to return from a frame https://github.com/JuliaDebug/Debugger.jl/blob/4d9c0fc99b09ffdcf8ae1b4df8e6612ca0924884/src/commands.jl#L1-L29

PR #126, opened 2019-03-11
Implement core debug API here instead of Debugger.jl

timholy:
I haven't really played much with Debugger.jl. Porting over some of its functionality is proving to be a useful exercise, as several of these commits generalize JuliaInterpreter in useful ways or detect problems that might not have surfaced so readily otherwise.

This is a WIP; I should probably transition to other things now. If folks want to chip in, I stopped [here](https://github.com/JuliaDebug/Debugger.jl/blob/4d9c0fc99b09ffdcf8ae1b4df8e6612ca0924884/test/stepping.jl#L93). Otherwise I'll pick this up again tomorrow.

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -1,8 +1,8 @@
 """
-    ret = finish!(recurse, frame, istoplevel=false)
-    ret = finish!(frame, istoplevel=false)
+    pc = finish!(recurse, frame, istoplevel=false)
+    pc = finish!(frame, istoplevel=false)
 
-Run `frame` until execution terminates. `ret` is either `nothing` (if execution terminates
+Run `frame` until execution terminates. `pc` is either `nothing` (if execution terminates
 when it hits a `return` statement) or a reference to a breakpoint.
 In the latter case, `leaf(frame)` returns the frame in which it hit the breakpoint.
 
@@ -22,8 +22,8 @@ finish!(frame::Frame, istoplevel::Bool=false) = finish!(finish_and_return!, fram
     ret = finish_and_return!(recurse, frame, istoplevel::Bool=false)
     ret = finish_and_return!(frame, istoplevel::Bool=false)
 
-Call [`JuliaInterpreter.finish!`](@ref) and pass back the return value. If execution
-pauses at a breakpoint, the reference to the breakpoint is returned.
+Call [`JuliaInterpreter.finish!`](@ref) and pass back the return value `ret`. If execution
+pauses at a breakpoint, `ret` is the reference to the breakpoint.
 """
 function finish_and_return!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false)
     pc = finish!(recurse, frame, istoplevel)
@@ -33,26 +33,29 @@ end
 finish_and_return!(frame::Frame, istoplevel::Bool=false) = finish_and_return!(finish_and_return!, frame, istoplevel)
 
 """
-    bpref = dummy_breakpoint(recurse, frame::Frame)
+    bpref = dummy_breakpoint(recurse, frame::Frame, istoplevel)
 
-Return a fake breakpoint. This can be useful as the `recurse` argument to `evaluate_call!`
-(or any of the higher-order commands) to ensure that you return immediately after stepping
-into a call.
+Return a fake breakpoint. `dummy_breakpoint` can be useful as the `recurse` argument to
+`evaluate_call!` (or any of the higher-order commands) to ensure that you return immediately
+after stepping into a call.
 """
-dummy_breakpoint(@nospecialize(recurse), frame::Frame) = BreakpointRef(frame.framecode, 0)
+dummy_breakpoint(@nospecialize(recurse), frame::Frame, istoplevel) = BreakpointRef(frame.framecode, 0)
 
 """
-    ret = finish_stack!(recurse, frame, istoplevel=false)
-    ret = finish_stack!(frame, istoplevel=false)
+    ret = finish_stack!(recurse, frame, rootistoplevel=false)
+    ret = finish_stack!(frame, rootistoplevel=false)
 
 Unwind the callees of `frame`, finishing each before returning to the caller.
-`frame` itself is also finished
-If execution hits a breakpoint, `ret` will be a reference to the breakpoint.
+`frame` itself is also finished. `rootistoplevel` should be true if the root frame is top-level.
+
+`ret` is typically the returned value. If execution hits a breakpoint, `ret` will be a
+reference to the breakpoint.
 """
-function finish_stack!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false)
+function finish_stack!(@nospecialize(recurse), frame::Frame, rootistoplevel::Bool=false)
     frame0 = frame
     frame = leaf(frame)
     while true
+        istoplevel = rootistoplevel && frame.caller === nothing
         ret = finish_and_return!(recurse, frame, istoplevel)
         isa(ret, BreakpointRef) && return ret
         frame === frame0 && return ret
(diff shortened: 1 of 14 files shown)


src/commands.jl
```diff
@@ -177,13 +197,138 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
...
+        pc = maybe_next_call!(recurse, frame, istoplevel)
+        (isa(pc, BreakpointRef) || pc === nothing) && return maybe_reset_frame!(recurse, frame, pc, rootistoplevel)
+        stmt0 = stmt = pc_expr(frame, pc)
+        if isexpr(stmt, :(=))
+            stmt = stmt.args[2]
+        end
+        ret = evaluate_call!(dummy_breakpoint, frame, stmt; enter_generated=enter_generated)
+        isa(ret, BreakpointRef) && return maybe_reset_frame!(recurse, frame, ret, rootistoplevel)
+        maybe_assign!(frame, stmt0, ret)
+        frame.pc = ret + 1
```
KristofferC: There is some issue here when evaluating builtins

```
julia> f(x) = x^2

julia> frame = JuliaInterpreter.enter_call(f, 3)
Frame for f(x) in Main at REPL[38]:1
  1 1  1 ─ %1 = (Core.apply_type)(Val, 2)
  2 1  │   %2 = (%1)()
  3 1  │   %3 = (Base.literal_pow)(^, x, %2)
⋮
x = 3

julia> frame, bp = JuliaInterpreter.debug_command(JuliaInterpreter.finish_and_return!, frame, "s")
ERROR: MethodError: no method matching +(::Type{Val{2}}, ::Int64)
```

KristofferC:
Out of curiosity, what is the reason not to use recursive interpretation by default here? That it is more expensive and only those with breakpoint support will need it?

timholy:
Actually, that was dumb; it should be recursive, precisely because of breakpoint support. I was influenced by history and the fact that we know JuliaInterpreter still has things it doesn't handle properly. But breakpoints mean that of course everything should be recursive by default.

src/commands.jl
KristofferC: Another when the expression is a return statement:

```
> f(x) = x

> fr = JuliaInterpreter.enter_call(f, 2)
Frame for f(x) in Main at REPL[8]:1
  1 1  1 ─     return x
x = 2

> debug_command(fr, "s")
(frame.framedata).exception_frames = Int64[]
ERROR: MethodError: objects of type Int64 are not callable
Stacktrace:
 [1] #prepare_call#12(::Bool, ::Function, ::Any, ::Array{Any,1}) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:216
```

timholy:
Almost there. c5c3ec8 is scary---could have big implications for #13.

KristofferC:
Should the following return a break on error breakpoint?

```jl
julia> using JuliaInterpreter

julia> f() = undef_func()

julia> JuliaInterpreter.break_on_error[] = true
true

julia> f() = undef_func()

julia> frame = JuliaInterpreter.enter_call(f)
Frame for f() in Main at REPL[7]:1
  1 1  1 ─ %1 = (Main.undef_func)()
  2 1  └──      return %1

julia> debug_command(frame, "s")
ERROR: UndefVarError: undef_func not defined
Stacktrace:
 [1] lookup_var at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:9 [inlined]
 [2] macro expansion at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:56 [inlined]
 [3] maybe_evaluate_builtin(::Frame, ::Expr, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\builtins-julia1.1.jl:29
```

KristofferC:
Another one

```jl
f22() = string(:(a+b))

function step_through(f, args...; kwargs...)
    fr = JuliaInterpreter.enter_call(f, args...; kwargs...)
    while true
        ret = JuliaInterpreter.debug_command(fr, "s")
        ret === nothing && break
        fr, pc = ret
    end
end

step_through(f22)
ERROR: MethodError: objects of type Bool are not callable
Stacktrace:
 [1] #prepare_call#12(::Bool, ::Function, ::Any, ::Array{Any,1}) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:216
 [2] #prepare_call at .\none:0 [inlined]
 [3] #get_call_framecode#32(::Bool, ::Function, ::Array{Any,1}, ::JuliaInterpreter.FrameCode, ::Int64) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\localmethtable.jl:53
 [4] #get_call_framecode at .\none:0 [inlined]
 [5] #evaluate_call_recurse!#37(::Bool, ::Function, ::Any, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:203
 [6] #evaluate_call_recurse! at .\none:0 [inlined]
 [7] #evaluate_call!#39 at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:237 [inlined]
 [8] #evaluate_call! at .\none:0 [inlined]
 [9] debug_command(::Any, ::Frame, ::String, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:336
 [10] debug_command at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:358 [inlined] (repeats 2 times)
```

KristofferC:
One way of fixing https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/126#discussion_r264610347 is to just piggy back on `"finish"`:

```diff
diff --git a/src/commands.jl b/src/commands.jl
index a1b3d3b..09649a1 100644
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -333,6 +333,9 @@ function debug_command(@nospecialize(recurse), frame::Frame, cmd::AbstractString
             if isexpr(stmt, :(=))
                 stmt = stmt.args[2]
             end
+            if isexpr(stmt, :return)
+                return debug_command(recurse, frame, "finish", rootistoplevel)
+            end
             ret = evaluate_call!(dummy_breakpoint, frame, stmt; enter_generated=enter_generated)
             isa(ret, BreakpointRef) && return maybe_reset_frame!(recurse,
frame, ret, rootistoplevel)
             maybe_assign!(frame, stmt0, ret)
```

Not sure that is the cleanest way but I'll make a PR just to get some tests in.

src/commands.jl
timholy: fixed

KristofferC:
When calling the command `n` (next line) is there anyway to know if the returned breakpoint is a "real" breakpoint or comes from the fact that `next_line!` changed frame?

For example:

```
julia> frame = JuliaInterpreter.enter_call_expr(:($(optional)()))
Frame for optional() in Main at REPL[29]:2
  1 2  1 ─ %1 = (sin)(1)
  2 2  │   %2 = (#self#)(%1)
  3 2  └──      return %2

julia> fr, pc =  JuliaInterpreter.debug_command(frame, "n")
(Frame for optional(n) in Main at REPL[29]:2
  1 2  1 ─      x = (asin)(n)
  2 3  │   %2 = (cos)(x)
  3 3  └──      return %2
n = 0.841471
x = 1.0
caller: optional() in Main at REPL[29]:2, breakpoint(optional(n) in Main at REPL[29]:2, line 3))

julia> pc.stmtidx
2
```

`pc` is a breakpoint with a `stmtidx != 0` so it is indistinguishable from a breakpoint added by a user, no?

src/construct.jl
```diff
@@ -20,8 +20,8 @@ rather than recursed into via the interpreter.
 """
 const compiled_methods = Set{Method}()
 
-const junk = FrameData[]      # to allow re-use of allocated memory (this is otherwise a bottleneck)
-recycle(frame) = push!(junk, frame.framedata)
+const junk = IdDict{FrameData,Nothing}()      # to allow re-use of allocated memory (this is otherwise a bottleneck)
```
KristofferC: `Base.IdSet`?

KristofferC:
I think in Debugger.jl when returning from a frame, we kept returning out of wrapper functions:

https://github.com/JuliaDebug/Debugger.jl/blob/20e746a2e81c4bd535e4f87b074b5f4e211e33eb/src/commands.jl#L25-L27.

I'm not sure what the correct behavior is but when interactively working I guess everyone except ~5 people would want to step through the wrappers.

timholy:
> pc is a breakpoint with a stmtidx != 0 so it is indistinguishable from a breakpoint added by a user, no?

Correct. Can you clarify what we need to distinguish them for?

timholy:
>  think in Debugger.jl when returning from a frame, we kept returning out of wrapper functions:

That does make sense. I don't like adding the info to the `framecode`, if there's another way.

KristofferC:
Just a UI thing, if we break for a breakpoint I want to show to the user that we hit a breakpoint. If we step into a function it is expected that we break so no need to show an info message.

test/debug.jl
```diff
@@ -0,0 +1,236 @@
...
+        try
+            JuliaInterpreter.break_on_error[] = true
+            fr = JuliaInterpreter.enter_call(f_outer)
+            fr, pc = debug_command(JuliaInterpreter.finish_and_return!, fr, "finish")
+            @test fr.framecode.scope.name == :error
+
+            fundef() = undef_func()
+            frame = JuliaInterpreter.enter_call(fundef)
+            fr, pc = debug_command(frame, "s")
+            @test isa(pc, BreakpointRef)
```
KristofferC:
```suggestion
            @test isa(pc, BreakpointRef)
            @test pc.err isa MethodError
```

timholy:
Sure, let's fix that. Can I ask you to tackle that? I desperately need to pivot to other things. I will merge this as soon as tests "pass," unless you object.

KristofferC:
Sure! I'll figure something out.

KristofferC:
All green with the exception of some docstrings not being in the generated docs. I pushed a commit adding them which should make us all green.

Merge at will!

KristofferC:
Boom! Nice job Tim!

timholy:
Thanks, you too! Could not have done this without you.

PR #127, opened 2019-03-11
Debug commands: implement unwinding of stack on thrown exception

KristofferC:


KristofferC:
Hm, presumably this need to be done for `finish` as well? And also for `s` (when you step your way all the way into `throw`)?

src/commands.jl
```diff
@@ -272,6 +272,22 @@ function maybe_reset_frame!(@nospecialize(recurse), frame::Frame, @nospecialize(
...
+# or rethrow the error
+function unwind_exception(frame::Frame, exc)
+    while frame !== nothing
+        if !isempty(frame.framedata.exception_frames)
+            # Exception caught
+            frame.pc = frame.framedata.exception_frames[end]
+            frame.framedata.last_exception[] = exc
+            return frame
+        end
+        frame = frame.caller
```
timholy: Should we set `frame.callee = nothing`? If it gets caught and execution continues, that `@assert` in `step_expr!` will be trouble.

timholy:
Perhaps basically the whole method body should go in the try/catch? I haven't thought deeply about this yet, you're way ahead of me here.

src/commands.jl
timholy: Also could probably `recycle`? Do we need to check `break_on_err`? This is starting to feel a bit like `handle_err`.

KristofferC: `break_on_error` would have made this not throw in the first place, no? It would have looked down the stack, found that something would catch the exception and then instead returned breakpoint.

Yes, setting the `callee` is needed.

src/commands.jl
```diff
@@ -294,40 +312,48 @@ or one of the 'advanced' commands
...
+        if cmd == "c"
+            r = root(frame)
+            ret = finish_stack!(recurse, r, rootistoplevel)
+            return isa(ret, BreakpointRef) ? (leaf(r), ret) : nothing
+        end
+        cmd == "finish" && return maybe_reset_frame!(recurse, frame, finish!(recurse, frame, istoplevel), rootistoplevel)
+    catch err
+        frame = unwind_exception(frame, err)
+        if cmd == "c"
+            return debug_command(recurse, frame, "c", istoplevel)
```
KristofferC: Continue is special because then we just want to keep going.

KristofferC:
I tried adding the `recycle` (it is commented out now) but I have some problems. Without the `recycle` I can run:

```jl
   function f_exc_outer()
            try 
                f_exc_inner()
            catch err;
                err_caught[1] = err
            end
            x = 1 + 1
            return x
        end
        f_exc_inner() = error()

        err_caught = Any[nothing]
        fr = JuliaInterpreter.enter_call(f_exc_outer)
        fr, pc = debug_command(fr, "s")
        debug_command(fr, "c")
        @test get_return(root(fr)) == 2
        @test first(err_caught) isa ErrorException
        @test stacklength(root(fr)) == 1
```

twice and it passes both time. If I add the recycle it fails the second time by throwing the exception:

```
ERROR: 
Stacktrace:
 [1] error() at .\error.jl:42
 [2] #evaluate_call_compiled!#36(::Bool, ::Function, ::Compiled, ::Frame, ::Expr) at
C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:186
```

If I `empty!` the `junk` then it passes again so it has something to do with the caching of frames. If I show the `exception_frames` during the unwinding I get on the first run:

```
(frame.framedata).exception_frames = Int64[]
(frame.framedata).exception_frames = [6]
```

but on the second run:

```
(frame.framedata).exception_frames = Int64[]
(frame.framedata).exception_frames = Int64[]
```

timholy:
Interesting. I'll merge this and then take a look at it myself.

PR #131, opened 2019-03-13
fix vararg unionall handling

### LINKED ISSUE #130: MWE for failing test for errorshow.jl or "Varargs, the undead of the Julia Type system"
It turns out that the failure in errorshow.jl has to do with handling
Varargs.  Due to https://github.com/JuliaLang/julia/issues/30995, there is
a check in prepare_call in construct.jl that will punt if any of the args
is a Vararg.  But that is the problem, there is not an easy way to test the
different forms of Vararg that come through.

The test is currently ```elseif any(x->isa(x, Type) && x <: Vararg, allargs)```
however, one form of that comes through is ```x = Vararg{AbstractString,N} where {N}```
and ```isa(x, Type)``` tests true but ```x <: Vararg``` is false. ```typeof(x)``` returns
a ```UnionALL```.  Another arg that comes through is ```Vararg{AbstractString,N}```, that
is,  there is no where clause?!  I don't understand how that is possible.

I could not construct a test to get these two, and only these two, using type checking.
For the purposes of testing, I used the brute force approach of
```elseif any(x->isa(x, Type) && occursin("Vararg", string(x)), allargs)```  The unit
tests for JuliaInterpreter run OK with this and the following MWE passes, but there should
be a better way to do that, so I did not prepare a PR.

```
using JuliaInterpreter
include("utils.jl")

module m
    method_c1(x::Float64, s::AbstractString...) = true
    buf = IOBuffer()
    me = Base.MethodError(method_c1,(1, 1, ""))
end

ex = quote
    Base.show_method_candidates(buf, me)
end

# TypeError: in Type, in parameter, expected Type, got Vararg{AbstractString,N} where N

frame = JuliaInterpreter.prepare_thunk(m, ex)
JuliaInterpreter.finish_and_return!(frame)
```

pfitzseb:
Fixes #130.

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -206,7 +206,8 @@ function prepare_call(@nospecialize(f), allargs; enter_generated = false)
     # Can happen for thunks created by generated functions
     if isa(f, Core.Builtin) || isa(f, Core.IntrinsicFunction)
         return nothing
-    elseif any(x->isa(x, Type) && x <: Vararg, allargs)
+    elseif any(x -> isa(x, Type) &&
+                    (x <: Vararg || (typeof(x) in (UnionAll, DataType) && nameof(x) == :Vararg)), allargs)
         return nothing  # https://github.com/JuliaLang/julia/issues/30995
     end
     argtypes = Tuple{map(_Typeof,allargs)...}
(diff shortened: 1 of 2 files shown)


timholy:
Lovely, thanks!

PR #133, opened 2019-03-13
add support for in-source breakpoints

### LINKED ISSUE #125: RFC: Allow access to the interpreter state from interpreted code
It could be convenient in some situations to be able to access the interpreter state (most relevant is the current frame) from the code that itself is being interpreted. 

This could be done (modulo macro hygiene) by something like

```julia
is_interpreted() = false
current_frame() = nothing

macro interpreted(ex)
    return :(
        if is_interpreted()
            frame = current_frame()
            $ex
        else
             nothing
        end
    )
end
```

One step of creating the `Frame` would then be to replace `is_interpreted()` with `true` and inject a reference to itself instead of `current_frame()`.

One could then do things like

```jl
function f(x)
    print("hello")
    @interpreted print("I am being interpreted, pc currently at $(frame.pc)")
end
```

which could aid in debugging the interpreter itself but might also allow other cool stuff.

There are some considerations:

- Are the use cases strong enough to implement it. I am thinking that e.g. file:line breakpoint support (without Revise) would be possible with this feature fairly trivially with a `@breakpoint` macro that inserts a breakpoint into the current frame.
- With this you can in theory do things that are generally not allowed in julia due to not being optimizable, like local evals. I think that we don't really want people to use the interpreter to enable running non-standard julia code, but it is perhaps unlikely that people will use the interpreter for this.

KristofferC:
~~This is slightly WIP because ideally we would remove the breakpoint marker from the code completely. The reason I haven't done that yet is because I am not familiar enough (but working on becoming) with all of the invariants that need to be kept when modifying the `.code` array (SSA renumbering etc).~~

I think we can close #125 with this since this fixes my original motivation of having that feature.


pfitzseb:
Take a look at https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/91911967124e208b96387c8b9bca201064256a46/src/optimize.jl#L148.

KristofferC:
Yeah, that's where my editor is at the moment ;)

KristofferC:
I think this is correct but it is a bit messy. Review carefully :)

KristofferC:
Ok, should be reviewable.

KristofferC:
I wonder if the expression  equality check could be expensive to run on all statements. Would be good to check.

src/breakpoints.jl
```diff
@@ -281,3 +281,14 @@ macro breakpoint(call_expr, args...)
...
 end
+
+const __BREAKPOINT_MARKER__ = nothing
+
+"""
+    @bp
+
+Insert a breakpoint at a location in the source code.
+"""
+macro bp()
```
timholy: Add an optional condition?

src/types.jl
```diff
@@ -71,8 +71,34 @@ function FrameCode(scope, src::CodeInfo; generator=false, optimize=true)
...
+    breakpoint_pcs = Int[]
+    for (i, pc_expr) in enumerate(src.code)
+        if pc_expr == :($(QuoteNode(getproperty))($JuliaInterpreter, :__BREAKPOINT_MARKER__))
+            push!(breakpoint_pcs, i)
+        end
+    end
     breakpoints = Vector{BreakpointState}(undef, length(src.code))
+    if !isempty(breakpoint_pcs)
+        ssalookup = collect(1:length(src.codelocs))
+        # Renumber SSA values in preparation for deleting breakpoint markers
```
timholy: Could replace the statement with `nothing` so numbers don't change?

src/breakpoints.jl
KristofferC: I thought about this but is `@bp i > 0` much easier than `i > 0 && @bp`?

src/types.jl
KristofferC: Didn't know that was allowed. Is it possible to set a breakpoint on a `nothing` statement?

KristofferC: The reason I did the whole `deleteat!` for the statements was so that the debugger wouldn't stop at the `@bp` statement and print that it would run `nothing`. However, I can just work around that on the Debugger side: https://github.com/JuliaDebug/Debugger.jl/pull/61.

KristofferC:
I pushed a new commit without the bp statement deletion and as seen by the changes to the test, this means that we now stop at the actual `@bp` statement here in the interpreter (which can be worked around in a front end if it wants to as in https://github.com/JuliaDebug/Debugger.jl/pull/61

src/breakpoints.jl
KristofferC: For example

```jl
julia> function f(x)
      println("hello")
      x > 5 && @bp
      return x
  end;

julia> @interpret f(2)
hello
2

julia> @interpret f(7)
hello
(Frame for f(x) in Main at REPL[41]:2
⋮
  2 3  │   %2 = (>)(x, 5)
  3 3  └──      goto #3 if not %2
b 4 3  2 ─      nothing
  5 3  └──      goto #3
  6 4  3 ┄      return x
x = 7, breakpoint(f(x) in Main at REPL[41]:2, line 3))
```

KristofferC:
It seems the expression check is fast.

src/breakpoints.jl
timholy: That works great, good point! Just need to make sure it's documented; obviously I didn't think of it, so others may not either.

src/types.jl
timholy: I'm fine with either approach, whichever you prefer.

KristofferC:
Thanks for review!

PR #139, opened 2019-03-14
move builtin generation logic to precompilation time

KristofferC:
The current build system has the problem that if someone adds JuliaInterpreter on two julia versions, the second add will not build JuliaInterpreter (since it is already downloaded and assumed built). This will then give an error when loading it that it misses the file.

Instead, just run this on precompilation time if we cant find the file.

The negative is that we need to load the code for generating the builtin file every time we load the package but from my measurements it is within noise.

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -22,12 +22,11 @@ end
 
 const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(Int(VERSION.minor)).jl")
 
-if ccall(:jl_generating_output, Cint, ()) == 1
-    @info "Generating builtins for this julia version..."
-    gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
-    run(`$(Base.julia_cmd()) $gen_builtins_file`)
-    include_dependency(gen_builtins_file)
-end
+@info "Generating builtins for this julia version..."
+gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
+# Run as separate command to prevent including the generate_builtins into the precompile cache
+run(`$(Base.julia_cmd()) $gen_builtins_file`)
+include_dependency(gen_builtins_file)
 
 include("types.jl")
 include("utils.jl")
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -22,12 +22,11 @@ end
 
 const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(Int(VERSION.minor)).jl")
 
-if !isfile(BUILTIN_FILE)
+if ccall(:jl_generating_output, Cint, ()) == 1
     @info "Generating builtins for this julia version..."
-    include("generate_builtins.jl")
-    open(BUILTIN_FILE, "w") do io
-        generate_builtins(io)
-    end
+    gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
+    run(`$(Base.julia_cmd()) $gen_builtins_file`)
+    include_dependency(gen_builtins_file)
 end
 
 include("types.jl")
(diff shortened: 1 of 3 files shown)


src/JuliaInterpreter.jl
```diff
@@ -20,12 +20,22 @@ module CompiledCalls
 # This module is for handling intrinsics that must be compiled (llvmcall)
 end
 
+const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(Int(VERSION.minor)).jl")
+
+if !isfile(BUILTIN_FILE)
```
timholy: Should we also check `Base.stale_cachefile("generate_builtins.jl", cachefile)`? I'm not sure I'm thinking about the logic of this properly, but it seems to make sense to rebuild it if the generator changes.

If possible, it would be lovely to preserve the ability to edit the "builtins-" file for the purposes of debugging. If I'm thinking about it correctly, it seems that checking the mtime on "generate_builtins.jl" but *not* "builtins-" will preserve that capability.

KristofferC: Storing the `mtime` of `generate_builtins.jl` and rerun the build script if it changed makes sense.

> If possible, it would be lovely to preserve the ability to edit the "builtins-" file for the purposes of debugging.

The changes here shouldn't affect that?

timholy: Right, that's what I'm expecting too. But sometimes the chain of dependencies makes my head hurt.

timholy: I don't think you have to store the `mtime`, do you? That comparison should be run for you by `stale_cachefile`.

timholy: Maybe another approach: put this logic inside a `ccall(:jl_generating_output, Cint, ()) == 1` and add the `generate_builtins.jl` as an `include_dependency`?

KristofferC: Something like this then?

src/JuliaInterpreter.jl
```diff
@@ -20,12 +20,21 @@ module CompiledCalls
 # This module is for handling intrinsics that must be compiled (llvmcall)
 end
 
+const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(Int(VERSION.minor)).jl")
+
+if ccall(:jl_generating_output, Cint, ()) == 1
```
KristofferC: Is this check really needed?

pfitzseb:
Hm, I think we'll need a fallback if someone runs Julia with `--compiled-modules=no`, right? With the current implementation the `BUILTIN_FILE` would never be generated in that case.

KristofferC:
Yeah, that was my point with https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/139#discussion_r265523749. Should just be able to remove that.

pfitzseb:
Hm but then the builtin file would be regenerated every time the module is loaded, which is a safer but more annoying default, no?

KristofferC:
Yes, every time it is loaded when running with `--compiled-modules=no` because `generate_builtins` might have changed.

src/JuliaInterpreter.jl
```diff
@@ -20,12 +20,20 @@ module CompiledCalls
 # This module is for handling intrinsics that must be compiled (llvmcall)
 end
 
+const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(Int(VERSION.minor)).jl")
+
+@info "Generating builtins for this julia version..."
+gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
+# Run as separate command to prevent including the generate_builtins into the precompile cache
+run(`$(Base.julia_cmd()) $gen_builtins_file`)
```
timholy:
```suggestion
run(`$(Base.julia_cmd()) --startup-file=no $gen_builtins_file`)
```

KristofferC: Good catch!

PR #140, opened 2019-03-14
An export-worthy API for break_on_error

timholy:
Some day we hope to have more possibilities for triggering automatic breakpoints (issue #102). For now, let's at least establish an extensible user-level API.

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -130,6 +130,40 @@ function remove()
 end
 
 """
+    break_on(states...)
+
+Turn on automatic breakpoints when any of the conditions described in `states` occurs.
+The supported states are:
+
+- `:error`: trigger a breakpoint any time an uncaught exception is thrown
+"""
+function break_on(states::Vararg{Symbol})
+    for state in states
+        if state == :error
+            break_on_error[] = true
+        else
+            throw(ArgumentError(string("unsupported state :", state)))
+        end
+    end
+end
+
+"""
+    break_off(states...)
+
+Turn off automatic breakpoints when any of the conditions described in `states` occurs.
+See [`break_on`](@ref) for a description of valid states.
+"""
+function break_off(states::Vararg{Symbol})
+    for state in states
+        if state == :error
+            break_on_error[] = false
+        else
+            throw(ArgumentError(string("unsupported state :", state)))
+        end
+    end
+end
+
+"""
     breakpoint(f, sig)
     breakpoint(f, sig, line)
     breakpoint(f, sig, condition)
--- a/test/breakpoints.jl
+++ b/test/breakpoints.jl
@@ -130,7 +130,9 @@ end
 
     # break on error
     try
-        JuliaInterpreter.break_on_error[] = true
+        @test_throws ArgumentError("unsupported state :missing") break_on(:missing)
+        @test_throws ArgumentError("inputs must be Symbols, got String") break_on("missing")
+        break_on(:error)
 
         inner(x) = error("oops")
         outer() = inner(1)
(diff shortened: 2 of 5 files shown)


timholy:
I should confess that this is a bit of a pun, and that might deserve some thought. Internally, `break_on_error` uses "on" in the sense of "reason for" ("break because an error was thrown"), whereas this API uses "on" in the sense of "turn on." I'm not sure there has to be concordance between internal and external, especially because eventually the internal may change, but I'm also happy to make further changes.

KristofferC:
Symbols instead of strings seems like it is more often used in situations like this. Personally I really prefer enums for stuff like this but the enums in Julia are unfortunately not very "ergonomic" so symbols for now are probably fine. (I thought about this for the different `debug_command`s as well).

timholy:
I kind of wondered about that. I remember thinking "Symbols look weird" when I first started with Julia, so I most often use strings in public APIs, but I agree that in a situation where they are basically being used as enums, they make sense.

KristofferC:
The only real advantage of Symbols is that you need one extra letter `:` vs `" "` and they are faster to compare since they are interned (and thus just a pointer comparison). But whatever you prefer is fine and to be consistent with `debug_command` we can use Strings.

timholy:
@pfitzseb, any thoughts on Symbols or Strings? If your have a clear preference, that could tip this towards decision.

pfitzseb:
I think I'd prefer something enum like (i.e. `@enum` in a module or just a module that has some singleton types or something).
If it comes down to Strings vs Symbols I'd prefer Symbols though :)

timholy:
Sounds like I should definitely explore enums. Funny story: in LoweredCodeUtils, extracting signatures from LibGit2 was painful because the expansion of `@enum` (which is not small!) revealed many deficiencies that I struggled to overcome. Now I will be actually *spreading* the use of the little buggers :smile:.

timholy:
Switched to Symbols commensurate with https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/160

src/breakpoints.jl
```diff
@@ -129,6 +129,42 @@ function remove()
...
 
+"""
+    break_on(states...)
+
+Turn on automatic breakpoints when any of the conditions described in `states` occurs.
+The supported states are:
+
+- `:error`: trigger a breakpoint any time an uncaught exception is thrown
+"""
+function break_on(states...)
```
KristofferC:
```suggestion
function break_on(states::Vararg{Symbol})
```

maybe?

timholy: Sure, why not.

src/breakpoints.jl
```diff
@@ -129,6 +129,42 @@ function remove()
...
+The supported states are:
+
+- `:error`: trigger a breakpoint any time an uncaught exception is thrown
+"""
+function break_on(states::Vararg{Symbol})
+    for state in states
+        if state == :error
+            break_on_error[] = true
+        else
+            isa(state, Symbol) || throw(ArgumentError("inputs must be Symbols, got $(typeof(state))"))
```
KristofferC: and then

```suggestion
```

test/breakpoints.jl
```diff
@@ -130,7 +130,9 @@ end
 
     # break on error
     try
-        JuliaInterpreter.break_on_error[] = true
+        @test_throws ArgumentError("unsupported state :missing") break_on(:missing)
+        @test_throws ArgumentError("inputs must be Symbols, got String") break_on("missing")
```
KristofferC: This this will be a method error now, so

```suggestion
```

PR #142, opened 2019-03-14
fix interpreting getting env vars on windows

### LINKED ISSUE #141: Divergence between compiler and interpreter with logging macros
```jl
julia> function f()
           x = 1 + 1
           @info "hello"
       end

julia> f()
[ Info: hello

julia> @interpret f()
Exception handling log message: ArgumentError("input string is empty or only contains whitespace")
  module=Main  file=REPL[3]  line=3
  Second exception: ArgumentError("input string is empty or only contains whitespace")
```

KristofferC:
We should probably enable AV heh.

Fixes #141

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -59,6 +59,11 @@ function set_compiled_methods()
     push!(compiled_methods, which(reenable_sigint, Tuple{Function}))
     # Signal-handling in the `print` dispatch hierarchy
     push!(compiled_methods, which(Base.unsafe_write, Tuple{Base.LibuvStream, Ptr{UInt8}, UInt}))
+    # Libc.GetLastError()
+    @static if Sys.iswindows()
+        push!(compiled_methods, which(Base.access_env, Tuple{Function, AbstractString}))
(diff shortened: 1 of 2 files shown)


src/JuliaInterpreter.jl
```diff
@@ -59,6 +59,11 @@ function set_compiled_methods()
...
     # Signal-handling in the `print` dispatch hierarchy
     push!(compiled_methods, which(Base.unsafe_write, Tuple{Base.LibuvStream, Ptr{UInt8}, UInt}))
+    # Libc.GetLastError()
+    @static if Sys.iswindows()
+        push!(compiled_methods, which(Base.access_env, Tuple{Function, AbstractString}))
+        push!(compiled_methods, which(Base._hasenv, Tuple{Vector{UInt16}}))
+    end
 end
 
 function __init__()
```
KristofferC: Are methods not persistent between julia sessions? We really need this in `__init__()`?

timholy: Not certain; if they are compared via pointer I wonder if you'd have to rehash.

Issue #144, opened 2019-03-14
Binary builds cannot step into source code for stdlib functions due to bad location info


KristofferC:
Due to https://github.com/JuliaLang/julia/issues/26314 we are getting spurious location info for functions in the stdlib:

```jl
julia> using LinearAlgebra

julia> import JuliaInterpreter.enter_call

julia> enter_call(norm, [1,2])
Frame for norm(itr) in LinearAlgebra at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.1\LinearAlgebra\src\generic.jl:448
  1 448  1 ─ %1 = (#self#)(itr, 2)
  2 448  └──      return %1
itr = [1, 2]
```

If not too difficult, we could perhaps detect this and rewrite the prefix to `Sys.STDLIB`. Best would of course be to fix it in Base but not sure how hard that is.

pfitzseb:
Should CodeTracking handle this, maybe?

PR #146, opened 2019-03-14
 work around problem with finding stdlib files 

KristofferC:
Commited on top of #145 

Previously we couldnt see the source for stdlib files:

```jl
julia> @enter norm([1,2])
In norm(itr, p) at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.1\LinearAlgebra\src\generic.jl:448
1   1 ── %1  = (isempty)(itr)
2   └───       goto #4 if not %1
3   2 ── %3  = (eltype)(itr)

About to run: (isempty)([1, 2])
1|debug> 
```

Now we can:

```jl
julia> @enter norm([1,2])
In norm(itr, p) at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.1\LinearAlgebra\src\generic.jl:448
448       isempty(itr) && return float(norm(zero(eltype(itr))))
449       if p == 2
450           return norm2(itr)
```

The method printing is still wrong though.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -4,10 +4,10 @@
 uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 
 [[CodeTracking]]
-deps = ["Test", "UUIDs"]
-git-tree-sha1 = "12aa4d41c7926afd7a71af5af603a4d8b94292c2"
+deps = ["InteractiveUtils", "Test", "UUIDs"]
+git-tree-sha1 = "983f5f7a57c604322917ab8bc5b86ba914d3c345"
 uuid = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
-version = "0.3.1"
+version = "0.3.2"
 
 [[Distributed]]
 deps = ["Random", "Serialization", "Sockets"]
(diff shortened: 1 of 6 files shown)


KristofferC:
I pushed a commit to also fixup the printing of the frame.

If this is too "hand-wavy" I'm happy just moving this to the debugger instead.

pfitzseb:
I still kinda feel like CodeTracking's `whereis` should handle this.

KristofferC:
Yeah, I'm inclined to agree. Would be simple to move there. Let's wait and see what Tim thinks.

KristofferC:
I guess we would still need it here for the frame printing and `getfile` but we could call the equivalent of `maybe_find_stdlib_file` from CodeTracking then.

timholy:
https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/144#issuecomment-473212593.

But I'd be happy to have a PR in CodeTracking with this replacement operation! It should be performed [here](https://github.com/timholy/CodeTracking.jl/blob/b3f081a2be0e2ee0dde77b1c3b0323b2d09ee12a/src/CodeTracking.jl#L41).

KristofferC:
https://github.com/timholy/CodeTracking.jl/pull/16

KristofferC:
Updated the PR based on https://github.com/timholy/CodeTracking.jl/pull/16. Note that the Manifest here now refers to that PR branch (to run CI) on it but should not be merged as is.

PR #149, opened 2019-03-15
Ensure that "s" steps through wrappers

timholy:
My sense is that `"s"` should step through wrappers but `"si"` should not. This adds such functionality and tests it.

One thing worth pointing out is that there is a certain amount of "junk" that gets lowered into the *caller*: using the examples in the new tests,

```julia
 julia> Base.uncompressed_ast(first(methods(callopkw2)))
CodeInfo(
1 ─ %1 = (Main.opkw)(0, -1)
└──      return %1
)
```
but
```julia
julia> Base.uncompressed_ast(first(methods(callopkw5)))
CodeInfo(
1 ─ %1 = (:c, :d)
│   %2 = (Core.apply_type)(Core.NamedTuple, %1)
│   %3 = (Core.tuple)(-2, -3)
│   %4 = (%2)(%3)
│   %5 = (Core.kwfunc)(Main.opkw)
│   %6 = (%5)(%4, Main.opkw, 0)
└──      return %6
)
```
Note that we're in the caller of a kwarg function, not the kwarg function itself. For "normal" stepping (i.e., non-advanced commands) it would be lovely to be able to automatically skip through this. The nicest way to do this would presumably be based on https://github.com/JuliaLang/julia/issues/31162---if each frame could also hold line-by-line references to the expression that's being lowered, then you could just go to the last call of the same expression.

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -227,7 +227,7 @@ function maybe_step_through_wrapper!(@nospecialize(recurse), frame::Frame)
     length(stmts) < 2 && return frame
     last = stmts[end-1]
     isexpr(last, :(=)) && (last = last.args[2])
-    is_kw = isa(scope, Method) && startswith(String(Base.unwrap_unionall(scope.sig).parameters[1].name.name), "#kw")
+    is_kw = isa(scope, Method) && startswith(String(Base.unwrap_unionall(Base.unwrap_unionall(scope.sig).parameters[1]).name.name), "#kw")
     if is_kw || isexpr(last, :call) && any(x->x==Core.SlotNumber(1), last.args)
         # If the last expr calls #self# or passes it to an implementation method,
         # this is a wrapper function that we might want to step through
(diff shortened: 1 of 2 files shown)


timholy:
Any objections? Might (temporarily) break the stepping logic of Debugger.

KristofferC:
I think `s` should step through wrappers as well.

KristofferC:
Thinking about it a bit, I kinda disagree with this.

A frontend who wants to step through wrappers could always previously just insert the `maybe_step_through_wrapper!`. However, now, there is no way for a frontend to chose not to step through the wrapper?

In addition, previously, I could record that a frame was a wrapper frame by looking if ``maybe_step_through_wrapper!` returned a new frame. And upon returning to that frame I could then use that information to finish the frame. Now, when I return from a frame it is slightly more difficult to know what frames are wrapper frames (but you can probably do it by looking at the length of the stack)?

timholy:
> However, now, there is no way for a frontend to chose not to step through the wrapper?

You can choose to step via `si`, right?

> Now, when I return from a frame it is slightly more difficult to know what frames are wrapper frames (but you can probably do it by looking at the length of the stack)?

Easier would be `newframe.caller === oldframe`. Probably also want `newframe === oldframe` to handle builtins and the like that don't permit stepping in.

KristofferC:
> You can choose to step via si, right?

Yeah but what if I want to step to the next call but not through a wrapper.

I guess my point is that since this is the most low level of the debugger commands, any "magic" we introduce here will propagate to all users of it, and it is harder to clean away the magic, then to just conditionally apply it yourself.

PR #152, opened 2019-03-15
add functionality to step through macro expansion

KristofferC:
Fixes #152

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -214,8 +214,8 @@ end
 next_line!(frame::Frame, istoplevel::Bool=false) = next_line!(finish_and_return!, frame, istoplevel)
 
 is_macro_expansion(lin::LineInfoNode) = lin.method == Symbol("macro expansion")
-is_macro_expansion(::Any) = false
-function is_macro_expansion(frame)
+is_macro_expansion(::LineNumberNode) = false
+function is_macro_expansion(frame::Frame)
     src = frame.framecode.src
     codeloc = src.codelocs[frame.pc]
     codeloc == 0 && return false
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -213,6 +213,49 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
 end
 next_line!(frame::Frame, istoplevel::Bool=false) = next_line!(finish_and_return!, frame, istoplevel)
 
+is_macro_expansion(lin::LineInfoNode) = lin.method == Symbol("macro expansion")
+is_macro_expansion(::Any) = false
+function is_macro_expansion(frame)
+    src = frame.framecode.src
+    codeloc = src.codelocs[frame.pc]
+    codeloc == 0 && return false
+    linfo = src.linetable[codeloc]
+    return is_macro_expansion(linfo)
+end
+
+"""
+    pc = maybe_step_through_macro_expansion!(recurse, frame)
+    pc = maybe_step_through_macro_expansion!(frame)
+
+Execute the current statement. Then step through the statements of `frame` until the next
+statmenet is not inside a macro expansion.
+"""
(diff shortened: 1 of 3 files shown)


src/commands.jl
```diff
@@ -213,6 +213,35 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
...
+    pc = maybe_step_through_macro_expansion!(frame)
+
+Return the current program counter of `frame` if it is not inside a macro expresion
+Otherwise, step through the statements of `frame` until the next statement is not inside
+a macro expansion.
+"""
+function maybe_step_through_macro_expansion!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false)
+    pc = frame.pc
+    is_macro_expansion(frame) || return pc
+    while pc !== nothing && !isa(pc, BreakpointRef)
```
KristofferC: This is kinda similar to `next_until!` but the predicate there takes an expression, but here we need more than the expression.

KristofferC:
Does anyone know exactly when a 

For example:

```

julia> function f(x)
           x = sin(2)
           @test 1 == 1
           x = sin(2)
       end

julia> (@code_lowered f(2)).linetable
3-element Array{Any,1}:
 Core.LineInfoNode(Main, :f, Symbol("REPL[16]"), 2, 0)
 Core.LineInfoNode(Main, :f, Symbol("REPL[16]"), 3, 0)
 Core.LineInfoNode(Main, :f, Symbol("REPL[16]"), 4, 0)
```

has no macro expansions line info.

KristofferC:
It would be good to have a discussion of what should be the default here etc.

src/commands.jl
```diff
@@ -213,6 +213,49 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
 end
 next_line!(frame::Frame, istoplevel::Bool=false) = next_line!(finish_and_return!, frame, istoplevel)
 
+is_macro_expansion(lin::LineInfoNode) = lin.method == Symbol("macro expansion")
```
timholy: This function should be moved to the "predicates" section of `utils.jl`.

src/commands.jl
```diff
@@ -213,6 +213,49 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
...
+is_macro_expansion(::LineNumberNode) = false
```
timholy: Hmm, I might prefer this to be handled by `return isa(linfo, LineInfoNode) ? is_macro_expansion(linfo) : false` check in the method for `::Frame`. Wouldn't want this used on an expression that someone had manually `macroexpand`ed.

src/commands.jl
```diff
@@ -213,6 +213,49 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
...
+function is_macro_expansion(frame::Frame)
```
timholy: Possibly consider naming this `is_in_macro_expansion`?

src/commands.jl
```diff
@@ -213,6 +213,49 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
...
+    src = frame.framecode.src
+    codeloc = src.codelocs[frame.pc]
+    codeloc == 0 && return false
+    linfo = src.linetable[codeloc]
+    return is_macro_expansion(linfo)
+end
+
+"""
+    pc = maybe_step_through_macro_expansion!(recurse, frame)
```
```suggestion
    pc = next_non_macro_expansion!(recurse, frame)
```

src/commands.jl
```diff
@@ -213,6 +213,49 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
...
+    pc = maybe_step_through_macro_expansion!(frame)
```
```suggestion
    pc = next_non_macro_expansion!(frame)
```

KristofferC:
Im not sure there is much worth salvaging here.

PR #156, opened 2019-03-16
ensure we leave wrapper if we step to the end of it

### LINKED ISSUE #155: maybe_step_through_wrapper! can enter an infinite loop
Consider stepping through a method using the `"s"` command:

```jl
function step_through(f, args...)
    frame = JuliaInterpreter.enter_call(f, args...)
    while true
        ret = JuliaInterpreter.debug_command(JuliaInterpreter.finish_and_return!, frame, "s")
        ret == nothing && return JuliaInterpreter.get_return(frame)
        frame, pc = ret
    end
end
```

Calling this on

```jl
f22() = string(:(a + b))
step_through(f22) == "a + b"
```

never finishes executing. This is causing the Debugger.jl tests to hang. Putting in some debug output at:

https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/c41e768454cd71b3fcc028eb772dd0c16e292609/src/commands.jl#L234-L236

and it prints:

```jl
frame.pc = 3
length(stmts) - 1 = 2
frame.pc = 3
length(stmts) - 1 = 2
frame.pc = 3
length(stmts) - 1 = 2
frame.pc = 3
length(stmts) - 1 = 2
```

which means `frame.pc != length(stmts)-1` is always true

Maybe there should be some:

```
frame.pc == length(stmts) && return frame
```

or

```
pc === nothing && return frame
```

exit condition as well? That makes the debugger tests pass again.

The frame this happens on is:

```jl
frame = Frame for kwfunc(f) in Core at boot.jl:330
  1* 330  1 ─      nothing
  2  330  │   %2 = $(Expr(:foreigncall, :(:jl_get_keyword_sorter), Any, svec(Any), :(:ccall), 1, :(f)))
  3  330  └──      return %2
f = GenericIOBuffer{Array{UInt8,1}}
```

and `next_call!` makes us go to `pc == 3`.

KristofferC:
Fixes #155

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -233,6 +233,7 @@ function maybe_step_through_wrapper!(@nospecialize(recurse), frame::Frame)
         # this is a wrapper function that we might want to step through
         while frame.pc != length(stmts)-1
             pc = next_call!(recurse, frame, false)  # since we're in a Method we're not at toplevel
+            pc === nothing && return frame
         end
         ret = evaluate_call!(dummy_breakpoint, frame, last)
         @assert isa(ret, BreakpointRef)
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -231,7 +231,7 @@ function maybe_step_through_wrapper!(@nospecialize(recurse), frame::Frame)
     if is_kw || isexpr(last, :call) && any(x->x==Core.SlotNumber(1), last.args)
         # If the last expr calls #self# or passes it to an implementation method,
         # this is a wrapper function that we might want to step through
-        while frame.pc < length(stmts)-1
+        while frame.pc != length(stmts)-1
             pc = next_call!(recurse, frame, false)  # since we're in a Method we're not at toplevel
             pc === nothing && return frame
         end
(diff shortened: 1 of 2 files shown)


test/debug.jl
```diff
@@ -9,6 +9,18 @@ function step_through(frame)
     return get_return(r)
 end
```
KristofferC: I wonder if we should have a function that tries to step through a function with all different commands we have. And one that just does a random command each time :D

src/commands.jl
```diff
@@ -233,6 +233,7 @@ function maybe_step_through_wrapper!(@nospecialize(recurse), frame::Frame)
         # this is a wrapper function that we might want to step through
         while frame.pc != length(stmts)-1
```
```suggestion
        while frame.pc < length(stmts)-1
```

KristofferC: Hm, seems like this broke the tests. Thoughts?

test/debug.jl
KristofferC: #157

src/commands.jl
```diff
@@ -231,8 +231,9 @@ function maybe_step_through_wrapper!(@nospecialize(recurse), frame::Frame)
     if is_kw || isexpr(last, :call) && any(x->x==Core.SlotNumber(1), last.args)
         # If the last expr calls #self# or passes it to an implementation method,
         # this is a wrapper function that we might want to step through
-        while frame.pc != length(stmts)-1
+        while frame.pc < length(stmts)-1
```
KristofferC:
```suggestion
        while frame.pc != length(stmts)-1
```

src/commands.jl
KristofferC: I'll just keep the current for now to get this in and then we can think about the `<` :P

PR #157, opened 2019-03-16
more thorough testing of commands

KristofferC:
Stepping with "sg" currently errors with

```
ERROR: MethodError: no method matching #s5#3(::Symbol, ::Symbol)
Closest candidates are:
  #s5#3(::Any, ::Any, ::Any, ::Any) at boot.jl:557
Stacktrace:
 [1] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any,N} where N) at .\boot.jl:522
 [2] get_source at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:51 [inlined]
 [3] #prepare_framecode#11(::Bool, ::Function, ::Method, ::Any) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:141
 [4] (::getfield(JuliaInterpreter, Symbol("#kw##prepare_framecode")))(::NamedTuple{(:enter_generated,),Tuple{Bool}}, ::typeof(JuliaInterpreter.prepare_framecode), ::Method, ::Type) at .\none:0
 [5] #prepare_call#12(::Bool, ::Function, ::Any, ::Array{Any,1}) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:219
 [6] #prepare_call at .\none:0 [inlined]
 [7] #get_call_framecode#30(::Bool, ::Function, ::Array{Any,1}, ::JuliaInterpreter.FrameCode, ::Int64) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\localmethtable.jl:53
 [8] #get_call_framecode at .\none:0 [inlined]
 [9] #evaluate_call_recurse!#35(::Bool, ::Function, ::Any, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:203
 [10] #evaluate_call_recurse! at .\none:0 [inlined]
 [11] #evaluate_call!#37 at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:237 [inlined]
 [12] #evaluate_call! at .\none:0 [inlined]
 [13] debug_command(::Any, ::Frame, ::String, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:341
 [14] debug_command at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:315 [inlined]
 [15] step_through_command(::Frame, ::String) at .\REPL[15]:3
 [16] #step_through#11(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Function) at .\REPL[19]:6
 [17] step_through(::Function) at .\REPL[19]:2
 [18] top-level scope at none:0
```

### DIFF
--- a/test/debug.jl
+++ b/test/debug.jl
@@ -2,24 +2,31 @@ using JuliaInterpreter, Test
 using JuliaInterpreter: enter_call, enter_call_expr, get_return, @lookup
 using Base.Meta: isexpr
 
-function step_through(frame)
-    r = root(frame)
-    @test debug_command(JuliaInterpreter.finish_and_return!, frame, "c") === nothing
-    @test r.callee === nothing
-    return get_return(r)
+const ALL_COMMANDS = ("n", "s", "c", "finish", "nc", "se", "si") # ,"sg") sg current broken
+
+function step_through_command(fr::Frame, cmd::String)
+    while true
+        ret = JuliaInterpreter.debug_command(JuliaInterpreter.finish_and_return!, fr, cmd)
+        ret == nothing && break
+        fr, pc = ret
+    end
+    @test fr.callee === nothing
+    @test fr.caller === nothing
+    return get_return(fr)
 end
 
-function step_through_function(f, args...)
-    frame = JuliaInterpreter.enter_call(f, args...)
-    while true
-        ret = JuliaInterpreter.debug_command(JuliaInterpreter.finish_and_return!, frame, "s")
-        ret == nothing && break
-        frame, pc = ret
+function step_through_frame(frame_creator)
+    rets = []
+    for cmd in ALL_COMMANDS
+        frame = frame_creator() 
+        ret = step_through_command(frame, cmd)
+        push!(rets, ret)
     end
-    @test frame.callee === nothing
-    @test frame.caller === nothing
-    return JuliaInterpreter.get_return(frame)
+    @test all(ret -> ret == rets[1], rets)
+    return rets[1]
 end
+step_through(f, args...; kwargs...) = step_through_frame(() -> enter_call(f, args...; kwargs...))
+step_through(expr::Expr) = step_through_frame(() -> enter_call_expr(expr))
 
 @generated function generatedfoo(T)
     :(return $T)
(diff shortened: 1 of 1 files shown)


timholy:
Once this merges, OK to tag a new version? What should we call it? v0.2.0? There are definitely some breaking behavior changes (e.g., stepping through wrappers) as well as the lovely new `@bp` support.

LoweredCodeUtils was registered this morning. Because of that, theoretically the Revise/Rebugger stack could release at any time. Obviously we are still fixing up the rules for debug-stepping, so I'm in no hurry. You should make the call regarding when to tag Debugger.jl. But if we tag this, it makes it easier to travis-test the new branches of Revise & Rebugger.

KristofferC:
Was going to suggest the same thing. I think a tag is very much in order. `v0.2.0` sounds good to me!

KristofferC:
Seems we are not out of the woods yet with `"sg"` (I presume that is what is failing). Should we comment it out again and release or do we want to look more into it before releae.

timholy:
Hmm, I think testing "sg" this way doesn't really make sense. If I understand @Keno's original intention wrt sg, this lets you debug the generator, meaning the method that creates the expressions returned by the `@generated` function. In general that is not expected to give the same answer as running the generated method. (I.e., `x+1` and `:(x + 1)` are not the same.)

But the following also seems wrong to me:
```julia
julia> @generated function f(x)
           :(return x)
       end

julia> callf() = f("Hello")
callf (generic function with 1 method)

julia> using JuliaInterpreter

julia> fr = JuliaInterpreter.enter_call(callf)
Frame for callf() in Main at REPL[2]:1
  1 1  1 ─ %1 = (f)("Hello")
  2 1  └──      return %1

julia> fr, pc = debug_command(fr, "s")
(Frame for f(x) in Main at REPL[1]:2
  1 2  1 ─     return x
x = "Hello"
caller: callf() in Main at REPL[2]:1, breakpoint(f(x) in Main at REPL[1]:2, %0))

julia> JuliaInterpreter.locals(fr)
2-element Array{JuliaInterpreter.Variable,1}:
 #self# = f 
 x = "Hello"

julia> fr = JuliaInterpreter.enter_call(callf)
Frame for callf() in Main at REPL[2]:1
  1 1  1 ─ %1 = (f)("Hello")
  2 1  └──      return %1

julia> fr, pc = debug_command(fr, "sg")
(Frame for f(x) in Main at REPL[1]:2
  1 2  CodeInfo(quote
x = "Hello"
caller: callf() in Main at REPL[2]:1, breakpoint(f(x) in Main at REPL[1]:2, %0))

julia> JuliaInterpreter.locals(fr)
2-element Array{JuliaInterpreter.Variable,1}:
 #self# = f 
 x = "Hello"
```
Shouldn't `x = String`? Note the `CodeInfo(quote` in the method body; the full expansion is
```julia
julia> fr.framecode.src
CodeInfo(quote
    return x
end)
```

timholy:
I'd say drop `sg` here and then go ahead and tag v0.2.0, can deal with "sg" later.

KristofferC:
Yeah, I also had some hunch that `sg` was perhaps not meant for running through the function. Found a bug with it at least :)

PR #159, opened 2019-03-16
 Fix "sg" for parametric generated functions 

timholy:
Ref #157


KristofferC:
Seems travis didn't catch this PR... Trying reopening.

PR #162, opened 2019-03-16
use symbol instead of strings in debug command

KristofferC:


KristofferC:
If we want this I suggest tagging after this. If not, I suggest closing this and then tagging:D

timholy:
Will wait for #140 and then tag.

PR #166, opened 2019-03-16
fix accessing undef sparam

### LINKED ISSUE #165: Access to undefined reference with undefined static parameter
```
julia> function foo(x::T) where {T <: AbstractString, S <: AbstractString}
           return S
       end

julia> foo("fdjksl")
ERROR: UndefVarError: S not defined
Stacktrace:
 [1] foo(::String) at ./REPL[2]:2
 [2] top-level scope at none:0

julia> @interpret foo("fdjksl")
ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getindex at .\array.jl:729 [inlined]
 [2] lookup_expr at C:\Users\Kristoffer\.julia\packages\JuliaInterpreter\3VHjK\src\interpret.jl:19 [inlined]
 [3] get_return(::Frame) at C:\Users\Kristoffer\.julia\packages\JuliaInterpreter\3VHjK\src\interpret.jl:56
 [4] finish_and_return! at C:\Users\Kristoffer\.julia\packages\JuliaInterpreter\3VHjK\src\commands.jl:31 [inlined]
 [5] finish_and_return! at C:\Users\Kristoffer\.julia\packages\JuliaInterpreter\3VHjK\src\commands.jl:33 [inlined] (repeats 2 times) [6] top-level scope at C:\Users\Kristoffer\.julia\packages\JuliaInterpreter\3VHjK\src\construct.jl:641
```

KristofferC:
Fixes #165

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -16,7 +16,15 @@ end
 function lookup_expr(frame, e::Expr)
     head = e.head
     head == :the_exception && return frame.framedata.last_exception[]
-    head == :static_parameter && return frame.framedata.sparams[e.args[1]::Int]
+    if head == :static_parameter
+        arg = e.args[1]::Int
+        if isassigned(frame.framedata.sparams, arg)
+            return frame.framedata.sparams[arg]
+        else
+            syms = sparam_syms(frame.framecode.scope)
+            throw(UndefVarError(syms[arg]))
+        end
+    end
     head == :boundscheck && length(e.args) == 0 && return true
     error("invalid lookup expr ", e)
 end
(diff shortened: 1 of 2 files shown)


src/interpret.jl
```diff
@@ -16,7 +16,15 @@ end
 function lookup_expr(frame, e::Expr)
     head = e.head
     head == :the_exception && return frame.framedata.last_exception[]
-    head == :static_parameter && return frame.framedata.sparams[e.args[1]::Int]
+    if head == :static_parameter
+        arg = e.args[1]::Int
+        if isassigned(frame.framedata.sparams, arg)
+            return frame.framedata.sparams[arg]
+        else
+            syms = sparam_syms(frame.framecode.scope)
```
KristofferC: This is arguably a bit slower than needed but I really don't think this is a case to worry too much about.

PR #167, opened 2019-03-17
Update README

davidanthoff:


timholy:
This seems like a good default. Any reason not to have both?

KristofferC:
You can always click the "dev" in the drop down menu in the docs. 99% of the people clicking this link should probably go to the latest release version, so it seems a bit unnecessary to list both here.

fredrikekre:
L15 and L16 can be removed too.

Issue #170, opened 2019-03-17
New "build system" has a possible failure mode


KristofferC:
If JuliaInterpreter ends up getting precompiled during testing of a package that does not have InteractiveUtils in its test- or normal dependencies, the process that runs the generate builtins file will fail because it cant find the InteractiveUtils package.

The solution would be to have the spawned process run with the project file here as the active project.

PR #174, opened 2019-03-18
fix what project is used when building builtins

KristofferC:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/170

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -25,7 +25,7 @@ const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(I
 @info "Generating builtins for this julia version..."
 gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
 # Run as separate command to prevent including the generate_builtins into the precompile cache
-withenv("JULIA_LOAD_PATH"=joinpath(@__DIR__, "..")) do
+withenv("JULIA_LOAD_PATH" => joinpath(@__DIR__, "..")) do
     run(`$(Base.julia_cmd()) --startup-file=no $gen_builtins_file`)
 end
 include_dependency(gen_builtins_file)
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -25,7 +25,8 @@ const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(I
 @info "Generating builtins for this julia version..."
 gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
 # Run as separate command to prevent including the generate_builtins into the precompile cache
-run(`$(Base.julia_cmd()) --startup-file=no $gen_builtins_file`)
+const thisdir = @__DIR__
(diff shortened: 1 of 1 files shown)


src/JuliaInterpreter.jl
```diff
@@ -25,7 +25,7 @@ const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(I
 @info "Generating builtins for this julia version..."
 gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
 # Run as separate command to prevent including the generate_builtins into the precompile cache
-run(`$(Base.julia_cmd()) --startup-file=no $gen_builtins_file`)
+run(`$(Base.julia_cmd()) --project=@. --startup-file=no $gen_builtins_file`)
```
fredrikekre:
```suggestion
run(`$(Base.julia_cmd()) --project=$(joinpath(@__DIR__, "..") --startup-file=no $gen_builtins_file`)
```

KristofferC: Thanks. I wonder if all this is worth it just to save caching the builtin generator... Oh well.

fredrikekre:
You need to step up one level from the `src` directory.

KristofferC:
Don't think so, `.@` will look upwards, no?

fredrikekre:
Actually, I ran into this issue because I had a custom `LOAD_PATH` where `@` was not included, so the fix should be 
```diff
diff --git a/src/JuliaInterpreter.jl b/src/JuliaInterpreter.jl
index b2c6d43..2625fa1 100644
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -25,7 +25,9 @@ const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(I
 @info "Generating builtins for this julia version..."
 gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
 # Run as separate command to prevent including the generate_builtins into the precompile cache
-run(`$(Base.julia_cmd()) --startup-file=no $gen_builtins_file`)
+withenv("JULIA_LOAD_PATH"=joinpath(@__DIR__, "..")) do
+    run(`$(Base.julia_cmd()) --startup-file=no $gen_builtins_file`)
+end
 include_dependency(gen_builtins_file)
 
 include("types.jl")
```

src/JuliaInterpreter.jl
```diff
@@ -25,7 +25,9 @@ const BUILTIN_FILE = joinpath(@__DIR__, "builtins-julia$(Int(VERSION.major)).$(I
 @info "Generating builtins for this julia version..."
 gen_builtins_file = joinpath(@__DIR__, "generate_builtins.jl")
 # Run as separate command to prevent including the generate_builtins into the precompile cache
-run(`$(Base.julia_cmd()) --startup-file=no $gen_builtins_file`)
+withenv("JULIA_LOAD_PATH"=joinpath(@__DIR__, "..")) do
```
fredrikekre:
```suggestion
withenv("JULIA_LOAD_PATH" => joinpath(@__DIR__, "..")) do
```

timholy:
Looks fine to me, but y'all know more about this part than I do.

PR #179, opened 2019-03-18
Overload show(io, mime, x) instead of show(io, x) for pretty-printing.

fredrikekre:
I think the three-arg method is recommended for pretty-printing. Otherwise the pretty-printing leaks out in e.g. error messages.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -85,7 +85,7 @@ end
 
 nstatements(framecode::FrameCode) = length(framecode.src.code)
 
-Base.show(io::IO, framecode::FrameCode) = print_framecode(io, framecode)
+Base.show(io::IO, ::MIME"text/plain", framecode::FrameCode) = print_framecode(io, framecode)
 
 """
 `FrameInstance` represents a method specialized for particular argument types.
@@ -175,7 +175,7 @@ Return the deepest callee in the call stack.
 """
 leaf(frame) = traverse(callee, frame)
 
-function Base.show(io::IO, frame::Frame)
+function Base.show(io::IO, ::MIME"text/plain", frame::Frame)
     frame_loc = CodeTracking.replace_buildbot_stdlibpath(repr(scopeof(frame)))
     println(io, "Frame for ", frame_loc)
     pc = frame.pc
@@ -208,7 +208,10 @@ struct Variable
     name::Symbol
     isparam::Bool
 end
-Base.show(io::IO, var::Variable) = (print(io, var.name, " = "); show(io,var.value))
+function Base.show(io::IO, m::MIME"text/plain", var::Variable)
+    print(io, var.name, " = ")
+    show(io, m, var.value)
+end
 Base.isequal(var1::Variable, var2::Variable) =
     var1.value == var2.value && var1.name == var2.name && var1.isparam == var2.isparam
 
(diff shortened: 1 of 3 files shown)


timholy:
If you can fix CI, I approve. Thanks!

fredrikekre:
The failures are doctest failures, and those make me think this is not a good change, since two-arg show is used when printing containers. For example the breakpoint example will turn from
```
julia> frame, bp = @interpret sum([1,2,5])
(Frame for sum(a::AbstractArray) in Base at reducedim.jl:645
c 1 645      1 ─      nothing
  2 645  645 │   %2 = (Base.#sum#538)(Colon(), #self#, a)
  3 645      └──      return %2
a = [1, 2, 5], breakpoint(sum(a::AbstractArray) in Base at reducedim.jl:645, line 645))
```
to
```
julia> frame, bp = @interpret sum([1,2,5])
(Frame(JuliaInterpreter.FrameCode(sum(a::AbstractArray) in Base at reducedim.jl:648, CodeInfo(
1 ─      nothing
│   @ reducedim.jl:648 within `sum'
│   %2 = ($(QuoteNode(Base.#sum#558)))($(QuoteNode(Colon())), #self#, a)
└──      return %2
), Union{Compiled, Core.TypeMapEntry}[#undef, #undef, #undef], JuliaInterpreter.BreakpointState[JuliaInterpreter.BreakpointState(true, ##slotfunction#373), #undef, #undef], BitSet([2]), false), JuliaInterpreter.FrameData(Union{Nothing, Some{Any}}[Some(sum), Some([1, 2, 5])], Any[#undef, #undef, #undef], Any[], Int64[], Base.RefValue{Any}(nothing), Dict(:a => 2,Symbol("#self#") => 1), Any[]), 1, nothing, nothing), BreakpointRef(JuliaInterpreter.FrameCode(sum(a::AbstractArray) in Base at reducedim.jl:648, CodeInfo(
1 ─      nothing
│   @ reducedim.jl:648 within `sum'
│   %2 = ($(QuoteNode(Base.#sum#558)))($(QuoteNode(Colon())), #self#, a)
└──      return %2
), Union{Compiled, Core.TypeMapEntry}[#undef, #undef, #undef], JuliaInterpreter.BreakpointState[JuliaInterpreter.BreakpointState(true, ##slotfunction#373), #undef, #undef], BitSet([2]), false), 1, nothing))
```
because there is a tuple returned. The examples can of course be updated to display them separately like so:
```
julia> frame, bp = @interpret sum([1,2,5]);

julia> frame
Frame for sum(a::AbstractArray) in Base at reducedim.jl:648
c 1 648  1 ─      nothing
  2 648  │   @ reducedim.jl:648 within `sum'
  3 648  │   %2 = (Base.#sum#558)(Colon(), #self#, a)
Variable([1, 2, 5], :a, false)

julia> bp
breakpoint(sum(a::AbstractArray) in Base at reducedim.jl:648, line 648)
```
I don't really know what is "correct", but from https://github.com/JuliaLang/julia/pull/30757#issuecomment-456925881 it sounds like three-arg should be used for pretty-printing. Also, with the PR as is, the fact that both [`Method`](https://github.com/JuliaLang/julia/blob/b93fd2350e64d626e35cd57759271e8d5045dbfa/base/methodshow.jl#L115) and [`CodeInfo`](https://github.com/JuliaLang/julia/blob/b93fd2350e64d626e35cd57759271e8d5045dbfa/base/show.jl#L1597) pretty-prints with two-arg makes things look weird anyway.

fredrikekre:
Actually, the [docs for `show`](https://docs.julialang.org/en/v1/base/io-network/#Base.show-Tuple{Any,Any,Any}) say that
> The default MIME type is MIME"text/plain". There is a fallback definition for text/plain output that calls show with 2 arguments. Therefore, this case should be handled by defining a 2-argument show(io::IO, x::MyType) method.

which contradicts the link from the previous post.

KristofferC:
Think we just ignore this for now.

PR #180, opened 2019-03-18
use compiled mode when there is no chance we will encounter a breakpoint

KristofferC:
This should probably be documented?

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -292,6 +292,19 @@ function unwind_exception(frame::Frame, exc)
     rethrow(exc)
 end
 
+const always_run_recursive_interpret = Ref(false)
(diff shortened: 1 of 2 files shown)


timholy:
Oh, arghh, I just thought of something: `break_on(:error)`?

That's a real bummer because otherwise this is a great workaround.

KristofferC:
But I check for that?

https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/180/files#diff-ae619a7f5a7be402945f88bcc3c5e049R302

timholy:
Hmm, I should have looked a little more carefully, this was broken in two respects. I wonder why the tests didn't catch it? #184

PR #182, opened 2019-03-18
step through invokes

KristofferC:
With this:

```jl
julia> @enter Dense(10,10)(randn(10,10))
In Dense(x) at C:\Users\Kristoffer\.julia\packages\Flux\WSB7k\src\layers\basic.jl
>81    W, b, σ = a.W, a.b, a.σ
 82    σ.(W*x .+ b)
 83  end
 84  
 85  function Base.show(io::IO, l::Dense)

About to run: (getproperty)(<suppressed 109 bytes of output>, :W)
```

TODO:

- [x] Tests
- [x] Figure out what to do when `prepare_framecode` returns `nothing`.

Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/177

### DIFF
--- a/src/generate_builtins.jl
+++ b/src/generate_builtins.jl
@@ -153,9 +153,7 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
             if !expand
                 return Some{Any}(invoke(argswrapped...))
             end
-            new_f = which(argswrapped[1], argswrapped[2])
-            new_expr = Expr(:call, new_f, argswrapped[3:end]...)
-            return new_expr
+            return Expr(:call, invoke, argswrapped...)
 """)
             continue
         end
--- a/src/generate_builtins.jl
+++ b/src/generate_builtins.jl
@@ -138,11 +138,24 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
     $head f === Core._apply
         argswrapped = getargs(args, frame)
         if !expand
-            return Some{Any}(Core._apply(getargs(args, frame)...))
+            return Some{Any}(Core._apply(argswrapped...))
         end
         argsflat = Base.append_any((argswrapped[1],), argswrapped[2:end]...)
         new_expr = Expr(:call, map(x->isa(x, Symbol) || isa(x, Expr) || isa(x, QuoteNode) ? QuoteNode(x) : x, argsflat)...)
         return new_expr
+""")
+            continue
+        elseif f === Core.invoke
+            print(io, 
+"""
+    $head f === invoke
+            argswrapped = getargs(args, frame)
+            if !expand
+                return Some{Any}(invoke(argswrapped...))
+            end
+            new_f = which(argswrapped[1], argswrapped[2])
+            new_expr = Expr(:call, new_f, argswrapped[3:end]...)
+            return new_expr
 """)
             continue
         end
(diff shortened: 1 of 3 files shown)


src/generate_builtins.jl
```diff
@@ -138,11 +138,24 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
...
+            continue
+        elseif f === Core.invoke
+            print(io, 
+"""
+    $head f === invoke
+            argswrapped = getargs(args, frame)
+            if !expand
+                return Some{Any}(invoke(argswrapped...))
+            end
+            new_f = which(argswrapped[1], argswrapped[2])
```
timholy: Since lowered AST shouldn't return methods as `Expr` args, I think this is probably safe. I worry a bit, though, about what might happen if we want to add more aggressive optimization passes; if we suddenly get to a point where we are stashing more objects in Exprs, I think we'll run into trouble.

Maybe best to wrap it in some novel type? Or could we make it a "typical :call expr" like above for `Core._apply`?

src/interpret.jl
```diff
@@ -208,7 +208,16 @@ function evaluate_call_recurse!(@nospecialize(recurse), frame::Frame, call_expr:
         err = length(fargs) > 1 ? fargs[2] : frame.framedata.last_exception[]
         throw(err)
     end
-    framecode, lenv = get_call_framecode(fargs, frame.framecode, frame.pc; enter_generated=enter_generated)
+    if fargs[1] isa Method # happens on invoke, fargs[1] is the method to call
+        f, sig = fargs[1], Tuple{typeof.(fargs[2:end])...}
```
timholy: Better to use `_Typeof` since it handles types better.

src/interpret.jl
```diff
@@ -208,7 +208,16 @@ function evaluate_call_recurse!(@nospecialize(recurse), frame::Frame, call_expr:
         err = length(fargs) > 1 ? fargs[2] : frame.framedata.last_exception[]
         throw(err)
     end
-    framecode, lenv = get_call_framecode(fargs, frame.framecode, frame.pc; enter_generated=enter_generated)
+    if fargs[1] isa Method # happens on invoke, fargs[1] is the method to call
```
timholy: Comments below are predicated on the possibility that you *won't* reconstruct an ordinary call. (But you don't need to worry about any of this if that's what you decide to do.)

src/interpret.jl
```diff
@@ -208,7 +208,16 @@ function evaluate_call_recurse!(@nospecialize(recurse), frame::Frame, call_expr:
...
+        f, sig = fargs[1], Tuple{typeof.(fargs[2:end])...}
+        ret = prepare_framecode(f, sig; enter_generated=enter_generated)
+        if ret === nothing
+            error("help...")
```
timholy: Just call `f` on the args directly. Same for `isa(ret, Compiled)`.

KristofferC: But `f` is a `Method` and you can't call methods. I guess I could just `invoke` it?

timholy:
I haven't thought about this at all, so take it for what little it's worth: any work that can be done in `optimize!` is work saved later. For `Core._apply`, it seemed that wasn't going to be very easy (at least in the circumstances I checked) so I didn't bother at the time (but it's something I want to revisit eventually). If the method is always "hard coded" in an invoke, then perhaps we can save some runtime? Anyway, something to keep a lookout for.

KristofferC:
Updated.

If the method is a literal then we can do the lookup at `optimize!` time but the amount `invoke` is used in Julia is quite (very) small so perhaps, for now, it isn't a big deal.

KristofferC:
Easier review: https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/182/files?w=1

timholy:
Awesome work, thanks so much!

PR #184, opened 2019-03-19
Step through invokelatest

timholy:
Also fixes bugs in #180

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -300,8 +300,9 @@ end
 
 const always_run_recursive_interpret = Ref(false)
 function any_active_breakpoint()
-    for bp in _breakpoints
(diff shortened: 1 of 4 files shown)


src/commands.jl
```diff
@@ -300,8 +300,9 @@ end
 
 const always_run_recursive_interpret = Ref(false)
 function any_active_breakpoint()
-    for bp in _breakpoints
```
KristofferC: Hmm, how did this just not error...

fredrikekre: It errors while running Debuggers test suite, so maybe its just not tested in this repo.

KristofferC:
Rebased on top of #185

PR #185, opened 2019-03-19
Revert "use compiled mode when there is no chance we will encounter a breakpoint"

KristofferC:
Reverts JuliaDebug/JuliaInterpreter.jl#180

We forgot `@bp`...


timholy:
Right... dang.

I guess FrameCompiler.jl will be high priority soon! (Or just wire it into this package, not sure of the best organization.)

KristofferC:
I think just putting such functionality here would be most convenient.

PR #189, opened 2019-03-19
add docs for break_on

KristofferC:


docs/src/index.md
```diff
@@ -94,6 +94,52 @@ combinations or that you want to break on entry to *any* method of a particular
...
+```jldoctest demo1
+julia> function f_outer()
+           println("before error")
+           f_inner()
+           println("after error")
+       end;
+
+julia> f_inner() = error("inner error");
+
+julia> JuliaInterpreter.break_on(:error)
```
timholy: `break_on` and `break_off` are exported. Since we don't qualify `@interpret` etc., let's do the same here.

KristofferC: Nice catch, fixed.

PR #190, opened 2019-03-19
Step through preparation for kwarg calls

timholy:
I am tired of stepping through `NamedTuple`, `pairs`, `merge`, and `Core.kwfunc`. This feels better.

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -250,6 +250,72 @@ function maybe_step_through_wrapper!(@nospecialize(recurse), frame::Frame)
 end
 maybe_step_through_wrapper!(frame::Frame) = maybe_step_through_wrapper!(finish_and_return!, frame)
 
+
+"""
+    frame = maybe_step_through_kwprep!(recurse, frame)
+    frame = maybe_step_through_kwprep!(frame)
+
+If `frame.pc` points to the beginning of preparatory work for calling a keyword-argument
+function, advance forward until the actual call.
+"""
+function maybe_step_through_kwprep!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false)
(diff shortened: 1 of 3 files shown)


KristofferC:
Great! A related question, should we also step "out" of wrappers, e.g.

```

**Comment by pfitzseb** [comment]
Yes please :) Although there probably should be some way to inspect all wrappers etc.

**Comment by KristofferC** [comment]
I guess it can also be up to the frontend to call `maybe_step_through_wrapper!` and remember if it was a wrapper and in that case `finish!` the frame when it becomes active again.

**Comment by timholy** [comment]
You can definitely do fine-grained stuff with `se` and `si`. Since this doesn't change frames at all, stepping out of wrappers is a bit orthogonal to this particular PR.

But I'd be fine with moving the magic out of `debug_command`. Is that the consensus? Vote thumbs-up or thumbs-down.

**Comment by KristofferC** [comment]
Personally, I don't mind since I would likely be using this myself. It is just in the situation where one does not want the automatic wrapper stepping, one would have to reconstruct `s` and co based on `si` then?

**Comment by timholy** [comment]
Yes, `si` should avoid the wrapper-stepping. To me that seems like a reasonable solution.

**Comment by timholy** [comment]
Test failures seem spurious, so if we decide we like this we could merge. Want to make the call, @KristofferC?

**Comment by KristofferC** [comment]
> Yes, si should avoid the wrapper-stepping.

Yes, but it also doesn't step to the next caller. But sure, someone can do

```
function s_without_step_through(frame)
    while true
        ret = si(frame)
        ret = nothing && ...
        new_frame, pc = ret
        frame !== new_frame && return new_frame
    end
end
```

PR #191, opened 2019-03-19
allow 0.4 semver for CodeTracking

KristofferC:


KristofferC:
We should be compatible here, right?

KristofferC:
Docs are broken, fixed in #192

PR #197, opened 2019-03-20
step out of wrapper calls

KristofferC:


src/commands.jl
```diff
@@ -339,8 +343,14 @@ function maybe_reset_frame!(@nospecialize(recurse), frame::Frame, @nospecialize(
         frame = caller(frame)
         frame === nothing && return nothing
         frame.callee = nothing
+        ssavals = frame.framedata.ssavalues
+        is_wrapper = isassigned(ssavals, frame.pc) && coalesce(ssavals[frame.pc] == Wrapper(), false)
```
KristofferC: This `coalesce` was fun...

test/debug.jl
```diff
@@ -154,7 +154,7 @@ struct B{T} end
         frame = JuliaInterpreter.enter_call(f, 2; b = 4)
         fr = JuliaInterpreter.maybe_step_through_wrapper!(frame)
         fr, pc = debug_command(fr, :nc)
-        fr, pc = debug_command(fr, :nc)
```
KristofferC: This now actually finishes the frame.

src/commands.jl
KristofferC: Actually, can just use `===`

timholy:
Beautiful!

PR #199, opened 2019-03-20
ensure si doesn't step over anything

### LINKED ISSUE #196: Should :si not step through wrappers?
`si` is a quite "low level" command so perhaps it should not step through wrappers / kwprep when the next statement is a call?

KristofferC:
Fixes #196

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -405,9 +405,11 @@ function debug_command(@nospecialize(recurse), frame::Frame, cmd::Symbol, rootis
 
     istoplevel = rootistoplevel && frame.caller === nothing
     cmd0 = cmd
+    is_si = false
     if cmd == :si
         stmt = pc_expr(frame)
         cmd = is_call(stmt) ? :s : :se
+        is_si = true
     end
     try
         cmd == :nc && return nicereturn!(recurse, frame, next_call!(recurse, frame, istoplevel), rootistoplevel)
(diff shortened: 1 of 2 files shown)


src/commands.jl
```diff
@@ -439,7 +441,8 @@ function debug_command(@nospecialize(recurse), frame::Frame, cmd::Symbol, rootis
             if isa(ret, BreakpointRef)
                 newframe = leaf(frame)
                 cmd0 == :si && return newframe, ret
-                newframe = maybe_step_through_wrapper!(recurse, newframe)
+                is_si || (newframe = maybe_step_through_wrapper!(recurse, newframe))
+                is_si || maybe_step_through_kwprep!(recurse, newframe, istoplevel)
```
KristofferC: The reason for this addition is the following case:

```jl
julia> f(x, y) = g(x; y=y);

julia> g(x; y=2) = x + y;

julia> h(x, y) = f(x, y);

julia> frame = JuliaInterpreter.enter_call(h, 2, 1);

julia> frame, pc =  debug_command(frame, :s);

julia> frame
Frame for f(x, y) in Main at REPL[3]:1
  1* 1  1 ─ %1 = (:y,)
  2  1  │   %2 = (Core.apply_type)(NamedTuple, %1)
  3  1  │   %3 = (tuple)(y)
⋮
x = 2
y = 1
```

We want the `:s` to have stepped over the kwprep even after stepping over the wrapper.

KristofferC:
Perhaps we can make a new release after this? I want to release Debugger 0.2.0 as well.

KristofferC:
I'll make a new release then?

timholy:
Yes, much needed.

Have you been running the new Revise? Aside from https://github.com/timholy/Revise.jl/pull/243#issuecomment-474251076 any problems with it? I'm a little nervous that all the extra analysis will be rough on slower machines.

KristofferC:
It seems I have been running Revise master, but I'll swap over to `teh/lowered` now and see if I find anything.

> I'm a little nervous that all the extra analysis will be rough on slower machines.

My machine is pretty much a monster so I wouldn't be able to answer that :P

PR #209, opened 2019-03-24
Support CodeTracking 0.5

timholy:


timholy:
Not sure if there's a different strategy we want to use for these compat issues, happy to accept guidance.

timholy:
Hmm, on Travis it's installing CodeTracking 0.3.2. Why?

KristofferC:
Because https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/c3854f57fbedf65b16b788ab12a7f41c9d0f1230/Manifest.toml#L10.

I don't see a better way than this (and I don't think the current way is bad). If a dependency releases a breaking version, we have to test that we are compatible and then add a line saying that we are.

timholy:
Why can't we just delete that Manifest now? For anyone who `add`s this, they won't even reference it.

KristofferC:
We can delete it. It's just up to us to decide if we want to test things on fixed versions or the versions the resolver happens to give us when CI is run.

timholy:
I think it's better to test what the majority of users will test.

Issue #18, opened 2019-02-12
Stacktraces point to the interpreter, not the code


timholy:
Using [this function from the test suite](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/3e5c4ab9bb11392ee49b1bc8c7a126791ad21b57/test/interpret.jl#L41-L48), we get this:
```julia
julia> f_exc_outer1()
ERROR: inner
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] f_exc_inner() at ./REPL[3]:2
 [3] f_exc_outer1() at ./REPL[6]:3
 [4] top-level scope at none:0
```
Unfortunately, when running in the interpreter
```julia
julia> @interpret f_exc_outer1()
ERROR: inner
Stacktrace:
 [1] evaluate_call!(::Array{JuliaStackFrame,1}, ::JuliaStackFrame, ::Expr, ::JuliaInterpreter.JuliaProgramCounter) at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:177
 [2] eval_rhs(::Array{JuliaStackFrame,1}, ::JuliaStackFrame, ::Expr, ::JuliaInterpreter.JuliaProgramCounter) at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:292
 [3] _step_expr!(::Array{JuliaStackFrame,1}, ::JuliaStackFrame, ::Any, ::JuliaInterpreter.JuliaProgramCounter, ::Bool) at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:399
 [4] _step_expr!(::Array{JuliaStackFrame,1}, ::JuliaStackFrame, ::JuliaInterpreter.JuliaProgramCounter, ::Bool) at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:437
 [5] finish!(::Array{JuliaStackFrame,1}, ::JuliaStackFrame, ::JuliaInterpreter.JuliaProgramCounter, ::Bool) at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:463
 [6] finish_and_return!(::Array{JuliaStackFrame,1}, ::JuliaStackFrame, ::JuliaInterpreter.JuliaProgramCounter, ::Bool) at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:482 (repeats 3 times)
[...]

PR #211, opened 2019-03-24
make `show_backtrace` and `display_error` work for Frame

KristofferC:
When an error is encountered in interpreted code, naively printing the stacktrace will just spew out internals of the interpreter which is not so nice to show a user. Fortunately, it is fairly easy to recreate the data structures that Base uses to print a "native" stacktrace from the leaf `Frame`. For example, with this PR:

```jl
julia> f() = "αβ"[2]

julia> break_on(:error)

julia> frame, bp = @interpret f();

julia> Base.display_error(stdout, bp.err, leaf(frame))
ERROR: StringIndexError("αβ", 2)
Stacktrace:
 [1] string_index_err(::String, ::Int64) at strings/string.jl:12
 [2] getindex_continued(::String, ::Int64, ::UInt32) at strings/string.jl:218
 [3] getindex(::String, ::Int64) at strings/string.jl:211
 [4] f() at REPL[122]:1
```

Compare vs the native printing in error:

```jl
julia> f()
ERROR: StringIndexError("αβ", 2)
Stacktrace:
 [1] string_index_err(::String, ::Int64) at ./strings/string.jl:12
 [2] getindex_continued(::String, ::Int64, ::UInt32) at ./strings/string.jl:218
 [3] getindex at ./strings/string.jl:211 [inlined]
 [4] f() at ./REPL[122]:1
 [5] top-level scope at none:0
```

Since we just use the Base functions for the printing, we get the nice colored printing for free.
A frontend for a debugger can then chose to catch the error and display this stacktrace instead.

Ref: https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/18, https://github.com/JuliaDebug/Debugger.jl/issues/145

TODO:
- [x] Docs.
- [x] Tests.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -338,3 +338,49 @@ function show_stackloc(io::IO, frame)
     end
 end
 show_stackloc(frame) = show_stackloc(stdout, frame)
+
+# Pretty printing of stacktraces and errors with Frame
+function Base.StackTraces.StackFrame(frame::Frame)
+    if frame.framecode.scope isa Method
+        method = frame.framecode.scope
+        method_args = something.(frame.framedata.locals[1:frame.framecode.scope.nargs])
+        atypes = Tuple{typeof.(method_args)...}
+        sig = frame.framecode.scope.sig
+        (ti, lenv) = ccall(:jl_type_intersection_with_env, Any, (Any, Any), atypes, sig)
+        mi = Core.Compiler.code_for_method(method, atypes, lenv, typemax(UInt))
+    else
+        mi = frame.framecode.src
+    end
+
+    Base.StackFrame(
+        frame.framecode.scope.name,
+        Symbol(JuliaInterpreter.getfile(frame)),
+        JuliaInterpreter.linenumber(frame),
+        mi,
+        false,
+        false,
+        C_NULL
+    )
+end
+
+function Base.show_backtrace(io::IO, frame::Frame)
+    stackframes = Tuple{Base.StackTraces.StackFrame, Int}[]
+    while frame !== nothing
+        push!(stackframes, (Base.StackTraces.StackFrame(frame), 1))
+        frame = JuliaInterpreter.caller(frame)
+    end
+
+    print(io, "\nStacktrace:")
+    try invokelatest(Base.update_stackframes_callback[], stackframes) catch end
+    frame_counter = 0
+    for (last_frame, n) in stackframes
+        frame_counter += 1
+        Base.show_trace_entry(IOContext(io, :backtrace => true), last_frame, n, prefix = string(" [", frame_counter, "] "))
+    end
+end
+
+function Base.display_error(io::IO, er, frame::Frame)
+    printstyled(io, "ERROR: "; bold=true, color=Base.error_color())
+    showerror(IOContext(io, :limit => true), er, frame)
+    println(io)
+end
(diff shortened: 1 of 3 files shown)


src/utils.jl
```diff
@@ -338,3 +338,49 @@ function show_stackloc(io::IO, frame)
...
 show_stackloc(frame) = show_stackloc(stdout, frame)
+
+# Pretty printing of stacktraces and errors with Frame
+function Base.StackTraces.StackFrame(frame::Frame)
+    if frame.framecode.scope isa Method
+        method = frame.framecode.scope
+        method_args = something.(frame.framedata.locals[1:frame.framecode.scope.nargs])
+        atypes = Tuple{typeof.(method_args)...}
+        sig = frame.framecode.scope.sig
+        (ti, lenv) = ccall(:jl_type_intersection_with_env, Any, (Any, Any), atypes, sig)
```
KristofferC: We already have the `sparams` but I need them as an `SimpleVector`. Is there another way to do this then just call the intersection again?

KristofferC:
Where should this be documented? In the docstring of `Frame` itself?

KristofferC:
One problem is if I want to show a stacktrace even when an error is not caught by `break_on(:error)`. The fact that `handle_err` "destroys" the stack by unlinking the frames means that we cannot get the backtrace after unwinding. Not sure how to solve that.

pfitzseb:
Couldn't we construct the stack trace [here](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/87314a476b89e12e4394d17f83f78c5cb7cfaf90/src/interpret.jl#L575) and save it in `JuliaInterpreter.last_stacktrace[]` or something so that the frontend can take the error message from the exception that's being rethrown and the stacktrace from that global ref?

KristofferC:
Yeah.

We could create a `StackFrame` from the `Frame` here: https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/87314a476b89e12e4394d17f83f78c5cb7cfaf90/src/interpret.jl#L589 and `push!` it to some global variable while unwinding. One question is, when do we `empty!` that stack?

src/utils.jl
```diff
@@ -338,3 +338,49 @@ function show_stackloc(io::IO, frame)
...
+        mi = Core.Compiler.code_for_method(method, atypes, lenv, typemax(UInt))
+    else
+        mi = frame.framecode.src
```
timholy: Is the CodeInfo what normally gets inserted when toplevel throws an error? I don't know the answer myself. If you haven't tested you can use `prepare_thunk` and see.

KristofferC:
> Or is it completely orthogonal? (I'm guessing the latter.)

Is is orthogonal since nothing calls the overloaded methods (since before this PR they didn't exist).

src/utils.jl
KristofferC: Yes, but with `framedata.sparams` instead of `framedata.locals`

src/utils.jl
KristofferC: Looking at the printing, that seems to be the case:

```
julia> foo(x) = x

julia> Base.StackTraces.StackFrame(:foo, Symbol("file.jl"), 5, (@code_lowered foo(2)), false, false, C_NULL)
top-level scope at file.jl:5
```

KristofferC:
Updated from review, added docs added tests.

PR #214, opened 2019-03-25
only unlink frames if exception will not be thrown to top level

KristofferC:
This allows a frontend to inspect the call stack after error by keeping the links between frames in case we would throw to top level and thus do something like:

```jl
julia> f() = "αβ"[2]

julia> frame = JuliaInterpreter.enter_call(f);

julia> err = try debug_command(frame, :c)
       catch err
            err
       end
StringIndexError("αβ", 2)

julia> Base.display_error(stdout, err, leaf(frame))
ERROR: StringIndexError("αβ", 2)
Stacktrace:
 [1] string_index_err(::String, ::Int64) at strings/string.jl:12
 [2] getindex_continued(::String, ::Int64, ::UInt32) at strings/string.jl:218
 [3] getindex(::String, ::Int64) at strings/string.jl:211
 [4] f() at REPL[2]:1
```

This seemed like the easiest implementation and it feels like it should be fine since we are just throwing anyway, so we can kinda do what we want with the frames? There are other possible implementations that would still allow us to do the unlinking but this has the fewest lines of codes changed, so putting this one up first.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -552,16 +552,18 @@ behaviors:
 """
 function handle_err(@nospecialize(recurse), frame, err)
     data = frame.framedata
+    err_will_be_thrown_to_top_level = isempty(data.exception_frames) && !data.caller_will_catch_err
     if break_on_error[]
         # See if the current frame or a frame in the stack will catch this exception,
         # otherwise this exception would have been thrown to the user and we should
         # return a breakpoint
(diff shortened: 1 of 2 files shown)


KristofferC:
>  this is the kind of change that feels like it could conceivably get us in trouble at some later point.

Yeah, but thinking about it, I don't really see how this is "worse" than what we do when breaking on error. So let's try this and see what happens.

Issue #206, opened 2019-03-22
Performance opportunities


timholy:
In my initial performance work, the big 3 improvements were local method tables (`framecode.methodtables`), reusing framedata (the `junk` mechanism), and getting rid of runtime dispatch internally within the interpreter (by manual union-splitting, aka, `if isa(x, ConcreteType)...`).

Some possible future improvements:

- [ ] Recent changes diving into `Core._apply`, `Core._apply_latest`, and `invoke` hurt performance. Perhaps some of that might be clawed back by moving more of this work into `optimize!`?
- [ ] `ccall`s are slow. I just pushed a `teh/compiled_ccall` that contains an old attempt to create one compiled function per `ccall`. I was most interested in seeing whether it would circumvent #28 (it didn't seem to), but I don't think I looked at its runtime implications (esp on a ccall-heavy workload). I suspect that branch is pretty close to usable, if anyone wants to pick it up and run with it.

Obviously spending some time with ProfileView will be useful. I did a ton of that at the beginning and fixed most everything I that I could. But I haven't done it in a while and we may have regressed in places as we expanded our functionality.

KristofferC:
I looked a bit at `teh/compiled_call` and I am stuck at:

```
julia> eval(:(ccall(:jl_alloc_string, $(Ref{String}), (UInt64,), 5)))
"\x01\0\0\0\0"

julia> eval(:(ccall(:jl_alloc_string, $(Ref{String}), $((UInt64,)), 5)))
ERROR: syntax: ccall argument types must be a tuple; try "(T,)"
Stacktrace:
 [1] eval at .\boot.jl:328 [inlined]
 [2] eval(::Expr) at .\client.jl:404
 [3] top-level scope at none:0
```

The second thing is what we are trying to do but the parser doesn't want the value, it want the expression `:((UInt64,))` and I am not sure how to go from the lowered code:

```
julia> f(n) = ccall(:jl_alloc_string, Ref{String}, (UInt64,), n)
f (generic function with 2 methods)

julia> @code_lowered f(5)
CodeInfo(
1 ─ %1 = (Base.cconvert)(Main.UInt64, n)
│   %2 = (Base.unsafe_convert)(Main.UInt64, %1)
│   %3 = $(Expr(:foreigncall, :(:jl_alloc_string), Ref{String}, svec(UInt64), :(:ccall), 1, :(%2), :(%1)))
└──      return %3
)
```

to generating the correct expression for the ccall.

[...]

PR #216, opened 2019-03-26
WIP: also build compiled calls for ccall

KristofferC:
Here are the issues I have found so far:

#### Problems with type parameters:

Trying to use a function that has a type parameter in its return value (there are some debug statements in here).:

```jl
julia>  @interpret Base.unsafe_convert(Ptr{Float64}, rand(5))
(RetType, ArgType) = (Ptr{T}, svec(Any))
def = :(function ccall_jl_array_ptr(arg1, ::Type{T}) where T
      return ccall(:jl_array_ptr, Ptr{T}, (Any,), arg1)
  end)
Calling compiled function JuliaInterpreter.CompiledCalls.ccall_jl_array_ptr with args: Any[[0.0503368, 0.692781, 0.877826, 0.0605047, 0.586223], Float64]
ERROR: ccall return type Ptr should have an element type (not Ptr{<:T})
```

This is what I meant with the comment at https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/206#issuecomment-476583694.

Copy pasting and this works:

```jl
julia> def = :(function ccall_jl_array_ptr(arg1, ::Type{T}) where T
             return ccall(:jl_array_ptr, Ptr{T}, (Any,), arg1)
         end);

julia> eval(def)
ccall_jl_array_ptr (generic function with 1 method)

julia> ccall_jl_array_ptr([1,2], Int64)
Ptr{Int64} @0x0000000008391250
```

#### Failure with toplevel things:

```jl
julia> ex = quote   # in lowered code, cf is a Symbol
           cf = @eval @cfunction(fcfun, Int, (Int, Int))
           ccall(cf, Int, (Int, Int), 1, 2)
       end;

julia> frame = JuliaInterpreter.prepare_thunk(Main, ex)
ERROR: UndefVarError: cf not defined
Stacktrace:
 [1] macro expansion at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:64 [inlined]
 [2] #build_compiled_call!#47(::Symbol, ::Function, ::Expr, ::Symbol, ::Symbol, ::Array{Any,1}, ::Core.CodeInfo, ::Int64, ::Int64, ::Array{Symbol,1}) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\optimize.jl:209
 [3] build_compiled_call!(::Expr, ::Symbol, ::Symbol, ::Array{Any,1}, ::Core.CodeInfo, ::Int64, ::Int64, ::Array{Symbol,1}) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\optimize.jl:190
 [4] optimize!(::Core.CodeInfo, ::Module, ::Array{Symbol,1}) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\optimize.jl:178
 [5] #FrameCode#3(::Bool, ::Bool, ::Type, ::Module, ::Core.CodeInfo) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\types.jl:71
 [6] JuliaInterpreter.FrameCode(::Module, ::Core.CodeInfo) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\types.jl:69
 [7] prepare_thunk(::Module, ::Expr, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:322
 [8] prepare_thunk(::Module, ::Expr, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:335
 [9] prepare_thunk(::Module, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\construct.jl:321
 [10] top-level scope at none:0
```

#### Some other failure:

```jl
julia> ex = quote
           length(code_typed(fcfun, (Int, Int)))
       end;

julia> fcfun(x::Int, y::Int) = 1;

julia> frame = JuliaInterpreter.prepare_thunk(Main, ex);

julia>  JuliaInterpreter.finish_and_return!(frame, true);
(RetType, ArgType) = (Bool, svec())
def = :(function $(Expr(:where, :(ccall_jl_is_in_pure_context())))
      #= C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\optimize.jl:231 =#
      return ccall(:jl_is_in_pure_context, Bool, ())
  end)
Calling compiled function JuliaInterpreter.CompiledCalls.ccall_jl_is_in_pure_context with args: Any[]
(RetType, ArgType) = (Ptr{Nothing}, svec(Any))
def = :(function $(Expr(:where, :(ccall_jl_value_ptr(arg1))))
      #= C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\optimize.jl:231 =#
      return ccall(:jl_value_ptr, Ptr{Nothing}, (Any,), arg1)
  end)
Calling compiled function JuliaInterpreter.CompiledCalls.ccall_jl_value_ptr with args: Any[svec(Int64, Int64)]
Calling compiled function JuliaInterpreter.CompiledCalls.ccall_jl_value_ptr with args: Any[svec(Int64, Int64)]
Calling compiled function JuliaInterpreter.CompiledCalls.ccall_jl_value_ptr with args: Any[svec(Int64, Int64)]
Calling compiled function JuliaInterpreter.CompiledCalls.ccall_jl_value_ptr with args: Any[Ptr{Nothing} @0x0000000005ad07f0]
ERROR: argument tuple type must contain only types
Stacktrace:
 [1] maybe_evaluate_builtin(::Frame, ::Expr, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\builtins-julia1.1.jl:111
 [2] #evaluate_call_recurse!#37(::Bool, ::Function, ::Any, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:208
 [3] evaluate_call_recurse! at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:205 [inlined]
 [4] eval_rhs(::Any, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:371
 [5] step_expr!(::Any, ::Frame, ::Any, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:504
 [6] step_expr! at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:543 [inlined]
 [7] finish!(::Any, ::Frame, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:14
 [8] finish_and_return! at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:29 [inlined]
 [9] #evaluate_call_recurse!#37(::Bool, ::Function, ::Any, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:242
```


KristofferC:
The positive thing

```
julia> const s = "foo";

julia>  @btime @interpret pointer_from_objref(s)
  4.886 μs (44 allocations: 2.38 KiB)

```

vs master

```
julia> @btime @interpret pointer_from_objref(s)
  1.170 ms (503 allocations: 30.91 KiB)
```

KristofferC:
Both of these calls exist in Base:

```
ccall(:jl_value_ptr, Any, (Ptr{Cvoid},), x)
ccall(:jl_value_ptr, Ptr{Cvoid}, (Any,), x)
```

so we cannot just name the wrapper based on the function itself.

KristofferC:
Fixed some stuff and upgraded to a segfault while running the tests.

timholy:
Looks like you've taken this quite far, and many of the former problems are gone.

I inserted some debugging statements. Ran out of time to get any further, but for the out-of-memory-error I saw this during the initial construction:

```
stmt = :($(Expr(:foreigncall, :(:jl_array_grow_end), Nothing, svec(Any, UInt64), :(:ccall), 2, :(_2), :(%2), :(%1))))
methname = :ccall_70e68ad9_99df_4f44_8dfc_af16c88206f5
def = :(function $(Expr(:where, :(ccall_70e68ad9_99df_4f44_8dfc_af16c88206f5(arg1, arg2))))
      #= /home/tim/.julia/dev/JuliaInterpreter/src/optimize.jl:241 =#
      #= /home/tim/.julia/dev/JuliaInterpreter/src/optimize.jl:241 =# @show ccall_70e68ad9_99df_4f44_8dfc_af16c88206f5 arg1 arg2
      #= /home/tim/.julia/dev/JuliaInterpreter/src/optimize.jl:242 =#
      return ccall(:jl_array_grow_end, Nothing, (Any, UInt64), arg1, arg2)
  end)
stmt pre: $(Expr(:foreigncall, :(:jl_array_grow_end), Nothing, svec(Any, UInt64), :(:ccall), 2, :(_2), :(%2), :(%1)))
stmt post: ($(QuoteNode(JuliaInterpreter.CompiledCalls.ccall_70e68ad9_99df_4f44_8dfc_af16c88206f5)))(_2, %2)
ccall_70e68ad9_99df_4f44_8dfc_af16c88206f5 = JuliaInterpreter.CompiledCalls.ccall_70e68ad9_99df_4f44_8dfc_af16c88206f5
arg1 = UInt8[]
arg2 = 0x0000000000000001
```

but for the one that causes the error

```
ccall_70e68ad9_99df_4f44_8dfc_af16c88206f5 = JuliaInterpreter.CompiledCalls.ccall_70e68ad9_99df_4f44_8dfc_af16c88206f5
arg1 = UInt8[0x66, 0x65, 0x76, 0x61, 0x6c, 0x5f, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]
arg2 = 0x00007fffafa3f63f
```

That's rather a lot of elements! Looks like a pointer.

I am now forgetting what's that extra `%1` argument when it was initially creating that ccall? (Compare `stmt pre` and `stmt post`, which come from `update_statement!`.)

KristofferC:
The way I understand it is:

```
julia> f() = ccall(:foo, Cvoid, (Ptr{Float64},Ptr{Int64}), A, B)

julia> @code_lowered f()
CodeInfo(
1 ─ %1 = (Core.apply_type)(Main.Ptr, Main.Float64)
│   %2 = (Base.cconvert)(%1, Main.A)
│   %3 = (Core.apply_type)(Main.Ptr, Main.Int64)
│   %4 = (Base.cconvert)(%3, Main.B)
│   %5 = (Core.apply_type)(Main.Ptr, Main.Float64)
│   %6 = (Base.unsafe_convert)(%5, %2)
│   %7 = (Core.apply_type)(Main.Ptr, Main.Int64)
│   %8 = (Base.unsafe_convert)(%7, %4)
│   %9 = $(Expr(:foreigncall, :(:foo), Nothing, svec(Ptr{Float64}, Ptr{Int64}), :(:ccall), 2, :(%6), :(%8), :(%4), :(%2)))
└──      return %9
)
```

The arguments `%4` and `%2` here are the result of `cconvert` and are only part of the `foreigncall` such that the GC will not think they are unused and free them.
What I think is the correct thing to do is to look at `%4, then go one step backwards and take the second argument from the `cconvert` ( `Main.B`) and that is the argument we should pass to the ccall-wrapper function. Right now, I am directly passing the `unsafe_convert´ value which is not right.

KristofferC:
However, the GC things don't seem to always be there:

```
julia> f(x) =  ccall(:jl_value_ptr, Ptr{Cvoid}, (Any, Ptr{Int64}), x, y)

julia> @code_lowered f(5)
CodeInfo(
1 ─ %1 = (Core.apply_type)(Main.Ptr, Main.Int64)
│   %2 = (Base.cconvert)(%1, Main.y)
│   %3 = (Core.apply_type)(Main.Ptr, Main.Int64)
│   %4 = (Base.unsafe_convert)(%3, %2)
│   %5 = $(Expr(:foreigncall, :(:jl_value_ptr), Ptr{Nothing}, svec(Any, Ptr{Int64}), :(:ccall), 2, :(x), :(%4), :(%2)))
└──      return %5
)
```

So here we only call `cconvert` on `y`. Is it because `Any` is special cased to not have a `cconvert` on them?

KristofferC:
Implemented the idea above. Getting most of the ways now but errors when stepping through test. The error is

```
julia> frame, pc = debug_command(frame, :s)
(Frame for print_to_string(xs...) in Base at strings/io.jl:113
⋮
  27 121  │   %27 = (tuple)(siz)
  28 121  │   %28 = (%26)(%27)
  29 121  │   %29 = (Core.kwfunc)(Base.GenericIOBuffer{Array{UInt8,1}})
  30 121  │         s = (%29)(%28, Base.GenericIOBuffer{Array{UInt8,1}})
  31 122  │   %31 = xs
⋮
xs = (:(a + b),)
siz = 8
#temp# = nothing
x = :(a + b)
caller: string(xs...) in Base at strings/io.jl:156, 29)

julia> frame, pc = debug_command(frame, :s)
ret = getfield(Core, Symbol("#kw#Type"))()
ERROR: AssertionError: ret isa BreakpointRef
Stacktrace:
 [1] maybe_step_through_wrapper!(::Any, ::Frame) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:253
 [2] debug_command(::Any, ::Frame, ::Symbol, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:455
 [3] debug_command at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:480 [inlined] (repeats 2 times)
 [4] top-level scope at none:0
```

timholy:
I'm not really sure, and `foreigncall` is not documented [here](https://docs.julialang.org/en/latest/devdocs/ast/#Expr-types-1). When i was working on the existing ccall implementation I struggled with this a lot but don't really remember the conclusions I came to. I suspect `Any` gets passed to C as a `jl_value_t*`?

KristofferC:
I've gotten this passing by excluding some tricky cases which is when the library has to be looked up... Running the interpreter linearly as is done right now doesn't work because there might be non evaluatable statements in between the foreign call and the definition of the library. We should probably run the Julia test suite on this.

Some performance numbers:

Before:

```
julia>  s = "foo"; @btime @interpret pointer_from_objref(s)
  1.089 ms (503 allocations: 30.94 KiB)

julia>  str = "abc 123\n"^100; @btime @interpret split(str, '\n')
  1.123 s (811505 allocations: 38.22 MiB)

julia> using Plots; @time @interpret plot(rand(5))
 18.342900 seconds (20.03 M allocations: 876.301 MiB, 1.09% gc time)
```


After:

```
julia> s = "foo"; @btime @interpret pointer_from_objref(s)
  4.400 μs (44 allocations: 2.38 KiB)

julia> str = "abc 123\n"^100; @btime @interpret split(str, '\n')
  188.822 ms (442822 allocations: 15.85 MiB)

julia> using Plots; @time @interpret plot(rand(5))
  8.551581 seconds (24.95 M allocations: 1.043 GiB, 2.78% gc time)
```

src/interpret.jl
```diff
@@ -171,14 +171,18 @@ function evaluate_foreigncall(frame::Frame, call_expr::Expr)
...
+# We have to intercept ccalls / llvmcalls before we try it as a builtin
 function bypass_builtins(frame, call_expr, pc)
     if isassigned(frame.framecode.methodtables, pc)
         tme = frame.framecode.methodtables[pc]
         if isa(tme, Compiled)
             fargs = collect_args(frame, call_expr)
             f = to_function(fargs[1])
-            return Some{Any}(f(fargs[2:end]...))
+            if parentmodule(f) === JuliaInterpreter.CompiledCalls
+                return Some{Any}(Base.invokelatest(f, fargs[2:end]...))
```
timholy: I wonder how much this hurts performance. Have you profiled? If the answer is "a lot," [this branch of SnoopCompile](https://github.com/timholy/SnoopCompile.jl/tree/teh/inference_timing) illustrates a dirty-rotten trick in its `__init__` function: we could theoretically set the world age to these constructed methods to be the same as the world age of the method we stole them from.

src/optimize.jl
```diff
@@ -150,42 +151,120 @@ function optimize!(code::CodeInfo, mod::Module)
...
+                @assert length(f.args) == 3
+                lib = f.args[3] isa String ? f.args[3] : f.args[3].value
+                prefix = f.args[2] isa String ? f.args[2] : f.args[2].value
+                f = Symbol(prefix, '_', lib)
+            end
+            # Punt on non literal ccall arguments for now
+            if !(isa(f, String) || isa(f, Symbol) || isa(f, Ptr))
+                continue
+            end
+            # TODO: Only compile one ccall per call and argument types
```
timholy: My "world age trick" above seems incompatible with this also-worthy goal. Might have to choose.

src/optimize.jl
```diff
@@ -150,42 +151,120 @@ function optimize!(code::CodeInfo, mod::Module)
...
+        # The result of this is useful to have next to you when reading this code:
+        # f(x, y) =  ccall(:jl_value_ptr, Ptr{Cvoid}, (Float32,Any), x, y)
+        # @code_lowered f(2, 3)
+        args = []
+        for (atype, arg) in zip(ArgType, stmt.args[6:6+nargs-1])
+            if atype === Any
+                push!(args, arg)
+            else
+                @assert arg isa SSAValue
+                unsafe_convert = code.code[arg.id]
```
timholy: maybe call this `unsafe_convert_expr` or something? Likewise for `cconvert` below.

src/interpret.jl
KristofferC: How do I profile it when it errors without this? :P

KristofferC: So by just calling things twice to get around the first world error I get

```
julia>  s = "foo"; @btime @interpret pointer_from_objref(s)
  3.737 μs (42 allocations: 2.34 KiB)
Ptr{Nothing} @0x0000000009463a10
```

so perhaps a 20% speedup?

timholy: Not worth worrying about then, at least not until other bottlenecks go away.

KristofferC: FWIW, on master:

```
julia> atomic_fence() = Base.llvmcall("""
                                 fence seq_cst
                                 ret void
                                 """, Cvoid, Tuple{})

julia> g() = atomic_fence()

julia> @interpret g()
┌ Warning: likely failure to return to toplevel, try JuliaInterpreter.split_expressions
└ @ JuliaInterpreter C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:580
┌ Warning: likely failure to return to toplevel, try JuliaInterpreter.split_expressions
└ @ JuliaInterpreter C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:580
ERROR: MethodError: no method matching llvmcall_83ebde91_bde5_4587_bba1_52f92157d766()
The applicable method may be too new: running in world age 25606, while current world is 25607.
```

KristofferC:
Here is a standalone example of the sparam problem:

First, a function that does a ccall. We extract the `Ptr{T}` return type:

```jl
julia> my_unsafe_convert(::Type{Ptr{T}}, a::Array{T}) where {T} = ccall(:jl_array_ptr, Ptr{T}, (Any,), a)my_unsafe_convert (generic function with 1 method)

julia> c = (@code_lowered my_unsafe_convert(Ptr{Int}, [1,2])).code
2-element Array{Any,1}:
 :($(Expr(:foreigncall, :(:jl_array_ptr), Ptr{T}, svec(Any), :(:ccall), 1, :(_3))))
 :(return %1)

julia> ptr_t = c[1].args[2]
Ptr{T}
```

Then we create a new function, and interpolate that variable:

```jl
julia> @eval my_unsafe_convert_2(::Type{Ptr{T}}, a::Array{T}) where {T} = ccall(:jl_array_ptr, $(ptr_t), (Any,), a):

julia> my_unsafe_convert_2(Ptr{Int}, [1,2])
ERROR: ccall return type Ptr should have an element type (not Ptr{<:T})
Stacktrace:
 [1] my_unsafe_convert_2(::Type{Ptr{Int64}}, ::Array{Int64,1}) at .\REPL[24]:1
 [2] top-level scope at none:0
```

As can be seen, the `Ptr{T}` we interpolated doesn't "bind" correctly with the `where {T}`. The `T` is probably just a label that is now not matching and I don't know how to update the `Ptr{T}` to work in the second method.

timholy:
```julia
julia> ex = Expr(:curly, ptr_t.name.name, ((tv->tv.name).(ptr_t.parameters))...)
:(Ptr{T})

julia> @eval my_unsafe_convert_3(::Type{Ptr{T}}, a::Array{T}) where {T} = ccall(:jl_array_ptr, $ex, (Any,), a)
my_unsafe_convert_3 (generic function with 1 method)

julia> my_unsafe_convert_3(Ptr{Int}, [1,2])
Ptr{Int64} @0x00007f403128d910
```

However, for generality one might want to handle the typevars to include the upper & lower bounds? Just check for `Union{}` and if it's not, insert a `:<` expr.

KristofferC:
Cool! That made it work for type vars as well. I also moved the mini interpreter to only be used for llvmcalls because I've never seen it needed for `ccall`. Do you have an example?

I didn't really get the point about bounds on the typevars.

KristofferC:
Plots got a bit faster from this I think:

```
julia>  using Plots; @time @interpret plot(rand(5))
  6.114885 seconds (24.14 M allocations: 1023.185 MiB, 2.96% gc time)
```

so from 8 seconds to 6 seconds.

KristofferC:
We are still missing some ccalls:

```
stmt = :(_7 = $(Expr(:foreigncall, :(:memchr), Ptr{UInt8}, svec(Ptr{UInt8}, Int32, UInt64), :(:ccall), 3, :(%28), :(%29), :(%30), :(%26), :(%23), :(%22))))
```

timholy:
> I didn't really get the point about bounds on the typevars.

May not be necessary, but if you had a `TypeVar(:T, Signed, Real)` then you'd need to turn it into `:(Signed<:T<:Real)`. Something like

```julia
function typevar_expr(tv::TypeVar)
    return tv.lb === Union{} ?
               (tv.ub === Union{} ? Expr(:curly, tv.name) :
                                    Expr(:curly, Expr(:(<:), tv.name, type_expr(tv.ub)))) :
               (tv.ub === Union{} ? Expr(:curly, type_expr(tv.lb), tv.name) :
                                    Expr(:curly, Expr(:comparison, type_expr(tv.lb), :(<:), tv.name, :(<:), type_expr(tv.ub))))
end
```
where `type_expr` would deconstruct the type and create an expr in a similar manner.

> I also moved the mini interpreter to only be used for llvmcalls because I've never seen it needed for ccall. Do you have an example?

Nope, it's just there because this code was an unfinished generalization of the handling of `llvmcall`.

timholy:
> We are still missing some ccalls

I may not be getting it, but do you just need to do the same treatment on each element of the `svec`?

KristofferC:
The thing that causes us to miss it is because it is a RHS in an assignment. Adding

```
  if isexpr(stmt, :(=))
            stmt = stmt.args[2]
        end
```` 

before the `isexpr` check for the `:ccall`:

```` 
julia> str = "abc 123\n"^100; @btime @interpret split(str, '\n')
  64.594 ms (394428 allocations: 12.95 MiB)
```

Now we're cooking with gas.

KristofferC:
What do you think about the 1.0 failure... In some way I don't really care to spend too much time on it when it is only for 1.0 and ratherjust disable compiling ccalls with static parameters on 1.0.

timholy:
That's fine, just file an issue as a TODO for anyone who is devoted to 1.0.

KristofferC:
I just fixed it instead (I think). Perhaps you can do a final review when you have time @timholy.

KristofferC:
Seeing this error with this branch

```
julia>  @interpret CSV.read("../SampleCSVFile_11kb.csv");
ERROR: type SlotNumber has no field value
Stacktrace:
 [1] getproperty(::Any, ::Symbol) at .\sysimg.jl:18
 [2] optimize!(::Core.CodeInfo, ::Method) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\optimize.jl:177
 [3] #FrameCode#1(::Bool, ::Bool, ::Type, ::Method, ::Core.CodeInfo) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\types.jl:70
```

KristofferC:
This is a bit interesting:

```
julia> @time @interpret plot(rand(5))
  8.857005 seconds (32.78 M allocations: 1.417 GiB, 2.96% gc time)

julia> @time @interpret plot(rand(5))
  6.373619 seconds (24.36 M allocations: 1.009 GiB, 3.94% gc time)

julia> @time @interpret plot(rand(5))
  3.156926 seconds (13.57 M allocations: 485.171 MiB, 3.34% gc time)
```

I wonder why the third call is so much faster.

timholy:
The inference-snooping branch of SnoopCompile says there's 69 inferred methods on the 2nd call and only one on the 3rd. They're mostly of this form:
```julia
...
MethodInstance for collect_to_with_first!(::Array{typeof(Base.index_shape),1}, ::Function, ::Base.Generator{Array{Any,1},getfield(JuliaInterpreter, Symbol("##41#44"))}, ::Int64)
MethodInstance for collect_to_with_first!(::Array{typeof(Base.afoldl),1}, ::Function, ::Base.Generator{Array{Any,1},getfield(JuliaInterpreter, Symbol("##41#44"))}, ::Int64)
...
```
and one of the form
```julia
MethodInstance for collect_to!(::Array{Union{typeof(Base.argtail), Missing},1}, ::Base.Generator{Array{Any,1},getfield(JuliaInterpreter, Symbol("##41#44"))}, ::Int64, ::Int64)
```

src/optimize.jl
```diff
@@ -150,42 +151,131 @@ function optimize!(code::CodeInfo, mod::Module)
...
+            nargs = stmt.args[5]
+            build_compiled_call!(stmt, methname, :ccall, stmt.args[1:3], code, idx, nargs, sparams)
+            methodtables[idx] = Compiled()
         end
     end
 
     return code, methodtables
 end
+
+function parametric_type_to_expr(t::Type)
```
timholy: I'd guess you should put `@nospecialize(t::Type)` and haul out the anonymous function on `tv` and use similar treatment.

KristofferC:
I think it is https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/226.

KristofferC:
I'll fix the coverage and then I plan to merge.

KristofferC:
Wait, the coverage report is completely bonkers. llvmcall is still being tested.

PR #227, opened 2019-03-28
fix bug when passing static parameters

KristofferC:
Having a bit of a hard time coming up with a test.


KristofferC:
I can't find a single such call in Base and since we only compile when the library is static I can't hack around it with a `cfunction`. So i'll just merge this.

PR #230, opened 2019-03-28
Revert compiled ccalls

KristofferC:
Running the julia tests `Dates/period` I see `MethodError(Base.unsafe_convert, (Ptr{Nothing}, svec()), 0x00000000000065d5)`.

Getting late here so will look at it tomorrow


KristofferC:
Amazing... This is one for the grand children. So the MWE is something along

```
using Dates
using Test

y = Dates.Year(1)
m = Dates.Month(1)
d = Dates.Day(1)
h = Dates.Hour(1)
ms = Dates.Millisecond(1)
emptyperiod = ((y + d) - d) - y
@testset "a" begin
    BigFloat(1)
end

@testset "b" begin
    pa = [1y 1m; 1h 1m]
end
```

Removing the `@testset` makes the bug go away, removing the `BigFloat` call makes the bug go away etc. A thing to keep in mind about about `@testset` is that it sets the RNG seed.

Ffirst we call `BigFloat` with a fixed seed and end up compiling the `ccall(:jl_value_ptr, Any, (Ptr{Cvoid},), x)` wrapper function where the name of the wrapper is computed as

```
uuid = uuid4()
ustr = replace(string(uuid), '-'=>'_')
methname = Symbol("ccall", '_', f, '_', ustr)
```

Note that `uuid4()` is a function only based on the rng seed.

Then we go into the second `@testset`. The `@testset` resets the seed to be *the same one as the one in the previous testset*. Now we want to compile the call `ccall(:jl_value_ptr, Ptr{Cvoid}, (Any,), x)` which has different types (but the same function name). Since the testset reset the seed, the `uuid4()` call will be the same name and we will clobber the previous function with the new definition.

The solution is to call `uuid1()` which also takes into account time or use a hash based on the argument types..

PR #231, opened 2019-03-29
use non rng seed dependent uuid to avoid collisions

KristofferC:
See https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/230#issuecomment-477906104

I have not been able to have this situation occur artificially, even by explicitly trying to set the RNG seed.

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -95,6 +95,8 @@ function lookup_global_refs!(ex::Expr)
     return nothing
 end
 
+const rng = MersenneTwister()
+
 """
     optimize!(code::CodeInfo, mod::Module)
 
@@ -161,7 +163,7 @@ function optimize!(code::CodeInfo, scope)
             # Check for :llvmcall
             arg1 = stmt.args[1]
             if (arg1 == :llvmcall || lookup_stmt(code.code, arg1) == Base.llvmcall) && isempty(sparams) && scope isa Method
-                uuid = uuid1()
+                uuid = uuid1(rng)
                 ustr = replace(string(uuid), '-'=>'_')
                 methname = Symbol("llvmcall_", ustr)
                 nargs = length(stmt.args)-4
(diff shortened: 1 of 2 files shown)


src/optimize.jl
```diff
@@ -161,7 +161,7 @@ function optimize!(code::CodeInfo, scope)
             # Check for :llvmcall
             arg1 = stmt.args[1]
             if (arg1 == :llvmcall || lookup_stmt(code.code, arg1) == Base.llvmcall) && isempty(sparams) && scope isa Method
-                uuid = uuid4()
+                uuid = uuid1()
```
fredrikekre: Maybe use an internal rng such that random numbers are the same for the user?
```suggestion
                uuid = uuid1(rng)
```

src/optimize.jl
```diff
@@ -171,7 +171,7 @@ function optimize!(code::CodeInfo, scope)
         elseif isexpr(stmt, :foreigncall) && scope isa Method
             f = lookup_stmt(code.code, stmt.args[1])
             if isa(f, Ptr)
-                f = string(uuid4())
+                f = string(uuid1())
```
fredrikekre:
```suggestion
                f = string(uuid1(rng))
```

src/optimize.jl
```diff
@@ -184,7 +184,7 @@ function optimize!(code::CodeInfo, scope)
                 continue
             end
             # TODO: Only compile one ccall per call and argument types
-            uuid = uuid4()
+            uuid = uuid1()
```
fredrikekre:
```suggestion
            uuid = uuid1(rng)
```

src/optimize.jl
KristofferC: What's the advantage?

fredrikekre: You get different random numbers if you interpret or not?

timholy:
I like the idea of maintaining our own `rng`.

src/optimize.jl
pfitzseb: Yeah that's a good idea.

KristofferC:
Updated to use our own RNG.

PR #232, opened 2019-03-29
type pirate the nop Some conversion fix

KristofferC:


timholy:
Note: do not merge until we get a version number from https://github.com/JuliaLang/julia/pull/31536.

I'd advocate merging that Julia PR pretty quickly just so you can get the VERSION number here.

KristofferC:
Added VERSION

KristofferC:
Ok to go?

PR #235, opened 2019-03-29
fix breakpoint on first line of function call

pfitzseb:
This fixes https://github.com/JunoLab/Juno.jl/issues/273 and https://github.com/JuliaDebug/Debugger.jl/issues/134, but I think the `shouldbreak(frame, frame.pc) && return BreakpointRef(frame.framecode, frame.pc)` guard statement should be in more of those functions, maybe.


pfitzseb:
Hm, doesn't seem like the best approach considering that tests fail. Will fix in Juno and Debugger.jl instead.

PR #237, opened 2019-03-29
Delete leftover cconvert and unsafe_convert after moving foreigncall to wrapper function

KristofferC:
Get rid of the unnecessary function that are left after we move the foregincall to the wrapper function.

Using

```jl
julia> const ss = "foo"
"foo"

julia> p = pointer_from_objref(ss)
Ptr{Nothing} @0x0000000007f1b2d0
```

Before

```jl
julia> @btime @interpret unsafe_pointer_to_objref(p)
  6.100 μs (54 allocations: 2.72 KiB)
"foo"
```

After

```jl
julia> @btime @interpret unsafe_pointer_to_objref(p)
  4.543 μs (40 allocations: 2.20 KiB)
"foo"
```

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -56,6 +56,19 @@ function renumber_ssa!(stmts::Vector{Any}, ssalookup)
     return stmts
 end
 
+function compute_ssa_mapping_delete_statements!(code::CodeInfo, stmts::Vector{Int})
+    stmts = unique!(sort!(stmts))
+    ssalookup = collect(1:length(code.codelocs))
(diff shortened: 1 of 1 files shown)


KristofferC:
Rebased

KristofferC:
> I guess the alternative is to start with delete_flag = falses(length(code.code)) and then set the corresponding entry to true. But given that we don't expect many of these per frame, I think this is perhaps better.

At least for now. We can always refactor it if we notices that we are adding more of these "passes" over time.

Issue #154, opened 2019-03-15
MWE for deepcopy failure


macd:
It looks like ```deepcopy``` does not like ```Expr```s to much.  It can either result in an undefined reference in the AST or a segfault.  Details below.
```
include("utils.jl")

module CopyTest
using Random
using Test
end

m = CopyTest

ex = quote
    #q = :(x = 55)      # this will segfault
    q = QuoteNode([1])  # results in "syntax: undefined reference in AST"
    g = deepcopy(q)
end

nstmts = 100000

modexs, _ = JuliaInterpreter.split_expressions(m, ex)
for modex in modexs
    #dump(modex)
    frame = JuliaInterpreter.prepare_thunk(modex)
    nstmtsleft = nstmts
    while true
        ret, nstmtsleft = evaluate_limited!(frame, nstmtsleft, true)
        if isa(ret, Aborted)
            JuliaInterpreter.finish_stack!(Compiled(), frame, true)
            break
        elseif isa(ret, Some)
            break
        end
    end
end
```

KristofferC:
Smaller MWE:

```
julia> q = QuoteNode([1])
:($(QuoteNode([1])))

julia> @interpret deepcopy(q)
ERROR: syntax: undefined reference in AST
Stacktrace:
 [1] eval at .\boot.jl:328 [inlined]
 [2] evaluate_foreigncall(::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:160
 [3] eval_rhs(::Any, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:348
 [4] step_expr!(::Any, ::Frame, ::Any, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:479
 [5] step_expr! at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:518 [inlined]
 [6] finish!(::Any, ::Frame, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\commands.jl:14
```

PR #239, opened 2019-03-29
add a test for deepcopy on QuoteNode

KristofferC:
Closes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/154


timholy:
Nice that an indirect effect of #216 is fixing bugs!

KristofferC:
Yep!

PR #243, opened 2019-03-29
Micro-optimizations that reduce dynamic dispatch

timholy:
Before:
```julia
julia> @time (p = @interpret(plot(rand(5))); @interpret(display(p)))
 14.597646 seconds (72.11 M allocations: 2.350 GiB, 3.33% gc time)
```
After:
```julia
julia> @time (p = @interpret(plot(rand(5))); @interpret(display(p)))
 11.713349 seconds (69.40 M allocations: 2.286 GiB, 3.02% gc time)
```
This is running on top of https://github.com/JuliaLang/julia/pull/31537.


KristofferC:
Nice!

PR #245, opened 2019-03-30
go back to using a Vector for recycled frames

KristofferC:
Before

```jl
julia> @time (p = @interpret(plot(rand(5))); @interpret(display(p)))
 12.059313 seconds (63.78 M allocations: 2.143 GiB, 2.58% gc time)

julia> @time (p = @interpret(plot(rand(5))); @interpret(display(p)))
 12.198480 seconds (65.08 M allocations: 2.175 GiB, 2.66% gc time)
```

After

```jl
julia> @time (p = @interpret(plot(rand(5))); @interpret(display(p)))
  9.768135 seconds (65.32 M allocations: 2.052 GiB, 2.97% gc time)

julia> @time (p = @interpret(plot(rand(5))); @interpret(display(p)))
  9.531303 seconds (64.33 M allocations: 2.011 GiB, 3.06% gc time)
```

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -20,8 +20,13 @@ rather than recursed into via the interpreter.
 """
 const compiled_methods = Set{Method}()
 
-const junk = Base.IdSet{FrameData}()      # to allow re-use of allocated memory (this is otherwise a bottleneck)
-recycle(frame) = push!(junk, frame.framedata)  # using an IdSet ensures that a frame can't be added twice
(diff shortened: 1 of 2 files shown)


KristofferC:
I wonder if the nightly error is due to this PR... It shouldn't be.

KristofferC:
`bypass_builtins at C:\projects\juliainterpreter-jl\src\interpret.jl:182` sounds like it is something with the compiled ccall

PR #247, opened 2019-03-30
add some debug output to diagnose AV failure

KristofferC:


KristofferC:
We are failing here: https://github.com/JuliaLang/julia/blob/ad40952896d134182c64aa4f26f8eb9290c3dc27/base/file.jl#L425. But nothing in the output explains to me why we are failing.

We are, AFAIU, passing the correct arguments to the wrapper function and the lowered form of the function also looks good.

I have also not been able to repro this locally, running the test suite over and over.

PR #251, opened 2019-03-31
Fix for change in CodeInfo.slotnames type on julia master

timholy:
Trivial, but I suspect we may want to tag 0.4.1 fairly soon.


KristofferC:
Add a TODO to remove this completely after support for 1.1 is dropped?

timholy:
> Add a TODO to remove this completely after support for 1.1 is dropped?

That'd be fine, but are we expecting to need to drop support for any Julia 1.x versions? So far I think it's not been hard to keep supporting them all.

KristofferC:
Probably not. The comment might just be distracting. Fine either way :)

PR #254, opened 2019-03-31
More optimizations and a fix for `is_vararg_type`

timholy:
These are all low-level optimizations. The cumulative effect is to go from ~18us/iteration in the `summer` benchmark to ~16us/iteration. One of the commits, the `is_vararg_type`, actually leads to a 3% hit on that benchmark, but it fixes what seems to be an outright mistake. (Before 22a8b2c it was not possible to step into any function that received `Union{}` as an argument; the case that I observed was for `promote_result`.)

On the `summer` benchmark, it seems the only remaining case of dynamic dispatch arises from https://github.com/JuliaLang/julia/issues/31565.

While performance improvements are always nice, my real goal in doing this is to prepare the way for a trial of the effect of lowered-code inlining (ref. https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/204#issuecomment-478046381). Since I am hoping it has a dramatic effect, it would be a shame if it were "spoiled" by pre-existing bottlenecks, so I thought I'd start by getting rid of them even though currently they aren't terribly important.

To develop inlining, I think the best approach would be to first take a simple case, like `summer`, and "inline by hand" to see if it provides as much improvement as I hope. If so, then it will be worth building out the more general machinery.

### DIFF
--- a/src/generate_builtins.jl
+++ b/src/generate_builtins.jl
@@ -22,8 +22,14 @@ function nargs(f, table, id)
         maxarg = typemax(Int)
     end
     # The tfunc tables are wrong for fptoui and fptosi (fixed in https://github.com/JuliaLang/julia/pull/30787)
-    if f == "Base.fptoui" || f == "Base.fptosi"
+    if f == Base.fptoui || f == Base.fptosi
         minarg = 2
+    end
+    # Specialize arrayref and arrayset for small numbers of arguments
+    if f == Core.arrayref
+        maxarg = 5
+    elseif f == Core.arrayset
+        maxarg = 6
     end
     return minarg, maxarg
 end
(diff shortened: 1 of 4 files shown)


KristofferC:
What is the `summer` benchmark?

timholy:
Sorry, it's just https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/cef53507b4b409547b9adbd060fe881319fd82d9/test/interpret.jl#L8-L14 with a `Float64` array of large enough size (e.g., 10^5) that initialization isn't a big deal. I usually do
```julia
frame = JuliaInterpreter.enter_call(summer, A)
@elapsed(JuliaInterpreter.finish_and_return!(frame))/length(A)
```

PR #256, opened 2019-04-01
WIP: file:line breakpoints without Revise

KristofferC:
The current breakpoint system for file:line is "eager" in the sense that it wants to insert breakpoints directly into the `FrameCode` when the breakpoint is created. This means that we need to know the full signature of the method at the location at breakpoint insertion time. We currently know that only if the file is tracked by Revise. It can sometimes be a bit unintuitive that you need to load Revise, then load a file with `includet` before you can breakpoint in it.

This PR provides a fallback in the case where we do not find any signatures at a given location. Instead of just erroring, we create something called an `UnresolvedBreakpoint` which we, at `FrameCode` creation time look through a list of to see if it is inside our method body. In that case we add the breakpoint.
We also need to handle the case where a breakpoint is inserted when the `FrameCode` is already cached. We do that by, at `FrameCode` creation, store a lookup from a file and line range to a `FrameCode` and upon breakpoint insertion, first look through this lookup and see if we have a matching `FrameCode`.

For example, a possible use case is:

```jl
julia> breakpoint("../foo.jl", 3)

julia> include("../foo.jl")

julia> @interpret foo(2)
hello
(Frame for foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:2
  1 2  1 ─     (println)("hello")
b 2 3  │       (sin)(2.0)
  3 4  └──     return 5
x = 2, breakpoint(foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:2, line 3))

julia> breakpoint("../foo.jl", 2)

julia> @interpret foo(2)
(Frame for foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:2
b 1 2  1 ─     (println)("hello")
b 2 3  │       (sin)(2.0)
  3 4  └──     return 5
x = 2, breakpoint(foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:2, line 2))
```

TODO:
   - [ ] Make it "Revise safe". Might need your help here @timholy 
   - [ ] Add tests for Base methods since their paths are a bit different
   - [ ] Expose unresolved breakpoints to the API somehow

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -1,25 +1,157 @@
-const _breakpoints = BreakpointRef[]
+const Condition = Union{Nothing,Expr,Tuple{Module,Expr}}
+abstract type AbstractBreakpoint end
+
+
+const _breakpoints = AbstractBreakpoint[]
 breakpoints() = copy(_breakpoints)
-const unresolved_breakpoints = Set{UnresolvedBreakpoint}()
+
+function add_to_existing_framecodes(bp::AbstractBreakpoint)
+    for framecode in values(framedict)
+        add_breakpoint_if_match!(framecode, bp)
+    end
+end
+
+function add_breakpoint_if_match!(framecode::FrameCode, bp::AbstractBreakpoint)
+    if framecode_matches_breakpoint(framecode, bp)
+        stmtidx = bp.line === 0 ? 1 : statementnumber(framecode, bp.line)
+        bpref = breakpoint!(framecode, stmtidx, bp.condition)
+        push!(bp.applications, BreakpointRef(framecode, stmtidx))
+    end
+end
+
+struct BreakpointSignature <: AbstractBreakpoint
+    f # Method or function
+    sig::Union{Nothing, Type}
+    line::Int # 0 is a sentinel for first statement
+    condition::Condition
+    applications::Vector{BreakpointRef}
+end
+Base.isequal(bp2::BreakpointSignature, bp::BreakpointSignature) = 
+    bp2.f == bp.f && bp2.sig == bp.sig && bp2.line == bp.line && bp.condition == bp2.condition
+function Base.show(io::IO, bp::BreakpointSignature)
+    print(io, "BreakpointSignature: ")
+    print(io, bp.f)
+    if bp.sig !== nothing
+        print(io, '(', join("::" .* string.(bp.sig.types), ", "), ')')
+    end
+    if bp.line !== 0
+        print(io, bp.line)
+    end
+    if bp.condition !== nothing
+        print(io, " ", bp.condition)
+    end
+end
+
+function extract_function_from_method(m::Method)
+    sig = m.sig
+    while sig isa UnionAll
+        sig = sig.body
+    end
+    f = sig.parameters[1]
+    if isdefined(f, :instance)
+        return f.instance
+    else
+        return f
+    end
+end
+
+function framecode_matches_breakpoint(framecode::FrameCode, bp::BreakpointSignature)
+    framecode.scope isa Method || return false
+    meth = framecode.scope
+    bp.f isa Method && return meth === bp.f
+    bp.f === extract_function_from_method(meth) || return false
+    bp.sig === nothing && return true
+    tt = Base.signature_type(bp.f, bp.sig)
+    return whichtt(tt) === meth
+end
+
+"""
+    breakpoint(f)
+    breakpoint(f, line)
+    breakpoint(f, condition)
+    breakpoint(f, line, condition)
+    breakpoint(f, sig)
+    breakpoint(f, sig, line)
+    breakpoint(f, sig, condition)
+    breakpoint(f, sig, line, condition)
+
+
+Add a breakpoint to `f` with the specified argument types `sig`.¨
+If `sig` is not given, the breakpoint will apply to all methods of `f`.
+If `f` is a method, the breakpoint will only apply to that method.
+Optionally specify an absolute line number `line` in the source file; the default
+is to break upon entry at the first line of the body.
+Without `condition`, the breakpoint will be triggered every time it is encountered;
+the second only if `condition` evaluates to `true`.
+`condition` should be written in terms of the arguments and local variables of `f`.
+
+# Example
+```julia
+function radius2(x, y)
+    return x^2 + y^2
+end
+
+breakpoint(radius2, Tuple{Int,Int}, :(y > x))
+```
+"""
+function breakpoint(f, sig=nothing, line::Integer=0, condition::Condition=nothing)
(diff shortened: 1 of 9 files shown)


pfitzseb:
This seems like a very elegant solution! 
Can we expose those unresolved breakpoints by also returning them in `JuliaInterpreter.breakpoints()`? I think that's necessary for a good UX, e.g. when setting breakpoints in Juno (I mean I could keep my own list of all breakpoints there, but that's annoying).

KristofferC:
> Can we expose those unresolved breakpoints by also returning them in JuliaInterpreter.breakpoints()? I think that's necessary for a good UX, e.g. when setting breakpoints in Juno (I mean I could keep my own list of all breakpoints there, but that's annoying).

Yeah, we need to do something like that. I forgot to add it to the todo list.

src/types.jl
```diff
@@ -81,7 +81,21 @@ function FrameCode(scope, src::CodeInfo; generator=false, optimize=true)
...
+    if scope isa Method && !is_generated(scope) # TODO: Relax this
+        method = scope
+        for bp in unresolved_breakpoints
+            method.file === bp.file || continue
+            if method_contains_line(method, bp.line)
+                stmtidx = statementnumber(framecode, bp.line)
+                breakpoint!(framecode, stmtidx, bp.condition)
+            end
+        end
+        lineranges = get!(() -> Pair{UnitRange, FrameCode}[], framecode_locations, method.file)
```
timholy: This line and the next need a comment, it took me a bit to understand what this was about.

src/types.jl
```diff
@@ -81,7 +81,21 @@ function FrameCode(scope, src::CodeInfo; generator=false, optimize=true)
         end
     end
     used = find_used(src)
-    return FrameCode(scope, src, methodtables, breakpoints, used, generator)
+    framecode = FrameCode(scope, src, methodtables, breakpoints, used, generator)
+    if scope isa Method && !is_generated(scope) # TODO: Relax this
+        method = scope
+        for bp in unresolved_breakpoints
+            method.file === bp.file || continue
```
timholy: This is fragile (not just for Base but also wrt symlinks for stdlibs, etc) but I think you know that already. Other than fixing those cases I think we just have to :crossed_fingers:

src/breakpoints.jl
```diff
@@ -242,31 +243,34 @@ end
...
     try
         sigs = signatures_at(filename, line)
     catch
         sigs = nothing
     end
     if sigs === nothing
-        # TODO: build a Revise-free fallback. Note this won't work well for methods with keywords.
-        error("no signatures found at $filename, $line.\nRestarting and `using Revise` and the relevant package may fix this problem.")
+        filename = Symbol(abspath(filename))
+        if haskey(framecode_locations, filename)
```
timholy: Does this loop ever run if Revise is running? Needs a comment.

test/breakpoints.jl
```diff
@@ -209,6 +209,24 @@ end
...
+        b = z^a
+        return a + b
+    end
+    """)
+    close(io)
+    breakpoint(path, 3)
+    include(path)
+    frame, bp = @interpret somefunc(2, 3)
+    @test JuliaInterpreter.whereis(frame) == (path, 3)
+    breakpoint(path, 2)
```
timholy: Does this one break in the wrapper or the body function?

Could also use a test on a kwarg method and on one-line definitions with optional and keywargs.

KristofferC:
A drawback with the eager version we do now is that they are not robust against method redefinitions (e.g. from Revise). For example (with some debug info printed):

```
julia> includet("../foo.jl")

julia> breakpoint(foo)
Creating FrameCode for foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:4
1-element Array{BreakpointRef,1}:
 breakpoint(foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:4, line 4)

julia> @interpret foo(2)
(Frame for foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:4
b 1 4  1 ─     (print)("df")
  2 5  │       (print)("helloss")
  3 7  │       (sin)(2.0)
⋮
x = 2, breakpoint(foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:4, line 4))

julia> # edit foo in foo.jl

julia> @interpret foo(2)
Creating FrameCode for foo(x) in Main at C:\Users\Kristoffer\Debugging\foo.jl:4
dfhellosss5
```

The Revise redefinition meant that we created a new `FrameCode` and the breakpoint is thus "lost".
Having all breakpoints be "lazy" like in this PR would fix some of that. Thoughts?

KristofferC:
Basically, what I am thinking is something like:

```jl
struct BreakpointSignature <: AbstractBreakpoint
    sig
    condition
    application::Vector{Pair{FrameCode, Int}} # FrameCode => stmtidx
end
breakpoint(sig, condition) = push!(_breakpoints, BreakpointSignature(sig, condition, [])

struct BreakpointLocation  <: AbstractBreakpoint
    file
    line
    condition
    application::Vector{Pair{FrameCode, Int}} # FrameCode => stmtidx
end

breakpoint(file, line, condition) = push!(_breakpoints, BreakpointLocation(file, line, condition, [])
```

Everytime we create a `FrameCode` we go through `_breakpoints` and see if the breakpoint applies. In that case, we push to that breakpoints `application` field the `FrameCode` and `stmtidx` where the breakpoint applied (so we can disable and toggle etc).

Toggling a breakpoint is then implemented as

```
function toggle(bp::AbstractBreakpoint)
    for (framecode, stmtidx) in bp.applications
        toggle!(framecode.breakpoints[stmtidx])
    end
end
```

KristofferC:
Also, I think this already is Revise safe since:

>  At that point I worry that the data in framecode_locations could become problematically wrong.

Does not happen since we create new FrameCode for new methods. I guess it might happen if people only change whitespace and Revise makes some optimization to only update line info but not update the method.

timholy:
> A drawback with the eager version we do now is that they are not robust against method redefinitions (e.g. from Revise).

You're right wrt breakpoints. Wrt to code, I already took measures to prevent julia265-like problems: https://github.com/timholy/Revise.jl/blob/f744f86cd0183fc41c1f9625eb608026284c2d3e/src/Revise.jl#L237-L245

It's not obvious to me that we want to keep the breakpoints when the method is redefined. The method, after all, is different; who's to say that this particular line number is special in some way? When I specified the line number I presumably did so with a specific intent (e.g., "right at the beginning of the inner loop"), and the new method may mean the line number does not align with the intent.

KristofferC:
Superseded  by https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/259.

Issue #257, opened 2019-04-01
The interpreter does not handle SSAValues that occur in code correctly


KristofferC:
Edit: Updated to an even more minimum MWE...

We are handling SSAValues in code erroneously:

```jl
function f()
    z = [Core.SSAValue(5),]
    @show z[1]
end
```

```jl
julia> f();
z[1] = :(%5)

julia> @interpret f();
ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getindex at .\array.jl:729 [inlined]
 [2] lookup_var at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:8 [inlined]
 [3] #collect_args#35(::Bool, ::Function, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:64
 [4] collect_args at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:139 [inlined]
 [5] #evaluate_call_recurse!#39(::Bool, ::Function, ::Any, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:211
 [6] evaluate_call_recurse! at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:205 [inlined]
 [7] eval_rhs(::Any, ::Frame, ::Expr) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:371
 [8] step_expr!(::Any, ::Frame, ::Any, ::Bool) at C:\Users\Kristoffer\Debugging\JuliaInterpreter\src\interpret.jl:504
```

KristofferC:
What happens is along the lines that we have the expression

```jl
julia> JuliaInterpreter.pc_expr(frame)
:(($(QuoteNode(Core._apply)))($(QuoteNode(print)), %2, _2))
```

We then lookup the variables:

```jl
call_expr = :((print)(Base.TTY(Base.Libc.WindowsRawSocket(0x00000000000002b0) open, 0 bytes waiting), %5))
```

And then we try to *again*  resolve the `%5` but that is a value now.

KristofferC:
So I don't think `_apply` should do `@lookup` but even after fixing this there is some problems. I don't think our handling of `_apply` is correct, for example:

```
julia> @code_lowered print(1)
CodeInfo(
[...]

PR #258, opened 2019-04-01
handle SSAValues better

KristofferC:
Not sure if this is the only fix needed.

Ref https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/257


timholy:
A more thorough option would be to (1) do a search/replace `SSAValue`->`SSAWrapper` throughout JuliaInterpeter and then (2) in `optimize!` replace all the `SSAValues` that lowering generates with `SSAWrapper`s. Or to be really confusing, we could call it `SSAValue`, define our own such type, and not import `Core.SSAValue` :smile:.

This would be very safe. The disadvantage is that the frame utilities in Core.Compiler become less useful; one might have to replicate a bunch of internal code so that printing, `find_used`, etc work properly.

PR #259, opened 2019-04-02
RFC: refactor breakpoints

KristofferC:
The current breakpoint system immediately "lowers" the high level breakpoints given by a user (a high level breakpoint is for example a breakpoint for a method, or for a file:line) into a set of "lower level breakpoints" (given by a statement index into a FrameCode). This has a major drawback:

- After breakpoint insertion we do not know why a breakpoint was set in a FrameCode. Was it because it was a file:line breakpoint or because it was set on a method or a function etc?
    - This makes it hard for a UI to show the active breakpoints. If a user sets a breakpoint on `convert` you don't want to show 60 breakpoints. From a users p.o.v, they added one breakpoint. That breakpoint can apply to multiple places, but it is still one. A user likely want to toggle / remove all the places where this breakpoint applies collectively.
    - We cannot insert breakpoint in new `FrameCode`s that are being created. That means we lose breakpoints if Revise decide to reevaluate a method because e.g. you fixed your bug and changed a `+` to a `-`. It means you can't set a breakpoint on `foo` and define a new `foo` and have the breakpoint apply.
    - For file:line breakpoints, we need to know at insertion time what signature exists at that point. This can be difficult and currently needs Revise

This PR introduces two "high level breakpoint" structs called `BreakpointSignature` (for breakpoints on methods, signatures etc) and `BreakpointFileLocation` for breakpoints on file locations. Each `breakpoint` invocation will add *one* of these to a list of breakpoints `_breakpoints`. Upon adding a breakpoint, we look through all `FrameCode` we have created so far and insert the breakpoint if it matches. Upon creation of `FrameCode`'s we insert the breakpoint if it matches. Each breakpoint struct has a field `applications::Vector{BreakpointRef}` which is added to as soon as we add a breakpoint to a `FrameCode`. This allows toggling, removal etc by just looping through this list.

This PR should be fairly non-breaking (doesn't break Debugger, but likely breaks Juno).

This PR enables the file:line breakpoints unconditionally (aka even if Revise is not loaded).

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -1,24 +1,102 @@
-const _breakpoints = BreakpointRef[]
+const _breakpoints = AbstractBreakpoint[]
 breakpoints() = copy(_breakpoints)
 
-Base.getindex(bp::BreakpointRef) = bp.framecode.breakpoints[bp.stmtidx]
-function Base.setindex!(bp::BreakpointRef, isactive::Bool)
-    bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(isactive, bp[].condition)
-end
-function toggle!(bp::BreakpointRef)
-    state = bp[]
-    bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(!state.isactive, state.condition)
+function add_to_existing_framecodes(bp::AbstractBreakpoint)
+    for framecode in values(framedict)
+        add_breakpoint_if_match!(framecode, bp)
+    end
 end
 
-function add_breakpoint(framecode, stmtidx)
-    bp = BreakpointRef(framecode, stmtidx)
-    # Since there can be only one BreakpointState for a given framecode/stmtidx,
-    # check whether _breakpoints is already storing a reference to that location
-    idx = findfirst(isequal(bp), _breakpoints)
-    if idx === nothing
-        push!(_breakpoints, bp)
+function add_breakpoint_if_match!(framecode::FrameCode, bp::AbstractBreakpoint)
+    if framecode_matches_breakpoint(framecode, bp)
+        stmtidx = bp.line === 0 ? 1 : statementnumber(framecode, bp.line)
+        breakpoint!(framecode, stmtidx, bp.condition, bp.enabled[])
+        push!(bp.applications, BreakpointRef(framecode, stmtidx))
     end
+end
+
+function framecode_matches_breakpoint(framecode::FrameCode, bp::BreakpointSignature)
+    function extract_function_from_method(m::Method)
+        sig = Base.unwrap_unionall(m.sig)
+        ft0 = sig.parameters[1]
+        ft = Base.unwrap_unionall(ft0)
+        if ft <: Function && isa(ft, DataType) && isdefined(ft, :instance)
+            return ft.instance
+        elseif isa(ft, DataType) && ft.name === Type.body.name
+            f = ft.parameters[1]
+        else
+            return ft
+        end
+    end
+
+    framecode.scope isa Method || return false
+    meth = framecode.scope
+    bp.f isa Method && return meth === bp.f
+    bp.f === extract_function_from_method(meth) || return false
+    bp.sig === nothing && return true
+    return bp.sig <: meth.sig
+end
+
+"""
+    breakpoint(f, [sig], [line], [condition])
+
+Add a breakpoint to `f` with the specified argument types `sig`.¨
+If `sig` is not given, the breakpoint will apply to all methods of `f`.
+If `f` is a method, the breakpoint will only apply to that method.
+Optionally specify an absolute line number `line` in the source file; the default
+is to break upon entry at the first line of the body.
+Without `condition`, the breakpoint will be triggered every time it is encountered;
+the second only if `condition` evaluates to `true`.
+`condition` should be written in terms of the arguments and local variables of `f`.
+
+# Example
+```julia
+function radius2(x, y)
+    return x^2 + y^2
+end
+
+breakpoint(radius2, Tuple{Int,Int}, :(y > x))
+```
+"""
+function breakpoint(f, sig=nothing, line::Integer=0, condition::Condition=nothing)
+    sig !== nothing && (sig = Base.to_tuple_type(sig))
+    bp = BreakpointSignature(f, sig, line, condition, Ref(true), BreakpointRef[])
+    add_to_existing_framecodes(bp)
+    idx = findfirst(bp2 -> same_location(bp, bp2), _breakpoints)
+    idx === nothing ? push!(_breakpoints, bp) : (_breakpoints[idx] = bp)
     return bp
+end
+breakpoint(f, sig, condition::Condition) = breakpoint(f, sig, 0, condition)
+breakpoint(f, line::Integer, condition::Condition=nothing) = breakpoint(f, nothing, line, condition)
+breakpoint(f, condition::Condition) = breakpoint(f, nothing, 0, condition)
+
+
+"""
+    breakpoint(file, line, [condition])
+
+Set a breakpoint in `file` at `line`. The argument `file` can be a filename, a partial path or absolute path.
+For example, `file = foo.jl` will match against all files with the name `foo.jl`,
+`file = src/foo.jl` will match against all paths containing `src/foo.jl`, e.g. both `Foo/src/foo.jl` and `Bar/src/foo.jl`.
+Absolute paths only matches against the file with that exact absolute path.
+"""
+function breakpoint(file::String, line::Integer, condition::Condition=nothing)
+    file = normpath(file)
+    bp = BreakpointFileLocation(file, CodeTracking.maybe_fix_path(abspath(file)), line, condition, Ref(true), BreakpointRef[])
+    add_to_existing_framecodes(bp)
+    idx = findfirst(bp2 -> same_location(bp, bp2), _breakpoints)
+    idx === nothing ? push!(_breakpoints, bp) : (_breakpoints[idx] = bp)
+    return bp
+end
+
+function framecode_matches_breakpoint(framecode::FrameCode, bp::BreakpointFileLocation)
+    framecode.scope isa Method || return false
+    meth = framecode.scope
+    methpath = CodeTracking.maybe_fix_path(String(meth.file))
+    if bp.abspath == methpath || endswith(methpath, bp.path)
+        return method_contains_line(meth, bp.line)
+    else
+        return false
+    end
 end
 
 function shouldbreak(frame::Frame, pc::Int)
@@ -58,48 +136,66 @@ function prepare_slotfunction(framecode::FrameCode, body::Union{Symbol,Expr})
     return Expr(:function, Expr(:call, funcname, framename), Expr(:block, assignments..., body))
 end
 
-const Condition = Union{Nothing,Expr,Tuple{Module,Expr}}
 _unpack(condition) = isa(condition, Expr) ? (Main, condition) : condition
 
 ## The fundamental implementations of breakpoint-setting
-function breakpoint!(framecode::FrameCode, pc, condition::Condition=nothing)
+function breakpoint!(framecode::FrameCode, pc, condition::Condition=nothing, enabled=true)
     stmtidx = pc
     if condition === nothing
-        framecode.breakpoints[stmtidx] = BreakpointState()
+        framecode.breakpoints[stmtidx] = BreakpointState(enabled)
     else
         mod, cond = _unpack(condition)
         fex = prepare_slotfunction(framecode, cond)
-        framecode.breakpoints[stmtidx] = BreakpointState(true, Core.eval(mod, fex))
+        framecode.breakpoints[stmtidx] = BreakpointState(enabled, Core.eval(mod, fex))
     end
-    return add_breakpoint(framecode, stmtidx)
 end
 breakpoint!(frame::Frame, pc=frame.pc, condition::Condition=nothing) =
     breakpoint!(frame.framecode, pc, condition)
 
+update_states!(bp::AbstractBreakpoint) = foreach(bpref -> update_state!(bpref, bp.enabled[]), bp.applications)
+update_state!(bp::BreakpointRef, v::Bool) = bp[] = v
+
 """
-    enable(bp::BreakpointRef)
+    enable(bp::AbstractBreakpoint)
 
 Enable breakpoint `bp`.
 """
-enable(bp::BreakpointRef)  = bp[] = true
+enable(bp::AbstractBreakpoint) = (bp.enabled[] = true; update_states!(bp))
+enable(bp::BreakpointRef) = bp[] = true
+
 
 """
-    disable(bp::BreakpointRef)
+    disable(bp::AbstractBreakpoint)
 
 Disable breakpoint `bp`. Disabled breakpoints can be re-enabled with [`enable`](@ref).
 """
+disable(bp::AbstractBreakpoint) = (bp.enabled[] = false; update_states!(bp))
 disable(bp::BreakpointRef) = bp[] = false
 
 """
-    remove(bp::BreakpointRef)
+    remove(bp::AbstractBreakpoint)
 
 Remove (delete) breakpoint `bp`. Removed breakpoints cannot be re-enabled.
 """
-function remove(bp::BreakpointRef)
+function remove(bp::AbstractBreakpoint)
     idx = findfirst(isequal(bp), _breakpoints)
-    deleteat!(_breakpoints, idx)
+    idx === nothing || deleteat!(_breakpoints, idx)
+    foreach(remove, bp.applications)
+end
+function remove(bp::BreakpointRef)
     bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(false, falsecondition)
     return nothing
+end
+
+"""
+    toggle(bp::AbstractBreakpoint)
+
+Toggle breakpoint `bp`.
+"""
+toggle(bp::AbstractBreakpoint) = (bp.enabled[] = !bp.enabled[]; update_states!(bp))
+function toggle(bp::BreakpointRef)
+    state = bp[]
+    bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(!state.isactive, state.condition)
 end
 
 """
@@ -107,14 +203,14 @@ end
 
 Enable all breakpoints.
 """
-enable() = for bp in _breakpoints enable(bp) end
+enable() = foreach(enable, _breakpoints)
 
 """
     disable()
 
 Disable all breakpoints.
 """
-disable() = for bp in _breakpoints disable(bp) end
+disable() = foreach(disable, _breakpoints)
 
 """
     remove()
@@ -123,10 +219,9 @@ Remove all breakpoints.
 """
 function remove()
     for bp in _breakpoints
-        bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(false, falsecondition)
+        foreach(remove, bp.applications)
     end
     empty!(_breakpoints)
-    return nothing
 end
 
 """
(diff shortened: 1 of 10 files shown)


src/breakpoints.jl
```diff
@@ -1,25 +1,109 @@
...
-    idx = findfirst(isequal(bp), _breakpoints)
-    if idx === nothing
+function framecode_matches_breakpoint(framecode::FrameCode, bp::BreakpointSignature)
+    function extract_function_from_method(m::Method)
+        sig = m.sig
+        while sig isa UnionAll
+            sig = sig.body
+        end
+        f = sig.parameters[1]
+        if isdefined(f, :instance)
```
KristofferC: TODO: Check this properly.

src/breakpoints.jl
```diff
@@ -1,26 +1,108 @@
...
+function add_to_existing_framecodes(bp::AbstractBreakpoint)
+    for framecode in values(framedict)
+        add_breakpoint_if_match!(framecode, bp)
+    end
 end
-function toggle!(bp::BreakpointRef)
-    state = bp[]
-    bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(!state.isactive, state.condition)
+
+function add_breakpoint_if_match!(framecode::FrameCode, bp::AbstractBreakpoint)
```
KristofferC: Inconsitent use of `!`

src/breakpoints.jl
```diff
@@ -1,26 +1,108 @@
...
 end
 
+function framecode_matches_breakpoint(framecode::FrameCode, bp::BreakpointFileLocation)
+    framecode.scope isa Method || return false
+    meth = framecode.scope
+    meth.file == bp.file || return false
+    return method_contains_line(meth, bp.line)
+end
+
+Base.getindex(bp::BreakpointRef) = bp.framecode.breakpoints[bp.stmtidx]
```
KristofferC: This methods feels kinda out of place. Perhaps move to `types.jl`?

src/breakpoints.jl
```diff
@@ -1,26 +1,103 @@
...
+            return ft
+        end
+    end
+
+    framecode.scope isa Method || return false
+    meth = framecode.scope
+    bp.f isa Method && return meth === bp.f
+    bp.f === extract_function_from_method(meth) || return false
+    bp.sig === nothing && return true
+    tt = Base.signature_type(bp.f, bp.sig)
```
timholy: Would `bp.sig <: meth.sig` work here?

src/breakpoints.jl
```diff
@@ -1,26 +1,103 @@
...
+    end
 end
-function toggle!(bp::BreakpointRef)
-    state = bp[]
-    bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(!state.isactive, state.condition)
+
+function add_breakpoint_if_match!(framecode::FrameCode, bp::AbstractBreakpoint)
+    if framecode_matches_breakpoint(framecode, bp)
+        stmtidx = bp.line === 0 ? 1 : statementnumber(framecode, bp.line)
+        breakpoint!(framecode, stmtidx, bp.condition)
```
timholy: Do we need to check if there is already a breakpoint there? I worry that running `add_to_existing_framecodes` will reinstate `remove`d or `inactive` breakpoints.

src/types.jl
```diff
@@ -252,3 +269,48 @@ function Base.show(io::IO, bp::BreakpointRef)
...
+
+# Possible types for breakpoint condition
+const Condition = Union{Nothing,Expr,Tuple{Module,Expr}}
+
+abstract type AbstractBreakpoint end
+
+struct BreakpointSignature <: AbstractBreakpoint
+    f # Method or function
+    sig::Union{Nothing, Type}
+    line::Int # 0 is a sentinel for first statement
```
timholy: How about making this a `lineoffset` from the first statement? That may facilitate more robust identity after revision.

src/breakpoints.jl
```diff
@@ -1,26 +1,103 @@
...
+function add_to_existing_framecodes(bp::AbstractBreakpoint)
+    for framecode in values(framedict)
+        add_breakpoint_if_match!(framecode, bp)
+    end
 end
-function toggle!(bp::BreakpointRef)
-    state = bp[]
-    bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(!state.isactive, state.condition)
+
+function add_breakpoint_if_match!(framecode::FrameCode, bp::AbstractBreakpoint)
```
timholy: Should we do this only for `BreakpointSignature`s? My biggest concern regards my last comment in https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/256#issuecomment-478649540: what to do about cases where users specified a breakpoint by file/line and then revised the file, thus changing the line numbers. Perhaps a more robust approach would be to convert BreakpointFileLocations to BreakpointSignatures as soon as they resolve?

src/breakpoints.jl
KristofferC: For full generality, each statement should be able to have arbitrary number of independent breakpoints which is doable but maybe not worth the effort.

I don't really see why overwriting that `BreakpointState` is worse than not doing anything.

src/breakpoints.jl
KristofferC: Hm, perhaps a `BreakpointFileLocation` should only be able to bind once to a method and if the method is reevaluated you need to re-add the breakpoint.  It would be similar to the current situation except that you can now do "late binding", i.e. add a breakpoint to a file:line before it is included.

src/breakpoints.jl
timholy: To be clear I'm simply asking what happens under circumstances like the following:
- `breakpoint(f)`
- disable that breakpoint
- define a new method for `f`
- debug either the new method or one of the old ones.

IIUC (which I may not), I am concerned that the breakpoint will be in `_breakpoints` without any record of the fact that you disabled it, so it will get added in active form. I am guessing this might be a problem for the new method but not the old ones? In which case my "check if there is already a breakpoint there" is a red herring---what I am really after is whether we're recording state everywhere we need it. Seems like you might need an `active` field in `BreakpointSignature` and `BreakpointFileLocations`?

KristofferC: > Seems like you might need an active field in BreakpointSignature and BreakpointFileLocations?

Yes, this sounds correct, gotcha. Will make sure doing those steps does something sensible.

src/breakpoints.jl
KristofferC: :+1:

src/breakpoints.jl
KristofferC: Ok, should be fixed, with a test added.

src/breakpoints.jl
KristofferC: But the same thing sort of applies to a breakpoint to a method or signature with a line number. Should those also only trigger once?

timholy: Yes, I'm fine with triggering only once. In Juno one might imagine seeing the breakpoint vanish when you save the file, so it should be very easy for users to compensate.

KristofferC: Hm, how would Juno know the breakpoint is invalid? The breakpoint instance will still be there in `_breakpoints`, it is just that when we interpret the method, another `FrameCode` will be executed.

FWIW, I think Juno should implement location based breakpoints on a level higher than here. In IDEs the breakpoint typically to follow along with edits, so if you insert a line above a breakpoint, the breakpoint will move down a line (together with the rest of the text). Clearly, we cannot `breakpoint(file, line)` and `remove` every time a user enter a line, so we need to "apply" the current set of line breakpoints when the debugger is started, at the locations they currently are.

cc @pfitzseb

timholy: Revise could fire a callback notifying Juno of deleted methods. Though I guess `revise()` will run when the user tries the next command, not when the user saves the file. So the appropriate spot may no longer be visible...

KristofferC: It's also not only Revise that can redefine methods. Ctrl + enter in Juno or just re-including a file for example. Losing breakpoints because you reinclude a file would perhaps be annoying.

src/types.jl
KristofferC: Could you give an example? If Revise has revised it, we create a new `FrameCode` which will have updated line numbers.

src/breakpoints.jl
```diff
@@ -1,25 +1,99 @@
...
+    add_to_existing_framecodes(bp)
+    idx = findfirst(bp2 -> same_location(bp, bp2), _breakpoints)
+    idx === nothing ? push!(_breakpoints, bp) : (_breakpoints[idx] = bp)
+    return bp
+end
+
+function framecode_matches_breakpoint(framecode::FrameCode, bp::BreakpointFileLocation)
+    framecode.scope isa Method || return false
+    meth = framecode.scope
+    meth.file == bp.file || return false
```
KristofferC: I wonder if we should be more relaxed here. For example, you might just want to bother writing a full path to it. Perhaps `endswith`? Then you can add a breakpoint to `foo.jl` if it is in `Package/src/foo.jl`?

KristofferC:
Happy for more comments here. @pfitzseb would be good to know how you feel about this with regards to Juno integration.

src/types.jl
timholy: This is conditioned on what I thought was your goal to try to preserve breakpoints through a revision. (You may be abandoning that goal, IIUC.) But suppose you wanted to preserve "break at the second line of the body" when you revise (1) a single `+` to a `-` somewhere in the body of your method, and (2) also make other changes elsewhere in the same source file that alter the absolute line numbers. Then if `BreakpointSignature` stored a line offset rather than an absolute line number, it would be unaffected by the fact that unrelated changes occurred elsewhere in the same file. Note that the user could specify the line numbers in absolute terms, I'm just proposing that when one creates the `BreakpointSignature` one should convert it to an offset.

src/breakpoints.jl
timholy: That is an attractive thought, it might also eliminate path weirdness (symlinks etc). OTOH I have been bit sometimes by Julia's `edit` if I happen to be in a directory that has a file of the same name as one of Julia's own base files. If lots of packages have the same file (e.g., `precompile.jl`) will this cause trouble?

KristofferC: Maybe, but you always have the option to specify the full path if you find it matches places where you don't want it to.

pfitzseb:
Ok, so greedy users (e.g. me) want at least the following:
1. Instant (and correct) feedback when setting a breakpoint.
2. If I have set a breakpoint on a certain line then I'd like to keep it at that line (meaning the actual code, not the line number) no matter what happens.
3. Have that breakpoint work no matter how often the method containing it is redefined/has it's name changed/whatever.

AFAICT, 1) works well with the current implementation of breakpoints, 2) kinda works on the Julia side but not the GUI side right now (because the breakpoints line number isn't updated, but maybe Revise had a hiccup). 3) fails.

The new implementation makes 1) basically impossible because we can only know if a breakpoint works once we hit it, right? 2) and 3) would work with this PR as long as a prospective breakpoint can bind to any number of methods.

I'll think more about this tomorrow, but for now I don't see a good way to have all three of the options above at the same time.

KristofferC:
> Instant (and correct) feedback when setting a breakpoint.

What kind of feedback? Breakpoints are set through the Juno interface right (I think it almost has to be) so you would have all information there.
You shouldn't even need to have JuliaInterpreter loaded to set breakpoints before starting execution.

> If I have set a breakpoint on a certain line then I'd like to keep it at that line (meaning the actual code, not the line number) no matter what happens.

Yes, you would have to move the breakpoint with editing of the code which is a front end problem.

bp

When the debugger starts, you need to set the breakpoints at the line they currently are.
When the debugger exits, you should probably clear all location breakpoints. 

> Have that breakpoint work no matter how often the method containing it is redefined/has it's name changed/whatever.

Yes, that should work as long as you reapply the breakpoints when you enter the debugger.

pfitzseb:
> What kind of feedback? Breakpoints are set through the Juno interface right (I think it almost has to be) so you would have all information there.
You shouldn't even need to have JuliaInterpreter loaded to set breakpoints before starting execution.

Hm. Can we guarantee that we can add a valid breakpoint wherever a user might click? What happens on empty lines/comments/outside of functions?

> Yes, you would have to move the breakpoint with editing of the code which is a front end problem.

Sure, the GUI side of this is (almost) trivial.

> When the debugger starts, you need to set the breakpoints at the line they currently are.
When the debugger exits, you should probably clear all location breakpoints.

Right, makes sense.

KristofferC:
> Hm. Can we guarantee that we can add a valid breakpoint wherever a user might click? What happens on empty lines/comments/outside of functions?

Lines outside of functions will never be reached since they will never be "activated".

For comments and empty lines, I have to look exactly into the current implementation but my guess is that it takes the next statement after the line.

KristofferC:
Pushed some fixes and changes to path handling. So you can set a breakpoint by just giving e.g. the file name. For absolute paths (e.g. from an IDE) this shouldn't change anything.

KristofferC:
Also pushed a commit to clear caches, showing that the breakpoint system still works and fixes #262

src/types.jl
KristofferC: I don't know. Currently I feel like just keeping the line number is the most straight forward. FWIW, I don't see breakpoints on signatures at a given line to be used that much. Just adding file:line is likely more ergonomic.

src/types.jl
```diff
@@ -252,3 +269,59 @@ function Base.show(io::IO, bp::BreakpointRef)
...
+
+same_location(::AbstractBreakpoint, ::AbstractBreakpoint) = false
+
+struct BreakpointSignature <: AbstractBreakpoint
+    f # Method or function
+    sig::Union{Nothing, Type}
+    line::Int # 0 is a sentinel for first statement
+    condition::Condition
+    enabled::Ref{Bool}
+    applications::Vector{BreakpointRef}
```
timholy: Would `instances` be a better name for this field? Whatever you choose, obviously use the same name for `BreakpointFileLocation`.

src/types.jl
```diff
@@ -252,3 +269,59 @@ function Base.show(io::IO, bp::BreakpointRef)
...
+    end
+    if !bp.enabled[]
+        print(io, " [disabled]")
+    end
+end
+
+struct BreakpointFileLocation <: AbstractBreakpoint
+    path::String
+    abspath::String
+    line::Integer
```
timholy:
```suggestion
    line::Int
```

src/breakpoints.jl
```diff
@@ -1,26 +1,104 @@
...
+    breakpoint(file, line, [condition])
+
+Set a breakpoint in `file` at `line`. The argument `file` can be a filename, a partial path or absolute path.
+For example, `file = foo.jl` will match against all files with the name `foo.jl`,
+`file = src/foo.jl` will match against all paths containing `src/foo.jl`, e.g. both `Foo/src/foo.jl` and `Bar/src/foo.jl`.
+Absolute paths only matches against the file with that exact absolute path.
+"""
+function breakpoint(file::String, line::Integer, condition::Condition=nothing)
+    file = normpath(file)
+    bp = BreakpointFileLocation(file, CodeTracking.maybe_fix_path(abspath(file)), line, condition, Ref(true), BreakpointRef[])
```
timholy: Perhaps it's deliberate, but this won't work for non-absolute Base paths, e.g., `"numbers.jl"`.

timholy:
To me it seems the main outstanding issues are (1) implementing pieces that need to go in Juno and (2) deciding exactly how we want these to work when methods or files are revised. For (2), I'd propose we add a new test file, `revise.jl`, for which Revise is a test dependency and in which we simulate various changes that might come from people using an editor other than Juno/Atom. We can then use these tests to decide what policy we want to adopt. If you agree with this proposal, I can probably help write these tests. (Many models can be found in Revise's `runtests.jl` file, of course.)

src/breakpoints.jl
KristofferC: What won't work? If we give it as `numbers.jl` it will be stored as `"numbers.jl"` in the `path` field and we will match against all paths that ends with `number.jl`.

The only reason for the `maybe_fix_path` is if someone gives the "buildbot path" to a method by e.g. using `(@which uuid1()).file` (which the tests does).

timholy: Good point. Perhaps I don't understand the motivation for both `path` and `abspath` in that struct.

KristofferC: Yeah, I can understand that is confusing. The reason is if someone adds a path like `"../foo.jl"`.  If we only store `"../foo.jl"` and the user moves directory we will start referring to another file which is weird. However, we can't store the absolute path only because then we cant do the "trailing path matching".

KristofferC: I should add a test for this though.

timholy: And a code comment.

KristofferC:
> deciding exactly how we want these to work when methods or files are revised.

I've been thinking about this during the last couple of days and all I feel is:

tenor

There is no good choice when all you get is the snapshot of a new method. From an IDE it is easy since you just let the breakpoints flow with the editing of the text. I don't feel strongly about what should happen after Revise as long as I can set a breakpoint in a arbitrary file without having to track it first.

src/types.jl
KristofferC: Sounds good.

KristofferC:
```
julia> tempdir()
"/var/folders/64/76tk_g152sg6c6t0b4nkn1vw0000gn/T"

julia> cd(tempdir())

julia> pwd()
"/private/var/folders/64/76tk_g152sg6c6t0b4nkn1vw0000gn/T"
```

KristofferC:
PR updated based on last comments.

KristofferC:
@pfitzseb, It would be interesting to see how Juno works on top of this PR. So, whenever you have the time, it would be valuable to try it out and see how things work.

pfitzseb:
I'll try this as soon as I have some time (hopefully tomorrow). Last days have been very busy for me, sorry.

KristofferC:
So, anything left to do here for me?

I kind of want to get this in because I think it provides a better breakpoint experience than the current. and I want to add a UI for breakpoints in Debugger.jl but I don't want to put in the time to do that unless this is likely to be mergeable in a reasonable timeframe.

pfitzseb:
Uh, sorry -- I was sure I left a comment here...

So yeah, I have a Atom.jl branch that uses this PR and it feels pretty good to use. Imho this is good to go as is.
We could add some convenience methods for e.g. getting the location of a breakpoint and stuff like that (basically just accessor functions for the various kinds of breakpoints this introduces), but I don't think that should block this PR.

KristofferC:
@timholy what type of interface do you want to use for Rebugger? Is it ok just to have breakpoint in Rebugger just set a breakpoint at file:line or do you need more fine grained control (e.g. statement index level control on specific methods).

timholy:
File/line is fine. Thanks!

KristofferC:
Added docs for the new types.

src/breakpoints.jl
```diff
@@ -1,26 +1,113 @@
...
-        push!(_breakpoints, bp)
+function framecode_matches_breakpoint(framecode::FrameCode, bp::BreakpointSignature)
+    function extract_function_from_method(m::Method)
+        sig = Base.unwrap_unionall(m.sig)
+        ft0 = sig.parameters[1]
+        ft = Base.unwrap_unionall(ft0)
+        if ft <: Function && isa(ft, DataType) && isdefined(ft, :instance)
+            return ft.instance
+        elseif isa(ft, DataType) && ft.name === Type.body.name
+            f = ft.parameters[1]
```
timholy:
```suggestion
            return ft.parameters[1]
```

KristofferC:
@pfitzseb, do you have a branch with the Atom update to use this PR? If so, I could try it out (would probably need installation instructions though).

pfitzseb:
I have something that's compatible with these changes, but it's not quite where I want it to be. Will ping you once I have PRs up though.

PR #263, opened 2019-04-03
RFC: simplify next_line! implementation

KristofferC:
I must say I don't really understand the complications of the current implementation. So I removed most of it. Happy to close this if there are good examples why the complications are useful.


timholy:
AFAICT the complexity was inherited. I don't see what's wrong with this.

KristofferC:
> AFAICT the complexity was inherited

Yeah, that's what I thought. Let's try with this and see if someone comes screaming.

Issue #175, opened 2019-03-18
Step out of loop.


fredrikekre:
Would be convenient to be able to step to the end of a for/while loop and break there.

KristofferC:
Debugger with text editors does this via "Run to cursor".

Gdb has `until`


>Continue running until a source line past the current line, in the current stack frame, is reached. This command is used to avoid single stepping through a loop more than once. It is like the next command, except that when until encounters a jump, it automatically continues execution until the program counter is greater than the address of the jump.
>This means that when you reach the end of a loop after single stepping though it, until makes your program continue execution until it exits the loop. In contrast, a next command at the end of a loop simply steps back to the beginning of the loop, which forces you to step through the next iteration.

Should `n` has the same semantics as `until`?

vchuravy:
Since we have better access to the source-code we maybe could have CFG related commands, like next-block, until, ...

From a IR perspective loops are just backwards branches, so until is neat but requires you to use it in place of next.

PR #266, opened 2019-04-03
add `until_line!` to step to next line lower down in source code

KristofferC:
This is like `next_line!` except it guarantees you to move to something "later" in the source code. For example:

```jl
julia> function f(x)
           s = 0
           for i in 1:length(x)
               s = s + 1
           end
           print(s)
           return s
       end

julia> frame = JuliaInterpreter.enter_call(f, rand(100))
Frame for f(x) in Main at REPL[23]:2
   1* 2  1 ─       s = 0
   2  3  │   %2  = (length)(x)
   3  3  │   %3  = (Colon())(1, %2)
⋮
x = [0.770521, 0.714896, 0.271307, 0.498012, 0.623033, 0.17695, 0.594108, 0.0127458, 0.287482, 0.396186  …  0.632738, 0.0830569, 0.586225, 0.287229, 0.0564309, 0.489518, 0.660743, 0.0207911, 0.908642, 0.925855]

julia>  JuliaInterpreter.until!(frame); frame
Frame for f(x) in Main at REPL[3]:2
   1  2  1 ─       s = 0
   2* 3  │   %2  = (length)(x)
   3  3  │   %3  = (Colon())(1, %2)
   4  3  │         #temp# = (iterate)(%3)
⋮
x = [0.953892, 0.828438, 0.327365, 0.47125, 0.266897, 0.727434, 0.89857, 0.152085, 0.269296, 0.485211  …  0.270606, 0.790149, 0.309138, 0.127694, 0.0475989, 0.674949, 0.762637, 0.769296, 0.407537, 0.476802]
s = 0

julia> JuliaInterpreter.until!(JuliaInterpreter.finish_and_return!, frame); frame
linenumber(frame, pc) <= line = true
Frame for f(x) in Main at REPL[23]:2
⋮
   9  3  │         i = (getfield)(%8, 1)
  10  3  │   %10 = (getfield)(%8, 2)
  11* 4  │         s = (+)(s, 1)
  12  4  │         #temp# = (iterate)(%3, %10)
  13  4  │   %13 = (===)(#temp#, nothing)
⋮
x = [0.770521, 0.714896, 0.271307, 0.498012, 0.623033, 0.17695, 0.594108, 0.0127458, 0.287482, 0.396186  …  0.632738, 0.0830569, 0.586225, 0.287229, 0.0564309, 0.489518, 0.660743, 0.0207911, 0.908642, 0.925855]
s = 0
#temp# = (1, 1)
i = 1

julia>  JuliaInterpreter.until!(frame); frame
Frame for f(x) in Main at REPL[3]:2
⋮
  15  4  └──       goto #4 if not %14
  16  4  3 ─       goto #2
  17* 6  4 ┄       (print)(s)
  18  7  └──       return s
x = [0.953892, 0.828438, 0.327365, 0.47125, 0.266897, 0.727434, 0.89857, 0.152085, 0.269296, 0.485211  …  0.270606, 0.790149, 0.309138, 0.127694, 0.0475989, 0.674949, 0.762637, 0.769296, 0.407537, 0.476802]
s = 100
#temp# = nothing
i = 100

```

Should hook this up to `debug_command` and add tests etc but wanted to float the idea first.

Ref https://github.com/JuliaDebug/Debugger.jl/issues/138

Fixes  https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/175

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -181,6 +181,25 @@ end
 next_line!(frame::Frame, istoplevel::Bool=false) = next_line!(finish_and_return!, frame, istoplevel)
 
 """
+    pc = until_line!(recurse, frame, line=nothing istoplevel=false)
+    pc = until_line!(frame, line=nothing, istoplevel=false)
+
+Execute until the current frame reaches a line greater than `line`. If `line == nothing`
+execute until the current frame reaches any line greater than the current line.
+"""
+function until_line!(@nospecialize(recurse), frame::Frame, line::Union{Nothing, Integer}=nothing, istoplevel::Bool=false)
+    pc = frame.pc
+    initialline, initialfile = linenumber(frame, pc), getfile(frame, pc)
+    line === nothing && (line = initialline + 1)
+    predicate(frame) = isexpr(pc_expr(frame), :return) || (linenumber(frame) >= line && getfile(frame) == initialfile)
+    pc = next_until!(predicate, frame, istoplevel)
+    (pc === nothing || isa(pc, BreakpointRef)) && return pc
+    maybe_step_through_kwprep!(recurse, frame, istoplevel)
+    maybe_next_call!(recurse, frame, istoplevel)
+end
+until_line!(frame::Frame, line::Union{Nothing, Integer}=nothing, istoplevel::Bool=false) = until_line!(finish_and_return!, frame, line, istoplevel)
+
+"""
     cframe = maybe_step_through_wrapper!(recurse, frame)
     cframe = maybe_step_through_wrapper!(frame)
 
@@ -343,13 +362,15 @@ function unwind_exception(frame::Frame, exc)
 end
 
 """
-    ret = debug_command(recurse, frame, cmd, rootistoplevel=false)
-    ret = debug_command(frame, cmd, rootistoplevel=false)
+    ret = debug_command(recurse, frame, cmd, rootistoplevel=false; line=nothing)
+    ret = debug_command(frame, cmd, rootistoplevel=false; line=nothing)
 
-Perform one "debugger" command. `cmd` should be one of:
+Perform one "debugger" command. The keyword arguments are not used for all debug commands.
+`cmd` should be one of:
 
 - `:n`: advance to the next line
 - `:s`: step into the next call
+- `:until`: advance the frame to line `line` if given, otherwise advance to the line after the current line
 - `:c`: continue execution until termination or reaching a breakpoint
 - `:finish`: finish the current frame and return to the parent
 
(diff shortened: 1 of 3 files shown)


timholy:
How about calling this `until_line!`? `until!` sounds approximately as general as `next_until!` but it doesn't accept a general predicate.

Another implementation option would be to generalize the predicate in `next_until!` to accept a `predicate(frame, stmt)` in which case this becomes a special case. Obviously that's a breaking change, and this is not.

KristofferC:
> Another implementation option would be to generalize the predicate in next_until! to accept a predicate(frame, stmt)

Yeah, I thought about this too. I think it is a good idea.

KristofferC:
Updated.

src/commands.jl
```diff
@@ -179,6 +181,25 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
 end
 next_line!(frame::Frame, istoplevel::Bool=false) = next_line!(finish_and_return!, frame, istoplevel)
 
+"""
+    pc = until!(recurse, frame, line=nothing istoplevel=false)
```
timholy: `until_line!`, here and below (GitHub won't let me make suggestions when I'm reviewing only one of the 2 commits here)

src/commands.jl
```diff
@@ -179,6 +181,25 @@ function next_line!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false
...
+"""
+function until_line!(@nospecialize(recurse), frame::Frame, line::Union{Nothing, Integer}=nothing, istoplevel::Bool=false)
+    pc = frame.pc
+    initialline, initialfile = linenumber(frame, pc), getfile(frame, pc)
+    line === nothing && (line = initialline + 1)
+    predicate(frame) = isexpr(pc_expr(frame), :return) || (linenumber(frame) >= line && getfile(frame) == initialfile)
+    pc = next_until!(predicate, frame, istoplevel)
+    (pc === nothing || isa(pc, BreakpointRef)) && return pc
+    maybe_step_through_kwprep!(recurse, frame, istoplevel)
+    maybe_next_call!(recurse, frame, istoplevel)
```
timholy: Is there a risk this will bypass the chosen line? What if the chosen line has contents `x = 5`?

KristofferC: Yes, it will, but that is not different from `next_line!`?

timholy: OK, so it's a bug there too :smile:

KristofferC: I guess the reason the code has originally liked to end up at calls is due to the many garbage assignments that happen, e.g. `t@_6 = y` etc.

If you prefer, I can remove the `maybe_next_call!` from here and `next_line!` and just remove the conditional https://github.com/JuliaDebug/Debugger.jl/blob/909afc889f7ed19679ab937cac64de6ed2b6a1e8/src/commands.jl#L48.

timholy: What if we passed a predicate to `maybe_next_call!` looking for another increment in line number, and terminated advancement if it is satisfied?

That could be a separate PR and you could merge this as-is.

KristofferC: So something like `advance_to_next_call_if_it_is_on_same_line_as_current_line` :P ? Perhaps that is a good compromise.

KristofferC: Anyway, let's deal with the issues of exactly where we should stop in another issue / PR.

PR #269, opened 2019-04-04
generalize next_until! to take a frame and add maybe_next_until!

KristofferC:
This is breaking since it is changing the argument type of the `predicate`.

The `maybe_next_until!` was added because it felt weird to have a `maybe` version of the less general `next_call!` but not one for `next_until!`.

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -86,13 +86,13 @@ finish_stack!(frame::Frame, istoplevel::Bool=false) = finish_stack!(finish_and_r
     pc = next_until!(predicate, frame, istoplevel=false)
 
 Execute the current statement. Then step through statements of `frame` until the next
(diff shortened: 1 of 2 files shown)


src/commands.jl
```diff
@@ -86,13 +86,13 @@ finish_stack!(frame::Frame, istoplevel::Bool=false) = finish_stack!(finish_and_r
     pc = next_until!(predicate, frame, istoplevel=false)
 
 Execute the current statement. Then step through statements of `frame` until the next
-statement satifies `predicate(stmt)`. `pc` will be the index of the statement at which
+statement satifies `predicate(frame)`. `pc` will be the index of the statement at which
```
```suggestion
statement satisfies `predicate(frame)`. `pc` will be the index of the statement at which
```

(unrelated to your change, but happened to notice)

PR #270, opened 2019-04-05
fix next_until! not properly returning breakpoint refs

KristofferC:
Noticed from a test failure in Debugger.jl


src/commands.jl
```diff
@@ -92,9 +92,8 @@ evaluation terminates, `nothing` (if the frame reached a `return`), or a `Breakp
 function next_until!(@nospecialize(predicate), @nospecialize(recurse), frame::Frame, istoplevel::Bool=false)
     pc = step_expr!(recurse, frame, istoplevel)
     while pc !== nothing && !isa(pc, BreakpointRef)
-        if predicate(frame) || shouldbreak(frame, pc)
```
KristofferC: It doesn't feel like we should need a `shouldbreak` at all here.

PR #275, opened 2019-04-10
check in the builtins file

### LINKED ISSUE #272: What should happen if the user does not have write permissions to the `src/` directory?
We have a workstation where we install Julia system-wide and install packages to the system-wide depot.

I run `import JuliaInterpreter` once with root permissions. This ensures that the `src/builtins-julia1.1.jl` file is generated.

However, if a non-root user then attempts to run `import JuliaInterpreter`, they get a permissions error because JuliaInterpreter tries to open the `src/builtins-julia1.1.jl` file with `w` permissions. This causes precompilation to fail for JuliaInterpreter and thus for every package that depends on JuliaInterpreter.

Could we have a fallback where if the user does not have write permissions to `src/builtins-julia1.1.jl``, but the file already exists, we don't throw an error?

KristofferC:
Fixes #272, closes https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/271, closes https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/273.

The dynamic generation of the builtin time at precompilation time is nice but it has some drawbacks (see linked issues). Other drawbacks are accidentally overwriting it while having inserted some debugging code.
I propose just committing the generated file with some `@static` checks for those builtins that have changed between versions. Yeah, the package will be broken for a while when a new builtin is added but I don't think it will happen too often and we'll fix it quickly. Not a bigger deal than any other breakage that happens due to changes in the internals on master.


timholy:
I think this is pretty reasonable. The main thing I wonder about is whether it's better to use `@static` or just have separate files for different Julia versions. For now at least, I agree that this version is better. So, I approve!

PR #277, opened 2019-04-16
Interpret most of Core.Compiler

### LINKED ISSUE #242: Don't blanket ban `Core.Compiler` from interpretation
In https://github.com/JuliaDebug/Debugger.jl/issues/150 I wanted to use Debugger.jl to step through a method in `Core.Compiler`. This is currently blocked by https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/6b93e041614b697ed0c05dc68075b863849095c2/src/construct.jl#L119-L121 

It would be awesome if I could use the Debugger for Core.Compiler, but I am okay with having this be a special case (similar to `Revise.track(Core.Compiler)`).

### LINKED ISSUE #257: The interpreter does not handle SSAValues that occur in code correctly

timholy:
This opens most of `Core.Compiler` to the interpreter. The unfortunate exceptions are some of the most interesting functions, the various `typeinf_*` functions: so far I have been unable to avoid segfaults from anything that starts a `ccall(:jl_typeinf_begin, ...)` block. However, hopefully this is still a considerable improvement from where we are currently. (Thoughts, @vchuravy?)

There are three substantive changes:
- Avoid using `Core.SlotNumber` and `Core.SSAValue`. Instead, we define equivalent types here and replace the values that come by from `Base.uncompressed_ast` with the local equivalents. This allows us to distinguish an `SSAValue` generated as a value from `Core.Compiler` methods vs one that's being used by the interpreter. Disadvantage: I suspect frame printing is broken. That's one reason this is a WIP.
- `ccall`s made from `Core.Compiler` have to have their compiled variants evaluated in `Core.Compiler` rather than `JuliaInterpreter.CompiledCalls`. The reason is that `Core.Compiler.unsafe_convert != Base.unsafe_convert` and `Core.Compiler.cconvert != Base.cconvert` and these end up getting used by Julia internally when implementing our compiled variants.
- In a (failed) attempt to avoid those segfaults, I set up a mechanism to re-used previous compiled `ccall` functions. As a consequence one no longer needs one per call site, it's just one per type signature. This seems useful on its own, even if it didn't fix the core problem I was trying to solve.

Fixes #257. Fixes #242. Closes #258.

### DIFF
--- a/src/JuliaInterpreter.jl
+++ b/src/JuliaInterpreter.jl
@@ -70,11 +70,43 @@ function set_compiled_methods()
     push!(compiled_methods, which(subtypes, Tuple{Module, Type}))
     push!(compiled_methods, which(subtypes, Tuple{Type}))
 
+    # Anything that ccalls jl_typeinf_begin cannot currently be handled
+    for finf in (Core.Compiler.typeinf_code, Core.Compiler.typeinf_ext, Core.Compiler.typeinf_type)
+        for m in methods(finf)
+            push!(compiled_methods, m)
+        end
+    end
+
     push!(compiled_modules, Base.Threads)
 end
 
 function __init__()
     set_compiled_methods()
+    # If we interpret into Core.Compiler, we need to take precautions to avoid needing
+    # inference of JuliaInterpreter methods in the middle of a `ccall(:jl_typeinf_begin, ...)`
+    # block.
+    # for (sym, RT, AT) in ((:jl_typeinf_begin, Cvoid, ()),
+    #                       (:jl_typeinf_end, Cvoid, ()),
+    #                       (:jl_isa_compileable_sig, Int32, (Any, Any)),
+    #                       (:jl_compress_ast, Any, (Any, Any)),
+    #                       # (:jl_set_method_inferred, Ref{Core.CodeInstance}, (Any, Any, Any, Any, Int32, UInt, UInt)),
+    #                       (:jl_method_instance_add_backedge, Cvoid, (Any, Any)),
+    #                       (:jl_method_table_add_backedge, Cvoid, (Any, Any, Any)),
+    #                       (:jl_new_code_info_uninit, Ref{CodeInfo}, ()),
+    #                       (:jl_uncompress_argnames, Vector{Symbol}, (Any,)),
+    #                       (:jl_get_tls_world_age, UInt, ()),
+    #                       (:jl_call_in_typeinf_world, Any, (Ptr{Ptr{Cvoid}}, Cint)),
+    #                       (:jl_value_ptr, Any, (Ptr{Cvoid},)),
+    #                       (:jl_value_ptr, Ptr{Cvoid}, (Any,)))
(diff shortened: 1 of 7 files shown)


KristofferC:
Pushed commits with test for SSAValue handling and fixup for the Frame printing.

src/JuliaInterpreter.jl
```diff
@@ -70,12 +70,43 @@ function set_compiled_methods()
...
+    #                       (:jl_value_ptr, Ptr{Cvoid}, (Any,)))
+    #     fname = Symbol(:ccall_, sym)
+    #     qsym = QuoteNode(sym)
+    #     argnames = [Symbol(:arg_, string(i)) for i = 1:length(AT)]
+    #     TAT = Expr(:tuple, [parametric_type_to_expr(t) for t in AT]...)
+    #     def = :($fname($(argnames...)) = ccall($qsym, $RT, $TAT, $(argnames...)))
+    #     f = Core.eval(Core.Compiler, def)
+    #     compiled_calls[(qsym, RT, Core.svec(AT...), Core.Compiler)] = f
+    #     precompile(f, AT)
+    # end
```
timholy: Should I strip this block out or leave it commented out? It's possible this is on the right track towards getting `typeinf_ext` and friends working (and simply not yet sufficient on its own), but it's also possible this is irrelevant.

KristofferC: Not sure, doesn't matter to me so do as you wish.

vchuravy:
Whoo! That is fabulous thank you guys!

PR #283, opened 2019-05-12
Support self-referential struct definitions

timholy:
Fixes https://github.com/timholy/Revise.jl/issues/304


timholy:
Seem OK?

timholy:
What needs doing before we can tag a new release? Nightly support?

PR #284, opened 2019-05-13
Update version to 0.6.0

KristofferC:
- Refactoring of breakpoints to allow breakpoints in lines to be set without using Revise (#259)
- Support self-referential struct definitions (#283)


KristofferC:
We could also make a release 0.5.2 release with just #283 first if you want @timholy

timholy:
If other tools are likely to take a while to catch up to the new breakpoints, that might not be a bad idea.

pfitzseb:
Let's merge this? Both Debugger.jl (master) and Juno (master) are compatible with this, right?

PR #289, opened 2019-05-24
Compile parametric methods that contain llvmcall

### LINKED ISSUE #112: MWE (sort of) for vecelement
The failure in vecelement is in the last test (figures).  It results in an "this intrinsic must be compiled to be called" error.  I _think_ the problem is that a llvmcall is buried inside of a generated function, which defeats the compiled methods recognition?

```
using JuliaInterpreter
include("utils.jl")

module VecTest

Vec{N,T} = NTuple{N,Base.VecElement{T}}

# The following test mimic SIMD.jl
const _llvmtypes = Dict{DataType, String}(
    Float64 => "double",
    Float32 => "float",
    Int32 => "i32",
    Int64 => "i64"
)

@generated function vecadd(x::Vec{N, T}, y::Vec{N, T}) where {N, T}
    llvmT = _llvmtypes[T]
    func = T <: AbstractFloat ? "fadd" : "add"
    exp = """
    %3 = $(func) <$(N) x $(llvmT)> %0, %1
    ret <$(N) x $(llvmT)> %3
    """
    return quote
        Base.@_inline_meta
        Core.getfield(Base, :llvmcall)($exp, Vec{$N, $T}, Tuple{Vec{$N, $T}, Vec{$N, $T}}, x, y)
    end
end

end
m = VecTest


ex = quote
    a = (VecElement{Float64}(1.0), VecElement{Float64}(2.0))
    b = vecadd(a, a)
end


nstmts = 1000000

modexs, _ = JuliaInterpreter.split_expressions(m, ex)
stack = JuliaStackFrame[]
for modex in modexs
    frame = JuliaInterpreter.prepare_thunk(modex)
    nstmtsleft = nstmts
    while true
        ret, nstmtsleft = evaluate_limited!(stack, frame, nstmtsleft)
        break
        if isa(ret, Aborted)
            run_compiled(frame)
            break
        elseif isa(ret, Some)
            break
        end
    end
end
```

### LINKED ISSUE #288: expected return statement, got JuliaInterpreter.SSAValue(2)
With the following (minified) package
```
module ReviseBug

using Tensors

struct A
    x::Tensor{2,3,Float64,9}
end

function foo()
    x = 1.0 * one(Tensor{2,3,Float64,9}) # removing the multiplication makes things work
    return A(x)
end

const a = foo()

end # module
```
I get the following when trying to revise the module:
```
julia> using Revise, ReviseBug

shell> touch src/ReviseBug.jl

julia> 1+1
┌ Warning: omitting call expression (ReviseBug.foo)() in ("none", 0)
└ @ Revise ~/.julia/packages/Revise/UtBAC/src/lowered.jl:169
┌ Error: evaluation error
│   mod = ReviseBug
│   ex = :(const a = foo())
│   exception =
│    expected return statement, got a = JuliaInterpreter.SSAValue(2)
│    Stacktrace:
│     [1] error(::String, ::Expr) at ./error.jl:42
│     [2] get_return(::JuliaInterpreter.Frame) at /home/fredrik/.julia/packages/JuliaInterpreter/6LimI/src/interpret.jl:598
│     [3] #methods_by_execution!#14(::Bool, ::Bool, ::Function, ::Any, ::Revise.CodeTrackingMethodInfo, ::Dict{Module,Array{Expr,1}}, ::JuliaInterpreter.Frame) at /home/fredrik/.julia/packages/Revise/UtBAC/src/lowered.jl:185
│     [4] #methods_by_execution! at ./none:0 [inlined]
│     [5] #methods_by_execution!#9(::Base.Iterators.Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:define,),Tuple{Bool}}}, ::Function, ::Any, ::Revise.CodeTrackingMethodInfo, ::Dict{Module,Array{Expr,1}}, ::Module, ::Expr) at /home/fredrik/.julia/packages/Revise/UtBAC/src/lowered.jl:49
│     [6] #methods_by_execution! at ./none:0 [inlined]
│     [7] #eval_with_signatures#59(::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Module, ::Expr) at /home/fredrik/.julia/packages/Revise/UtBAC/src/Revise.jl:344
│     [8] #eval_with_signatures at /home/fredrik/.julia/packages/Revise/UtBAC/src/Revise.jl:0 [inlined]
│     [9] #instantiate_sigs!#60(::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::OrderedCollections.OrderedDict{Module,OrderedCollections.OrderedDict{Revise.RelocatableExpr,Union{Nothing, Array{Any,1}}}}) at /home/fredrik/.julia/packages/Revise/UtBAC/src/Revise.jl:352
│     [10] instantiate_sigs! at /home/fredrik/.julia/packages/Revise/UtBAC/src/Revise.jl:349 [inlined]
│     [11] maybe_parse_from_cache!(::Revise.PkgData, ::String) at /home/fredrik/.julia/packages/Revise/UtBAC/src/pkgs.jl:224
│     [12] handle_deletions at /home/fredrik/.julia/packages/Revise/UtBAC/src/Revise.jl:473 [inlined]
│     [13] revise() at /home/fredrik/.julia/packages/Revise/UtBAC/src/Revise.jl:526
│     [14] revise at /home/fredrik/.julia/packages/Revise/UtBAC/src/Revise.jl:551 [inlined]
│     [15] run_backend(::REPL.REPLBackend) at /home/fredrik/.julia/packages/Revise/UtBAC/src/Revise.jl:941
│     [16] (::getfield(Revise, Symbol("##74#76")){REPL.REPLBackend})() at ./task.jl:259
└ @ Revise ~/.julia/packages/Revise/UtBAC/src/lowered.jl:52
2
```
Maybe there is something funky going on with `@pure`-usage in `Tensors.jl`? As noted, removing the multiplication solves the problem.

timholy:
Part of the problem is detecting the llvmcall (esp in #112), but a bigger issue is that we need to check [`isempty(sparams)`](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/04d1fb5ce969d7e41df6f7bb9dd52d96049df1bb/src/optimize.jl#L165) because our [mini-interpreter](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/04d1fb5ce969d7e41df6f7bb9dd52d96049df1bb/src/optimize.jl#L277-L291) runs at toplevel and can't access `Expr(:static_parameter, n)`. I'm sure this is fixable, but this PR is a workaround that skips interpreting such methods altogether. It's probably OK because I think most methods that contain `llvmcall` are low-level and unlikely to be of interest to the majority of people doing debugging.

fixes #112, fixes #288

### DIFF
--- a/test/interpret.jl
+++ b/test/interpret.jl
@@ -513,6 +513,36 @@ function f_mmap()
 end
 @interpret f_mmap()
 
+# parametric llvmcall (issues #112 and #288)
+module VecTest
+    using Tensors
+    Vec{N,T} = NTuple{N,VecElement{T}}
+    # The following test mimic SIMD.jl
+    const _llvmtypes = Dict{DataType, String}(
+        Float64 => "double",
+        Float32 => "float",
+        Int32 => "i32",
+        Int64 => "i64"
+    )
+    @generated function vecadd(x::Vec{N, T}, y::Vec{N, T}) where {N, T}
+        llvmT = _llvmtypes[T]
+        func = T <: AbstractFloat ? "fadd" : "add"
+        exp = """
+        %3 = $(func) <$(N) x $(llvmT)> %0, %1
+        ret <$(N) x $(llvmT)> %3
+        """
+        return quote
+            Base.@_inline_meta
+            Core.getfield(Base, :llvmcall)($exp, Vec{$N, $T}, Tuple{Vec{$N, $T}, Vec{$N, $T}}, x, y)
+        end
+    end
+    f() = 1.0 * one(Tensor{2,3})
+end
+let a = (VecElement{Float64}(1.0), VecElement{Float64}(2.0))
+    @test @interpret(VecTest.vecadd(a, a)) == VecTest.vecadd(a, a)
+end
+@test @interpret(VecTest.f()) == [1 0 0; 0 1 0; 0 0 1]
+
 # Test exception type for undefined variables
 f() = s = s + 1
 @test_throws UndefVarError @interpret f()
--- a/Project.toml
+++ b/Project.toml
@@ -17,7 +17,8 @@ Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 Mmap = "a63ad114-7e13-5084-954f-fe012c677804"
 SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
+Tensors = "48a634ad-e948-5137-8d70-aa71f2a747f4"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Distributed", "Dates", "SHA", "Mmap"]
+test = ["Test", "Dates", "Distributed", "Mmap", "SHA", "Tensors"]
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -165,6 +165,11 @@ function prepare_framecode(method::Method, @nospecialize(argtypes); enter_genera
                 generator = false
             end
         end
+        # Currenly, our strategy to deal with llvmcall can't handle parametric functions
+        # (the "mini interpreter" runs in module scope, not method scope)
+        if !isempty(lenv) && (hasarg(isequal(:llvmcall), code.code) || hasarg(a->is_global_ref(a, Base, :llvmcall), code.code))
+            return Compiled()
+        end
         framecode = FrameCode(method, code; generator=generator)
         if is_generated(method) && !enter_generated
             genframedict[(method, argtypes)] = framecode
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -87,6 +87,19 @@ function scan_ssa_use!(used::BitSet, @nospecialize(stmt))
     end
 end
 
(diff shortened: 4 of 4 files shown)


KristofferC:
Yeah, should be fine to compile these.

timholy:
Also note #285 could be fixed by the same cheat.

test/interpret.jl
```diff
@@ -513,6 +513,36 @@ function f_mmap()
...
+            Base.@_inline_meta
+            Core.getfield(Base, :llvmcall)($exp, Vec{$N, $T}, Tuple{Vec{$N, $T}, Vec{$N, $T}}, x, y)
+        end
+    end
+    f() = 1.0 * one(Tensor{2,3})
+end
+let a = (VecElement{Float64}(1.0), VecElement{Float64}(2.0))
+    @test @interpret(VecTest.vecadd(a, a)) == VecTest.vecadd(a, a)
+end
+@test VecTest.f() == [1 0 0; 0 1 0; 0 0 1]
```
fredrikekre: Should this be `@interpret`?

PR #292, opened 2019-05-25
Add some commented-out code for profiling

timholy:
This allows you to count the number of times each statement gets executed. I used this in https://discourse.julialang.org/t/evaluation-speed-during-debugging/24520/13. Seems like it might be worth keeping around.


timholy:
Any objections?

PR #303, opened 2019-07-22
Fix documentation of `determine_method_for_expr`

### LINKED ISSUE #302: The doc example of `determine_method_for_expr` spits error
The documentation example below fails:
> https://juliadebug.github.io/JuliaInterpreter.jl/stable/dev_reference/#JuliaInterpreter.determine_method_for_expr
```julia
JuliaInterpreter.determine_method_for_expr(:(sum([1,2])))
```

> Error
```
ERROR: MethodError: objects of type Symbol are not callable
Stacktrace:
 [1] #prepare_call#14(::Bool, ::Function, ::Any, ::Array{Any,1}) at C:\Users\aviat\.julia\packages\JuliaInterpreter\Or807\src\construct.jl:242
 [2] (::getfield(JuliaInterpreter, Symbol("#kw##prepare_call")))(::NamedTuple{(:enter_generated,),Tuple{Bool}}, ::typeof(JuliaInterpreter.prepare_call), ::Symbol, ::Array{Any,1}) at none:0
 [3] #determine_method_for_expr#21(::Bool, ::Function, ::Expr) at C:\Users\aviat\.julia\packages\JuliaInterpreter\Or807\src\construct.jl:504
 [4] determine_method_for_expr(::Expr) at C:\Users\aviat\.julia\packages\JuliaInterpreter\Or807\src\construct.jl:496
 [5] top-level scope at none:
```

I've found the `to_function` doesn't return the function but rather just return the symbol itself.
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/132bd507f89c3ec705435e6440f0112daebf463f/src/construct.jl#L496
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/132bd507f89c3ec705435e6440f0112daebf463f/src/utils.jl#L25

aviatesk:
The `$` character in [this line](https://github.com/aviatesk/JuliaInterpreter.jl/blob/01646ad1f035a8efa7f5a67465972ed72b84cc6c/src/construct.jl#L492) is currently not escaped and is being disappeared from the documentation.

This is better to be fixed since `determine_method_for_expr(:(sin([1, 2])))` is not the correct way to use this function and even causes an error.

- Closes #302

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -489,7 +489,7 @@ end
     framecode, frameargs, lenv, argtypes = determine_method_for_expr(expr; enter_generated = false)
 
 Prepare all the information needed to execute a particular `:call` expression `expr`.
-For example, try `JuliaInterpreter.determine_method_for_expr(:($(sum)([1,2])))`.
+For example, try `JuliaInterpreter.determine_method_for_expr(:(\$sum([1,2])))`.
 See [`JuliaInterpreter.prepare_call`](@ref) for information about the outputs.
 """
 function determine_method_for_expr(expr; enter_generated = false)


KristofferC:
Thanks!

PR #307, opened 2019-07-27
Replace the slow slotname Dict with a counter ("age")-based mechanism

timholy:
Old benchmarks:
```
          "recursive other 1_000" => Trial(7.510 ms)
          "recursive self 1_000" => Trial(2.450 ms)
          "long function 5_000" => Trial(33.257 ms)
          "throw long 1_000" => Trial(14.002 ms)
          "tight loop 10_000" => Trial(213.216 ms)
```

vs this branch:
```
         "recursive other 1_000" => Trial(7.502 ms)
          "recursive self 1_000" => Trial(1.996 ms)
          "long function 5_000" => Trial(30.500 ms)
          "throw long 1_000" => Trial(13.406 ms)
          "tight loop 10_000" => Trial(174.004 ms)
```

Take with a grain of salt, there is a fair amount of variance. But the last one is consistently cut to about 80% of the original.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -70,6 +70,7 @@ struct FrameCode
     src::CodeInfo
     methodtables::Vector{Union{Compiled,TypeMapEntry}} # line-by-line method tables for generic-function :call Exprs
     breakpoints::Vector{BreakpointState}
+    slotnamelists::Dict{Symbol,Vector{Int}}
     used::BitSet
     generator::Bool   # true if this is for the expression-generator of a @generated function
 end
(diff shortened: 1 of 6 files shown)


src/types.jl
```diff
@@ -176,10 +180,11 @@ mutable struct Frame
     framecode::FrameCode
     framedata::FrameData
     pc::Int
+    assignment_counter::Int
```
KristofferC: Could potentially make this a concrete `Int64` if we worry at all a out overflow (on 32-bit)

timholy:
Does the approval mean you know it doesn't break Debugger.jl? How about Juno?

KristofferC:
It means the PR on its own is good but it is indeed breaking Debugger. Will fix that up tomorrow.

KristofferC:
https://github.com/JuliaDebug/Debugger.jl/pull/198 should fix the Debugger.jl side.

timholy:
Let's wait for @pfitzseb to get a chance to look it over, in case it has more negatives than we might imagine.

KristofferC:
https://github.com/JunoLab/Atom.jl/pull/148 should fix the Juno side.

KristofferC:
I think we should just put `eval_code` into this package.

pfitzseb:
Agreed, it feels bad to have to keep both copies in sync...

KristofferC:
Ok, with `eval_code` in this repo and updated in this PR I will merge this.

PR #308, opened 2019-07-28
Create a "junk yard" even for Frames

KristofferC:
This avoids creating all the tiny `Frame` objects over and over.

Running the tight loop benchmark:

```jl
# Master
julia> @btime @interpret f(X)
  215.089 ms (1979122 allocations: 64.04 MiB)

# PR
julia> @btime @interpret f(X)
  207.504 ms (1749055 allocations: 49.24 MiB)
```

The time difference is not that big but the allocations are only 75% and it is possible this will start to matter more when other things are more optimized.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 deps/build.log
 docs/build/
 test/results.md
+Manifest.toml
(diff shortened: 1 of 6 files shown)


timholy:
Oh, interesting. Yes, I guess if we speed up the rest, this becomes important.

KristofferC:
It might improve GC times since we have a lot fewer objects to track as well. But for now, it doesn't seem like that big of a deal.

timholy:
I'm guessing the "one long vector" for all slots & ssavalues might supplant the need for this? It would also solve another problem: currently the caller packs the args into a vector, and then when you switch to the new frame they get copied again into the frame's slots. If we had one long vector, we could pack the args into it and thus do it only once.

PR #310, opened 2019-07-30
refactor out returning from a frame

KristofferC:
We might want to use a linear stack of memory in the future. It is then important that we unwind correctly (i.e return to the saved "stack pointer" after exiting the function).

This PR factors out a pattern we commonly used when returning from a function to simplify adding future functionality when this happens. I called it `unwind` but perhaps something like `return_from` is better?

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -246,7 +246,7 @@ function evaluate_call_recurse!(@nospecialize(recurse), frame::Frame, call_expr:
     end
     isa(ret, BreakpointRef) && return ret
     frame.callee = nothing
-    recycle(newframe)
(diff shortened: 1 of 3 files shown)


src/interpret.jl
```diff
@@ -581,10 +581,8 @@ function handle_err(@nospecialize(recurse), frame, err)
         return BreakpointRef(frame.framecode, frame.pc, err)
     end
     if isempty(data.exception_frames)
-        is_root_frame = frame.caller === nothing
```
KristofferC: The check for `is_root_frame` doesn't seem necessary.

PR #311, opened 2019-07-31
move sparams initialization into framedata creation

KristofferC:
This seemed like a better place to put it since we already for example initialize the local arguments in that function.


timholy:
Love it!

PR #313, opened 2019-07-31
use our own struct for the local method table instead of TypeMapEntry

KristofferC:
No performance difference but I feel this should be easier to maintain.

Slightly lifted from #309

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -53,9 +53,10 @@ function breakpointchar(bps::BreakpointState)
     return bps.condition === falsecondition ? ' ' : 'd'     # no breakpoint : disabled
 end
 
+abstract type AbstractFrameInstance end
 mutable struct DispatchableMethod
     next::Union{Nothing,DispatchableMethod}  # linked-list representation
-    frameinstance::Any # really a FrameInstance but we have a cyclic dependency
+    frameinstance::Union{Compiled, AbstractFrameInstance} # really a Union{Compiled, FrameInstance} but we have a cyclic dependency
     sig::Type # for speed of matching, this is a *concrete* signature. `sig <: frameinstance.framecode.scope.sig`
 end
 
@@ -124,7 +125,7 @@ Fields:
 - `framecode`: the [`FrameCode`](@ref) for the method.
 - `sparam_vals`: the static parameter values for the method.
 """
-struct FrameInstance
+struct FrameInstance <: AbstractFrameInstance
     framecode::FrameCode
     sparam_vals::SimpleVector
     enter_generated::Bool
--- a/src/types.jl
+++ b/src/types.jl
@@ -53,6 +53,12 @@ function breakpointchar(bps::BreakpointState)
     return bps.condition === falsecondition ? ' ' : 'd'     # no breakpoint : disabled
 end
 
+mutable struct DispatchableMethod
+    next::Union{Nothing,DispatchableMethod}  # linked-list representation
+    frameinstance::Any # really a FrameInstance but we have a cyclic dependency
+    sig::Type # for speed of matching, this is a *concrete* signature. `sig <: frameinstance.framecode.scope.sig`
+end
+
 """
 `FrameCode` holds static information about a method or toplevel code.
 One `FrameCode` can be shared by many calling `Frame`s.
@@ -68,7 +74,7 @@ Important fields:
 struct FrameCode
     scope::Union{Method,Module}
     src::CodeInfo
-    methodtables::Vector{Union{Compiled,TypeMapEntry}} # line-by-line method tables for generic-function :call Exprs
+    methodtables::Vector{Union{Compiled,DispatchableMethod}} # line-by-line method tables for generic-function :call Exprs
     breakpoints::Vector{BreakpointState}
     used::BitSet
     generator::Bool   # true if this is for the expression-generator of a @generated function
@@ -80,7 +86,7 @@ function FrameCode(scope, src::CodeInfo; generator=false, optimize=true)
         src, methodtables = optimize!(copy_codeinfo(src), scope)
     else
         src = replace_coretypes!(copy_codeinfo(src))
-        methodtables = Vector{Union{Compiled,TypeMapEntry}}(undef, length(src.code))
+        methodtables = Vector{Union{Compiled,DispatchableMethod}}(undef, length(src.code))
     end
     breakpoints = Vector{BreakpointState}(undef, length(src.code))
     for (i, pc_expr) in enumerate(src.code)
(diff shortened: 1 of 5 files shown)


src/types.jl
```diff
@@ -53,6 +53,12 @@ function breakpointchar(bps::BreakpointState)
     return bps.condition === falsecondition ? ' ' : 'd'     # no breakpoint : disabled
 end
 
+mutable struct DispatchableMethod
+    next::Union{Nothing,DispatchableMethod}  # linked-list representation
+    frameinstance::Any # really a FrameInstance but we have a cyclic dependency
+    sig::Type # for speed of matching, this is a *concrete* signature. `sig <: frameinstance.framecode.scope.sig`
```
timholy: If you're doing this, can you implement the `compiled` idea too? With this diff
```diff
tim@diva:~/.julia/dev/JuliaInterpreter$ git diff
diff --git a/src/construct.jl b/src/construct.jl
index 9b70786..796c4ab 100644
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -236,6 +236,7 @@ function prepare_call(@nospecialize(f), allargs; enter_generated = false)
     end
     argtypesv = Any[_Typeof(a) for a in allargs]
     argtypes = Tuple{argtypesv...}
+    println("Calling which (super-slow)")
     method = whichtt(argtypes)
     if method === nothing
         # Call it to generate the exact error
```
and the following test code
```julia
using JuliaInterpreter

bar(::Int) = 1
bar(::Float64) = 2
push!(JuliaInterpreter.compiled_methods, @which bar(1.0))

function foo(A)
    s = 0
    for a in A
        @bp
        s += bar(a)
    end
    return s
end
A = Real[1, 1, 1, 1, 1.0, 1.0, 1.0, 1.0]
frame = JuliaInterpreter.enter_call(foo, A)
for i = 1:length(A)
    println("iteration ", i)
    debug_command(frame, :c)
    for i = 1:3
        debug_command(frame, :se)  # step forward to the call to bar
    end
    # display(frame.framecode.methodtables)
end
```
I think you'll see what I'm worried about. (Note that the `Int` method of `bar` is recursively interpreted and the `Float64` one in `Compiled` mode.) If you inspect `frame.framecode.methodtables` you'll see that `Compiled` methods can't exploit the local method tables.

In the longer run we want to store the `MethodInstance`, and then use a `ccall` to skip the `jl_lookup_generic` from [`jl_apply_generic`](https://github.com/JuliaLang/julia/blob/5aea4569e988c6e231a2d9204c7e39193b961891/src/gf.c#L2206-L2214). That would basically mean DLLEXPORTing [this](https://github.com/JuliaLang/julia/blob/5aea4569e988c6e231a2d9204c7e39193b961891/src/gf.c#L2037-L2051). That will eliminate the runtime dispatch on compiled dispatch from the interpreter.

timholy: Correction: there's already [this](https://github.com/JuliaLang/julia/blob/5aea4569e988c6e231a2d9204c7e39193b961891/src/gf.c#L2053-L2057), so we don't have to modify Julia at all.

KristofferC: I think I am doing something wrong heh

```
julia> mi = Core.Compiler.specialize_method(@which(sin(2.0)), Tuple{typeof(sin), Float64}, Core.svec(Float64))
MethodInstance for sin(::Float64)

julia> args = [2.0]
1-element Array{Float64,1}:
 2.0

julia> ccall(:jl_invoke, Any, (Any, Vector{Any}, Cuint, Any), sin, args, 1, mi)
2.255484558e-314

julia> sin(args...)
0.9092974268256817
```

KristofferC: Anyway, I added caching for Compiled calls as well.

KristofferC: Ok, `args` should be a `Vector{Any}` and the args type should be `Ptr{Any}`.

KristofferC:
```jl
function jl_generic(args)
    +(args...)
end

function jl_invoke(args, mi)
     ccall(:jl_invoke, Any, (Any, Ptr{Any}, Cuint, Any), +, args, 2, mi)
end

args = Any[1, 2]
@btime jl_generic($args)
# 42 ns
mi = Core.Compiler.specialize_method(@which(1 + 1), Tuple{typeof(+), Int, Int}, Core.svec())
@btime foo($args, $mi)
# 11 ns
```

timholy: Not quite sure I get what `foo` is, but if you're testing `jl_invoke` against run-time dispatch keep in mind that the timings are rather dependent on how many methods there are. `convert` might be in a different category :smile:.

src/types.jl
```diff
@@ -68,7 +75,7 @@ Important fields:
 struct FrameCode
     scope::Union{Method,Module}
     src::CodeInfo
-    methodtables::Vector{Union{Compiled,TypeMapEntry}} # line-by-line method tables for generic-function :call Exprs
+    methodtables::Vector{Union{Compiled,DispatchableMethod}} # line-by-line method tables for generic-function :call Exprs
```
timholy: I don't see how this prevents `which`-based lookup on each iteration of a loop. In #309 I added a `compiled` field to `DispatchableMethod`, which seems like the right place for this information (once you've determined that a method matches, your next question is whether you should run it in interpreted or compiled mode).

src/types.jl
```diff
@@ -53,6 +53,13 @@ function breakpointchar(bps::BreakpointState)
     return bps.condition === falsecondition ? ' ' : 'd'     # no breakpoint : disabled
 end
 
+abstract type AbstractFrameInstance end
+mutable struct DispatchableMethod
+    next::Union{Nothing,DispatchableMethod}  # linked-list representation
+    frameinstance::Union{Compiled, AbstractFrameInstance} # really a Union{Compiled, FrameInstance} but we have a cyclic dependency
```
timholy: This could hurt. Union-splitting may not work, unless the compiler is smart enough to know there is only one concrete subtype.

KristofferC: Yeah but we type assert immediately when we use it anyway.

src/types.jl
KristofferC: It prevents it because `frameinstance` in `DispatchableMethod`is a `Compiled` in those cases.

src/types.jl
KristofferC: Sorry, had a typo, `foo` should have been `jl_invoke`. Just wanted to show that using `jl_invoke` with a MethodInstance was faster :)

src/types.jl
KristofferC: The reason I still want to have `Compiled` directly in `methodtables` is for ccall / fcall where we know for sure that the method we want to call are compiled so there is no need to do any matching against signatures.

src/types.jl
KristofferC: Looked a bit, there is a single place where we look at this field and all uses either just calls `isa` on it, or type asserts before it is used so there should be no dynamic behavior.

timholy:
Thanks for doing this and all the other PRs. I've been out on vacation (still am, actually), it's great to have the package in such good hands!

PR #319, opened 2019-08-12
Minor cleanup

oxinabox:
I was reading the code, and noticed a few minor mostly stylistic things.


src/construct.jl
```diff
@@ -241,14 +241,12 @@ function prepare_call(@nospecialize(f), allargs; enter_generated = false)
     elseif any(is_vararg_type, allargs)
         return nothing  # https://github.com/JuliaLang/julia/issues/30995
     end
-    argtypesv = Any[_Typeof(a) for a in allargs]
-    argtypes = Tuple{argtypesv...}
+    argtypes = Tuple{_Typeof.(allargs)...}
```
oxinabox: I'm not sure if this is the optimal way.
Annoyingly that depends on the input types

```
julia> f1(x) = Tuple{typeof.(x)...}
f1 (generic function with 1 method)

julia> @btime f1($(1, 1.0, 1f0, 'a', "a"));
  0.031 ns (0 allocations: 0 bytes)

julia> @btime f1($([1, 1.0, 1f0, 'a', "a"]));
  604.374 ns (11 allocations: 512 bytes)a

julia> f2(x) = Tuple{[typeof(xi) for xi in x]...}
f2 (generic function with 1 method)

julia> @btime f2($(1, 1.0, 1f0, 'a', "a"));
  329.788 ns (5 allocations: 192 bytes)

julia> @btime f2($([1, 1.0, 1f0, 'a', "a"]));
  604.374 ns (11 allocations: 512 bytes)a

julia> f3(x) = Tuple{(typeof(xi) for xi in x)...}
f3 (generic function with 1 method)

julia> @btime f3($(1, 1.0, 1f0, 'a', "a"));
  604.374 ns (11 allocations: 512 bytes)

julia> @btime f3($([1, 1.0, 1f0, 'a', "a"]));
  604.374 ns (11 allocations: 512 bytes)allocations
```

KristofferC: Anyway, this shouldn't be part of a "minor cleanup" PR.

oxinabox: This can wait for some other day then.

```suggestion
    argtypesv = Any[_Typeof(a) for a in allargs]
    argtypes = Tuple{argtypesv...}
```

PR #326, opened 2019-08-16
only test TableReader on 64 bit

KristofferC:


timholy:
FWIW I submitted a PR to fix TableReader on Julia 1.3+, see https://github.com/bicycle1885/TableReader.jl/pull/59

PR #331, opened 2019-08-31
Check that the id appears in Base.loaded_modules

timholy:
This is a little weird, ~~and I have not been able to come up with a good test~~, but it's needed for Revise issue 302.


timholy:
Let's hold off on merging this. I should probably poke at it a bit more.

timholy:
OK, I figured out why this was happening, and this PR does indeed make sense. From the (new) commit message:

> The problem is that if there's a file on the LOAD_PATH with the name of the module, Base.identify_package will return a PkgId for it. This section of the code is trying to avoid *re*-defining a module, so we need to check that if we create a PkgId that it corresponds to a
loaded module.

Issue #330, opened 2019-08-31
Segfault with IO locking


timholy:
On recent Julia versions (I'm using 1.4.0-DEV.77), I fear that anything calling `iolock_begin`/`iolock_end` may be problematic. Demo:

```julia
julia> using JuliaInterpreter

julia> repl = Base.active_repl;

julia> @interpret displaysize(repl.t.out_stream)

signal (11): Segmentation fault
in expression starting at REPL[3]:1
jl_eh_restore_state at /home/tim/src/julia-master/src/rtutils.c:254
step_expr! at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:55
step_expr! at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:556
unknown function (ip: 0x7fd83fa88771)
finish! at /home/tim/.julia/dev/JuliaInterpreter/src/commands.jl:14
finish_and_return! at /home/tim/.julia/dev/JuliaInterpreter/src/commands.jl:29 [inlined]
#evaluate_call_recurse!#46 at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:239
evaluate_call_recurse! at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:202 [inlined]
eval_rhs at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:378
step_expr! at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:517
step_expr! at /home/tim/.julia/dev/JuliaInterpreter/src/interpret.jl:556
unknown function (ip: 0x7fd83fa88771)
finish! at /home/tim/.julia/dev/JuliaInterpreter/src/commands.jl:14
finish_and_return! at /home/tim/.julia/dev/JuliaInterpreter/src/commands.jl:29 [inlined]
finish_and_return! at /home/tim/.julia/dev/JuliaInterpreter/src/commands.jl:33 [inlined]
finish_and_return! at /home/tim/.julia/dev/JuliaInterpreter/src/commands.jl:33
_jl_invoke at /home/tim/src/julia-master/src/gf.c:2136 [inlined]
jl_apply_generic at /home/tim/src/julia-master/src/gf.c:2300
jl_apply at /home/tim/src/julia-master/src/julia.h:1632 [inlined]
do_call at /home/tim/src/julia-master/src/interpreter.c:328
[...]

PR #335, opened 2019-09-08
Don't interpret methods calling iolock_begin

timholy:
This is a temporary workaround for #330


timholy:
OK to make a new release? Not many changes, but seems like an important one for people running nightly. Will also prevent segfaults in the Revise tests :smile:.

KristofferC:
Yep, go for it

PR #341, opened 2019-09-28
Add hooks to breakpoint changes

### LINKED ISSUE #261: Listeners for global state
It would be really cool if consumers of this package could register event listeners for changes in global state (meaning breakpoints and the `break_on` stuff) to keep the UI synced with JuliaInterpreter's internal state. This mainly comes up if the global state is changed directly with JuliaInterpreter.
Thoughts?

oxinabox:
Closes #261

I think the hooks fire too often.
For when creating breakpoints,
so there are two broken tests.

I only made 1 hook for all changes because I started to argue with myself as to if removal should trigger both the removal hook and the disable hook,
and concluded that i could just have one hook for everything than worry about those details,
that can be determined from the `BreakpointRef` anyway.

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -7,18 +7,34 @@ Return an array with all breakpoints.
 """
 breakpoints() = _breakpoints
 
-"""
-    breakpoint_update_hooks
 
-This is a collection of "hook" functions, each taking as input function, and a `AbstractBreakpoint`.
-After the global state in relation breakpoints is updated each this hook is called,
-passing in the triggering function, and the updated/created/removed breakpoint.
-The trigger function should not be called, and is provided for dispatch purposes.
-"""
 const breakpoint_update_hooks = []
+"""
+    on_breakpoints_updated(f)
+
+Register a one-argument function to be called after any update to the global set of
+breakpoints. This includes their creation, deletion, enabling and disabling.
+
+The function `f` should take two inputs:
+First argument is the function doing to update, such as `::typeof(breakpoint)` for the
+creation, or `::typeof(remove)` for the deletion.
+Second argument is the breakpoint object that was changed.
+If only desiring to handle some kinds of update, `f` should have fallback methods
+to do nothing in the general case.
+
+!!! warning
+    This feature is experimental, and may be modified or removed without the tagging of a
+    breaking release.
+"""
+on_breakpoints_updated(f) = push!(breakpoint_update_hooks, f)
+
 function firehooks(hooked_fun, bp::AbstractBreakpoint)
     for hook in breakpoint_update_hooks
-        hook(hooked_fun, bp)
+        try
+            hook(hooked_fun, bp)
+        catch err
+            @warn "Hook function errored" hook hooked_fun bp exception=err
+        end
     end
 end
 
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -119,7 +119,13 @@ function breakpoint(f::Union{Method, Function}, sig=nothing, line::Integer=0, co
     bp = BreakpointSignature(f, sig, line, condition, Ref(true), BreakpointRef[])
     add_to_existing_framecodes(bp)
     idx = findfirst(bp2 -> same_location(bp, bp2), _breakpoints)
-    idx === nothing ? push!(_breakpoints, bp) : (_breakpoints[idx] = bp)
+    if idx === nothing  # creating new
+        push!(_breakpoints, bp)
+    else  #Replace existing breakpoint
+        old_bp = _breakpoints[idx]
+        _breakpoints[idx] = bp
+        firehooks(remove, old_bp)
+    end
     firehooks(breakpoint, bp)
     return bp
 end
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -202,7 +202,7 @@ breakpoint!(frame::Frame, pc=frame.pc, condition::Condition=nothing) =
 
 function update_states!(bp::AbstractBreakpoint)
     foreach(bpref -> update_state!(bpref, bp.enabled[]), bp.instances)
-    firehooks(update_state!, bp)
+    firehooks(update_states!, bp)
 end
 update_state!(bp::BreakpointRef, v::Bool) = bp[] = v
 
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -7,6 +7,16 @@ Return an array with all breakpoints.
 """
 breakpoints() = _breakpoints
 
+"""
+    breakpoint_update_hooks
+
+This is a collection of "hook" functions, each taking as input a `BreakpointRef`.
+Whenever something that updates the global state in relation to this hook is called,
+each hook function in this list will be called in turn, getting as input the updated `bp`.
+"""
+const breakpoint_update_hooks = []
+firehooks(bp::BreakpointRef) = foreach(hook->hook(bp), breakpoint_update_hooks)
+
 function add_to_existing_framecodes(bp::AbstractBreakpoint)
     for framecode in values(framedict)
         add_breakpoint_if_match!(framecode, bp)
@@ -17,7 +27,9 @@ function add_breakpoint_if_match!(framecode::FrameCode, bp::AbstractBreakpoint)
     if framecode_matches_breakpoint(framecode, bp)
         stmtidx = bp.line === 0 ? 1 : statementnumber(framecode, bp.line)
         breakpoint!(framecode, stmtidx, bp.condition, bp.enabled[])
-        push!(bp.instances, BreakpointRef(framecode, stmtidx))
+        bp_ref = BreakpointRef(framecode, stmtidx)
+        firehooks(bp_ref)
+        push!(bp.instances, bp_ref)
     end
 end
 
@@ -184,7 +196,11 @@ breakpoint!(frame::Frame, pc=frame.pc, condition::Condition=nothing) =
     breakpoint!(frame.framecode, pc, condition)
 
 update_states!(bp::AbstractBreakpoint) = foreach(bpref -> update_state!(bpref, bp.enabled[]), bp.instances)
-update_state!(bp::BreakpointRef, v::Bool) = bp[] = v
+function update_state!(bp::BreakpointRef, v::Bool)
+    bp[] = v
+    firehooks(bp)
+    v
+end
 
 """
     enable(bp::AbstractBreakpoint)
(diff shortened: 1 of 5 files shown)


Project.toml
```diff
@@ -13,14 +13,14 @@ CodeTracking = "0.3.2, 0.4, 0.5"
 julia = "1"
 
 [extras]
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
```
oxinabox: ah Pkg3, editting Project.toml files against my will.

src/breakpoints.jl
```diff
@@ -17,7 +27,9 @@ function add_breakpoint_if_match!(framecode::FrameCode, bp::AbstractBreakpoint)
     if framecode_matches_breakpoint(framecode, bp)
         stmtidx = bp.line === 0 ? 1 : statementnumber(framecode, bp.line)
         breakpoint!(framecode, stmtidx, bp.condition, bp.enabled[])
-        push!(bp.instances, BreakpointRef(framecode, stmtidx))
+        bp_ref = BreakpointRef(framecode, stmtidx)
+        firehooks(bp_ref)
```
oxinabox: This is the _create_ hook,
but maybe it should be somewhere else?

oxinabox:
I think want two series of hooks.
The `update` hook, which is what we have now, that always gets a `BreakpointRef`
and the `create` hook, which gets a `BreakpointSignature`

oxinabox:
Bump

KristofferC:
I didn't understand the need for the hooks really. Is there some "reference implementation" that uses them?

oxinabox:
The general concept of when working with a UI, you attach listeners or subscribers to some thing in the model layer, and when it changes state those get notified/called.
Hooks are just that but with less indirection (like attaching a hook that calls `@async` would be basically exactly that)

And the main reason to have this is pretty much for having a UI.

So lets say that Juno displayed little dots in the margin for where breakpoints were.
And you could set breakpoints by clicking the margin to create those dots.
And maybe Juno also has a list of breakpoints in a breakpoint pane.

Then you both those to update if someone types `@breakpoint mysum(A) 15` in the REPL

KristofferC:
Yes, but I thought this was for cassette.

For Juno, I can see how a hook on when an `AbstractBreakpoint` is added is useful but right now, no callback gets called when a user does `breakpoint(sin)`. Instead, the callback fires when you do `@interpret sin(2.0)` and we create the `FrameCode` for `sin(::Float64)` for the first time. That doesn't seem useful from a UI perspective.

oxinabox:
> right now, no callback gets called when a user does breakpoint(sin). Instead, the callback fires when you do @interpret sin(2.0) and we create the FrameCode for sin(::Float64) for the first 

Yes, I think that is a mistake. See my comment from earlier:

> I think want two series of hooks.
The update hook, which is what we have now, that always gets a BreakpointRef
and the create hook, which gets a BreakpointSignature

I added the one on FrameCode because I worked out `BreakPointRef` first.
I figured I would open the PR, so I could get this feedback.

But maybe we don't need any on `BreakpointRef` just on `AvstractBreakpoint`

The operations that want hooks are creation, deletion, enable and disable.
I think that's all.

oxinabox:
I have changed the hooks over to be on the AbstractBreakpoints,
and made them also get the hooking functionm so can dispatch on that

src/breakpoints.jl
```diff
@@ -96,6 +112,7 @@ function breakpoint(file::AbstractString, line::Integer, condition::Condition=no
     add_to_existing_framecodes(bp)
     idx = findfirst(bp2 -> same_location(bp, bp2), _breakpoints)
     idx === nothing ? push!(_breakpoints, bp) : (_breakpoints[idx] = bp)
+    firehooks(breakpoint, bp)
```
oxinabox: Should this if it is replacing an existing one,
also do 
`firehooks(remove, bp)` ?

oxinabox:
bump

KristofferC:
Sorry, I am quite busy with other things these days. Is there anything else that is being blocked on this. I think a concrete application of the functionality here would make sure that what is implemented here is actually what is needed.

oxinabox:
Yes, it blocks MixedModeDebugger.
Which is currently using this branch.
https://github.com/oxinabox/MixedModeDebugger.jl

I feel like the Juno case is more understandable though,
since there is prior art for how that is done.
but I don't think there is a Juno PR to use this.

Where-as MixedModeDebugger is very much its own weird thing and is kinda a hack to avoid having to heavily edit JuliaInterpretter and JuliaDebugger while the idea is prototyped.
Though even if the code in MixedModeDebugger was relocated into JuliaInterpretter and JuliaDebugger then it would effectively still need the hooks,
though they could be hardcoded in that case.

src/breakpoints.jl
```diff
@@ -7,6 +7,21 @@ Return an array with all breakpoints.
 """
 breakpoints() = _breakpoints
 
+"""
+    breakpoint_update_hooks
+
+This is a collection of "hook" functions, each taking as input function, and a `AbstractBreakpoint`.
```
KristofferC: This section needs proofreading. I think structuring the API similar to the `atreplinit` function is a good idea.

Also, please add that this feature is experimental.

src/breakpoints.jl
```diff
@@ -224,10 +248,7 @@ end
 Toggle breakpoint `bp`.
 """
 toggle(bp::AbstractBreakpoint) = (bp.enabled[] = !bp.enabled[]; update_states!(bp))
-function toggle(bp::BreakpointRef)
-    state = bp[]
-    bp.framecode.breakpoints[bp.stmtidx] = BreakpointState(!state.isactive, state.condition)
-end
+toggle(bp::BreakpointRef) = update_state!(bp, !bp[].isactive)
```
KristofferC: This doesn't seem equivalent to the previous code?

oxinabox: I think it is.
substituting in `update_state!`
This is equie to `bp[] = !bp[].isactive`

Now we look with those do
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/08333287554a7aa54df447ee13dd465c87afef8b/src/types.jl#L290-L292
Subtitupe in `setindex`
And we get
`bp.framecode.breakpoints[bp.stmtidx] = BreakpointState( !bp[].isactive, bp[].condition)`
extract out the `bp[]` and we get
```
state=bp[]
bp.framecode.breakpoints[bp.stmtidx] = BreakpointState( !state.isactive, state.condition)
```

Which is what was there before.

src/breakpoints.jl
oxinabox: Sounds good, I just took a look at that code,
I will make that change tomorrow.

oxinabox: done

oxinabox:
idk why it is broken on nightly

oxinabox:
bump

src/breakpoints.jl
```diff
@@ -7,6 +7,37 @@ Return an array with all breakpoints.
 """
 breakpoints() = _breakpoints
 
+
+const breakpoint_update_hooks = []
+"""
+    on_breakpoints_updated(f)
+
+Register a one-argument function to be called after any update to the global set of
```
timholy: `global` as opposed to what?

src/breakpoints.jl
```diff
@@ -7,6 +7,37 @@ Return an array with all breakpoints.
...
+breakpoints. This includes their creation, deletion, enabling and disabling.
+
+The function `f` should take two inputs:
```
timholy: Might be clearer to say that this function will be called as `f(action, bp)`, where `action` is `breakpoint`, `enable`, `remove`, etc. (it might be important to spell out the entire list of actions that need to be handled, though), and `bp` is the breakpoint object. Then give an example of writing such an `f`, perhaps in the Documenter docs.

src/breakpoints.jl
```diff
@@ -7,6 +7,37 @@ Return an array with all breakpoints.
...
+First argument is the function doing to update, such as `::typeof(breakpoint)` for the
+creation, or `::typeof(remove)` for the deletion.
+Second argument is the breakpoint object that was changed.
+If only desiring to handle some kinds of update, `f` should have fallback methods
+to do nothing in the general case.
+
+!!! warning
+    This feature is experimental, and may be modified or removed without the tagging of a
```
timholy: I think we could tag a minor bump if it gets removed.

src/breakpoints.jl
timholy: That seems like a good idea.

src/breakpoints.jl
```diff
@@ -7,6 +7,37 @@ Return an array with all breakpoints.
...
+    breaking release.
+"""
+on_breakpoints_updated(f) = push!(breakpoint_update_hooks, f)
+
+function firehooks(hooked_fun, bp::AbstractBreakpoint)
+    for hook in breakpoint_update_hooks
+        try
+            hook(hooked_fun, bp)
```
timholy: Does this need `invokelatest`?

test/breakpoints.jl
```diff
@@ -322,3 +322,39 @@ end
...
+@testset "hooks" begin
+    remove()
+    f_break(x) = x
+
+    # Check creating hits hook
+    empty!(breakpoint_update_hooks)
+    hook_hit = false
+    push!(breakpoint_update_hooks, (f,_)->hook_hit = f == breakpoint)
+    bp = breakpoint(f_break)
+    @test hook_hit
```
timholy: Probably want to test one that doesn't hit first.

src/breakpoints.jl
oxinabox: A fair point.

```suggestion
Register a one-argument function to be called after any update to the set of
```

src/breakpoints.jl
oxinabox: I would not expect the hook functions to be redefined during the execution of the program.
Juno of MixedModeDebugger would set them at the start and then done.

So I don't think it needs to be.

oxinabox: But it can be, if you think its better

test/breakpoints.jl
oxinabox: I'm not sure what that would like like

timholy: Do something to breakpoints that isn't always caught by your hook, and then `@test !hook_hit`.

oxinabox: Such as?
What code am I trying to test, for what failure case?

 - Do I want to test the definition of `(f,_)->hook_hit = f == breakpoint` ? (so I should test e.g. by doing a `remove`)
 - Do I want to test that random functions don't trigger hooks, so I should test by doing `1+1` or something?

timholy:
This seems fine to me. I'll plan to merge tomorrow if there are no objections.

timholy:
Thanks for your patience, @oxinabox!

timholy:
Is there any chance this is breaking? Or could this just be 0.7.8?

oxinabox:
I don't see how it could be breaking

timholy:
I don't either. Let's release this as 0.7.8.

PR #350, opened 2019-11-18
fix breakage on master due to _apply_iterate

KristofferC:
There is still an error in master in `limits.jl`:

```
Abort: Test Failed at /home/kc/JuliaPkgs/JuliaInterpreter.jl/test/limits.jl:39
  Expression: (Aborted(frame, 1)).at.file == Symbol("util.jl")
   Evaluated: var"fake.jl" == var"util.jl"
```

@timholy, do you have any idea about that one?


KristofferC:
I think the util failure is just due to some restructuring of the `@elapsed` macro on master. Will just condition it on the julia version.

PR #352, opened 2019-11-29
Look up correct framecode for invoke

timholy:
Prior to this the tests failed due to the following:
```julia
julia> using JuliaInterpreter

julia> sin(1)
0.8414709848078965

julia> @interpret sin(1)
0.8370984093755471
```
This turns out to be due to the new [`evalpoly`](https://github.com/JuliaLang/julia/blob/f39cdc0c7da8230b102051681c9378ce15d90235/base/math.jl#L106-L117), it was looking up the wrong framecode. This might have been a bug on all previous versions, but only detected now because of the change from macro to generated function.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -214,12 +214,13 @@ function evaluate_call_recurse!(@nospecialize(recurse), frame::Frame, call_expr:
     end
     if fargs[1] === Core.invoke # invoke needs special handling
         f_invoked = which(fargs[2], fargs[3])
-        sig = Tuple{_Typeof.(fargs)...}
+        fargs_pruned = [fargs[2]; fargs[4:end]]
+        sig = Tuple{_Typeof.(fargs_pruned)...}
         ret = prepare_framecode(f_invoked, sig; enter_generated=enter_generated)
         isa(ret, Compiled) && invoke(fargs[2:end]...)
         framecode, lenv = ret
-        fargs = [fargs[2]; fargs[4:end]]
         lenv === nothing && return framecode  # this was a Builtin
+        fargs = fargs_pruned
     else
         framecode, lenv = get_call_framecode(fargs, frame.framecode, frame.pc; enter_generated=enter_generated)
         if lenv === nothing
--- a/test/interpret.jl
+++ b/test/interpret.jl
@@ -613,3 +613,11 @@ end
     f() = map(x -> 2x, 1:10)
     @test @interpret(f()) == f()
 end
+
+@testset "invoke" begin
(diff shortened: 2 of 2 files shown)


timholy:
OSX and Windows test failures seem unrelated (`install_archive`).

jmert:
For what it's worth, this helps with a case that interprets an `invoke`d call with a where clause — I was trying to debug a code path, but kept getting that the type variables were unknown. Merging this locally fixed the issue.

```julia
julia> using Debugger, LinearAlgebra

julia> f(d::Diagonal{T}) where {T} = invoke(f, Tuple{AbstractMatrix}, d)

julia> f(m::AbstractMatrix{T}) where {T} = T

julia> D = Diagonal([1.0, 2.0])
2×2 Diagonal{Float64,Array{Float64,1}}:
 1.0   ⋅ 
  ⋅   2.0

julia> f(D)
Float64

julia> @run f(D)
Breaking for error:
ERROR: UndefVarError: T not defined
Stacktrace:
 [1] f(::Diagonal{Float64,Array{Float64,1}}) at REPL[3]:1
 [2] f(::Diagonal{Float64,Array{Float64,1}}) at REPL[2]:1

In f(m) at REPL[3]:1
>1  f(m::AbstractMatrix{T}) where {T} = T

About to run: return $(QuoteNode(:($(Expr(:static_parameter, 1)))))
1|debug> q

julia> 
```

timholy:
Thanks @jmert, I shamelessly stole this test and added it here.

timholy:
Why is this still using
```sh
Julia Version 1.4.0-DEV.535
Commit 9babbf5dc6 (2019-11-30 20:29 UTC)
```
There's a newer version available:
```sh
Julia Version 1.4.0-DEV.546
Commit 84e839227f (2019-12-03 02:31 UTC)
```
Interestingly, over at https://github.com/timholy/SnoopCompile.jl/pull/43 one sees the "push" build is using a different nightly from the "pr" build on linux. (OSX is using a consistent nightly.)

CC @StefanKarpinski

StefanKarpinski:
I don't know. I'm not really an expert on these CI systems. Perhaps @staticfloat knows?

staticfloat:
I don't know why.  I re-ran the job and it got a newer one.  There are levels of caching that happen between the CI services and the buildbot output, and the caching is implemented differently for different CI providers (such as MacOS on Travis or Linux on Travis, which are within different datacenters)

timholy:
While nightly continues to be problematic, on my machine master passes, so I'll merge this.

PR #365, opened 2020-02-11
check if newmod is module in split_expressions!

pfitzseb:
This throws a more informative error when calling `split_expressions!` on e.g.
```
struct Foo end
module Foo end
```


timholy:
Since this is just a more informative error, seems OK without tests.

Issue #368, opened 2020-02-11
Make global breakpoints work


davidanthoff:
Right now both Juno and VS Code manually insert breakpoints for global level statements, with code that looks like [this](https://github.com/JunoLab/Atom.jl/pull/269/files#diff-74535f4e40b9f850dc9f0d02ae8feb04R80).

I think it would be great if that wasn't necessary :) I'm already calling `breakpoint` for those breakpoints, and ideally that would be enough and no additional hacking on the AST was necessary.

The hacks we have right now also don't work for conditions on the breakpoints.

CC @pfitzseb

KristofferC:
First step is to enable them by removing https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/08333287554a7aa54df447ee13dd465c87afef8b/src/types.jl#L107.

Then figure out if there is any fallout.

PR #370, opened 2020-02-12
global breakpoints

pfitzseb:
Seems to work fine afaict, but still needs tests.

Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/368.

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -155,14 +155,25 @@ function breakpoint(file::AbstractString, line::Integer, condition::Condition=no
 end
 
 function framecode_matches_breakpoint(framecode::FrameCode, bp::BreakpointFileLocation)
-    framecode.scope isa Method || return false
-    meth = framecode.scope
-    methpath = CodeTracking.maybe_fix_path(String(meth.file))
-    ispath(methpath) && (methpath = realpath(methpath))
-    if bp.abspath == methpath || endswith(methpath, bp.path)
-        return method_contains_line(meth, bp.line)
+    if framecode.scope isa Method
+        meth = framecode.scope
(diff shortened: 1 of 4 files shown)


KristofferC:
Implementation looks good to me.

timholy:
Looks good. Without trying it I'm not sure what's causing the failure on 1.0.

pfitzseb:
Turns out there's [no line number node for the first line on Julia 1.0/1.1](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/f4316aeb338b7cc666ec6bc7351a428ca82194e5/src/construct.jl#L420):
```
julia> expr = Base.parse_input_line(expr, filename = "foo.jl")
:($(Expr(:toplevel, :(1 + 1), quote
    #= foo.jl:3 =#
    2 + 2
    #= foo.jl:4 =#
    3 + 3
...
```
vs
```
julia> expr = Base.parse_input_line(expr, filename = "foo.jl")
:($(Expr(:toplevel, :(#= foo.jl:1 =#), :(1 + 1), :(#= line 2 =#), quote
    #= foo.jl:3 =#
    2 + 2
    #= foo.jl:4 =#
    3 + 3
...
```

Not sure what we can do about that though -- maybe insert a new LNN for the first line or something?

KristofferC:
Imo no need to put too much effort into it. I would just make it as `@test_broken` on it.

pfitzseb:
Done.

KristofferC:
Exciting new errors on master, doesn't have to be dealt with here of course.

timholy:
Thanks a ton, @pfitzseb!

PR #372, opened 2020-02-13
Version 0.7.10

timholy:
Seem OK? Or is there more we want to get into this?


pfitzseb:
All in favor. Both the VSCode extension and Juno want the global breakpoints sooner than later, I think ;)

PR #375, opened 2020-02-13
Fix StackFrame for top level frames

### LINKED ISSUE #374: show_backtrace crashes if there is a Module scope at the root of a frame
I believe [this line](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/d874384bbb40d05e8c04d4962449cb29151be727/src/utils.jl#L532) crashes if `scopeof(frame) isa Method` is not true.

davidanthoff:
Fixes #374.

Right now I'm just creating a fake name in the case of a Module, I don't think it is shown anywhere, so hopefully ok.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -525,11 +525,13 @@ function Base.StackTraces.StackFrame(frame::Frame)
         sig = method.sig
         sparams = Core.svec(frame.framedata.sparams...)
         mi = specialize_method(method, atypes, sparams)
+        fname = frame.framecode.scope.name
     else
         mi = frame.framecode.src
+        fname = gensym()
     end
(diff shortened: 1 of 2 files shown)


KristofferC:
Would be good with a test, could be added next to:

https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/32c7a320f3164cad6ae33d8007bd55bd0dccf13d/test/interpret.jl#L448-L454

davidanthoff:
Alright, with tests now.

davidanthoff:
Ok, I think the test failures are only on nightly, which also fails on `master`, so I think this is ready to be merged?

PR #378, opened 2020-02-14
Update travis config

davidanthoff:
I think I got it right, but not sure :) Lets see whether tests pass. If this works, we could switch off the appveyor stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,18 @@ language: julia
 
 os:
   - linux
+  - windows
   - osx
 
 julia:
   - 1.0
-  - 1.1
-  - 1.2
   - 1.3
+  - 1.4
   - nightly
+
+arch:
+  - x86
+  - x64
 
 branches:
   only:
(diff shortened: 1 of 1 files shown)


fredrikekre:
You are not allowed to use `matrix` and `jobs` anymore, need to combine them.

davidanthoff:
Hm, I don't understand why this doesn't create 32 bit Windows builds? And at least in the UI right now it also doesn't seem to show allowed failure rows anymore? Weird...

.travis.yml
```diff
@@ -3,14 +3,20 @@ language: julia
 os:
   - linux
   - osx
+  - windows
 
 julia:
   - 1.0
   - 1.1
```
KristofferC: While changing this, we might as well remove 1.1. and 1.2.

.travis.yml
```diff
@@ -19,16 +25,20 @@ branches:
...
         - julia --project=docs/ docs/make.jl
       name: "HTML"
       after_success: skip
-
-after_success:
-  - julia -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
+  exclude:
+  - os: osx
+    arch: x86
+  allow_failures:
```
KristofferC: This should be removed imo. We want this package to work on nightly and some red crosses might make someone annoyed enough to look into it.

timholy:
Seems fine to me, but I won't merge since I'm unsure whether you're still working on the 32-bit Windows issue.

davidanthoff:
I don't really understand why the 32 bit stuff doesn't work and don't have any more ideas. I asked on Slack, but I guess we could merge this in the meantime and always do another PR if someone figures this out.

timholy:
I don't seem to have access to the AppVeyor settings. @KristofferC, can you shut it off?

davidanthoff:
Maybe we wait with shutting appveyor off until we have 32 bit windows build working on travis?

KristofferC:
Yeah, I think this https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/a0e209e046c8ad650525714c2117f26024a8ec44/src/optimize.jl#L333-L337 code path is only hit on 32 bit Windows so probably want to still test that.

Issue #383, opened 2020-03-03
MethodError: no method matching reinterpret(::SSAValue, ::SlotNumber)


maleadt:
JuliaInterpreter#master fails to interpret launching a GPU kernel with CUDAnative#master (tested on 1.4-rc):

```julia
julia> using CUDAnative

julia> f() = @cuda (()->nothing)()
f (generic function with 1 method)

julia> f()

julia> @interpret f()
ERROR: MethodError: no method matching reinterpret(::JuliaInterpreter.SSAValue, ::JuliaInterpreter.SlotNumber)
Closest candidates are:
  reinterpret(::Type{T}, ::Any) where T at essentials.jl:416
Stacktrace:
 [1] top-level scope at REPL[13]:1
 [2] eval at ./boot.jl:331 [inlined]
 [3] eval at /home/tim/Julia/pkg/JuliaInterpreter/src/JuliaInterpreter.jl:1 [inlined]
 [4] build_compiled_call!(::Expr, ::Symbol, ::Core.CodeInfo, ::Int64, ::Int64, ::Array{Symbol,1}, ::Module) at /home/tim/Julia/pkg/JuliaInterpreter/src/optimize.jl:286
 [5] optimize!(::Core.CodeInfo, ::Method) at /home/tim/Julia/pkg/JuliaInterpreter/src/optimize.jl:175
 [6] JuliaInterpreter.FrameCode(::Method, ::Core.CodeInfo; generator::Bool, optimize::Bool) at /home/tim/Julia/pkg/JuliaInterpreter/src/types.jl:88
 [7] prepare_framecode(::Method, ::Any; enter_generated::Bool) at /home/tim/Julia/pkg/JuliaInterpreter/src/construct.jl:189
 [8] prepare_call(::Any, ::Array{Any,1}; enter_generated::Bool) at /home/tim/Julia/pkg/JuliaInterpreter/src/construct.jl:260
 [9] get_call_framecode(::Array{Any,1}, ::JuliaInterpreter.FrameCode, ::Int64; enter_generated::Bool) at /home/tim/Julia/pkg/JuliaInterpreter/src/localmethtable.jl:63
 [10] evaluate_call_recurse!(::Any, ::Frame, ::Expr; enter_generated::Bool) at /home/tim/Julia/pkg/JuliaInterpreter/src/interpret.jl:225
[...]

PR #384, opened 2020-03-03
Only eval tuple specification to ccall, use dynamic semantics otherwise.

maleadt:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/383. Couldn't immediately come up with a test case that resulted in the invalid reinterpret.

Ultimately this is only a workaround, Base has a `static_eval` that returns NULL if the arguments aren't static (as with #383): https://github.com/JuliaLang/julia/blob/4e609128ff5c23424b175be300360ab07e90cba9/src/ccall.cpp#L541

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -283,7 +283,7 @@ function build_compiled_call!(stmt, fcall, code, idx, nargs, sparams, evalmod)
     end
     dynamic_ccall = false
     if isa(cfunc, Expr)   # specification by tuple, e.g., (:clock, "libc")
-        cfunc = eval(cfunc)
+        cfunc = something(static_eval(cfunc), cfunc)
     end
     if isa(cfunc, Symbol)
         cfunc = QuoteNode(cfunc)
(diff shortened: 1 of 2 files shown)


maleadt:
Hmm, this doesn't seem entirely right. Apparently we encounter the following: 
```
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: QuoteNode
      value: tuple (function of type typeof(tuple))
    2: QuoteNode
      value: Symbol pcre2_match_data_create_from_pattern_8
    3: QuoteNode
      value: String "libpcre2-8"
```
Which looks an awful lot like `(:pcre2_match_data_create_from_pattern_8, "libpcre2-8")` but it's not clear to me how it ended up like that.

Changing the pattern to a try/catch to mimic `static_eval` seems like a better approach.

KristofferC:
> Changing the pattern to a try/catch to mimic static_eval seems like a better approach.

Agreed.

timholy:
> Which looks an awful lot like (:pcre2_match_data_create_from_pattern_8, "libpcre2-8") but it's not clear to me how it ended up like that.

For performance reasons JuliaInterpreter looks up `GlobalRef`s and performs other optimizations of lowered code: https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/910d1706cefc63230339cdab0655fe1392daa393/src/types.jl#L87-L92

maleadt:
Good to go, or anything else that should be added here?

PR #389, opened 2020-05-01
Fix on Julia nightly

timholy:
There are a couple of changes to lowering that recently crept in. There may be more coming, since Revise's signature tests are completely busted and I haven't yet figured out which package needs to be fixed.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -13,7 +13,7 @@ makedocs(
     strict = true,
     format = Documenter.HTML(prettyurls = get(ENV, "CI", nothing) == "true"),
     sitename = "JuliaInterpreter.jl",
-    authors = "Keno Fischer, Tim Holy, and others",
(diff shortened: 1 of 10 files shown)


docs/make.jl
```diff
@@ -13,7 +13,7 @@ makedocs(
     strict = true,
     format = Documenter.HTML(prettyurls = get(ENV, "CI", nothing) == "true"),
     sitename = "JuliaInterpreter.jl",
-    authors = "Keno Fischer, Tim Holy, and others",
+    authors = "Keno Fischer, Tim Holy, Kristoffer Carlsson, and others",
```
KristofferC, line 16: Woho :tada:

timholy: TBH I don't even know if this field does anything. But clearly we haven't maintained the docs in a long time!

timholy:
There will be more changes in LoweredCodeUtils & Revise, but it seems these were all we needed here.

PR #391, opened 2020-05-18
Reduce invalidations due to untyped fields of CodeInfo

timholy:
xref https://github.com/JuliaLang/julia/issues/35922


timholy:
Wow, lots of failures, but none seem related.

KristofferC:
The failure seems to just be download problems? I know there was some DNS issues with julialang recently.

timholy:
Right. If these seem OK to you I think we can merge.

PR #394, opened 2020-05-22
Utility for undoing more of the optimization

timholy:
Improves printing if you want to display `frame.framecode.src`.


timholy:
This should be pretty uncontroversial so I will merge.

PR #395, opened 2020-05-22
add ability to evaluate ssavalues and locals

KristofferC:
```
1|debug>
In sin(x) at special/trig.jl:30
  3  │          Core.NewvarNode(:(n))
  4  │          absx = (abs)(x)
  5  │    %5  = absx
  6  │    %6  = ($(Expr(:static_parameter, 1)))(π)
> 7  │    %7  = (/)(%6, 4)
  8  │    %8  = (<)(%5, %7)
  9  └───       goto #6 if not %8
 10  2 ── %10 = absx
 11  │    %11 = (eps)($(Expr(:static_parameter, 1)))

About to run: (/)(3.141592653589793, 4)
1|julia> var"%5", var"%6"
(2.0, 3.141592653589793)
```

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -493,6 +493,9 @@ julia> JuliaInterpreter.locals(frame)
  #self# = capture
  x = Core.Box(2)
 ```
+
+"Special" values like SSA values and slots (shown in lowered code as e.g. `%3` and `@_4`
+respectively) can be evaluated using the syntax `var"%3"` and `var"@_4"` respectively.
 """
 function eval_code end
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -508,9 +508,13 @@ function eval_code(frame::Frame, expr)
     end
     # see https://github.com/JuliaLang/julia/issues/31255 for the Symbol("") check
     vars = filter(v -> v.name != Symbol(""), locals(frame))
+    defined_ssa = findall(!=(0), [isassigned(data.ssavalues, i) for i in 1:length(data.ssavalues)])
+    defined_locals = findall(x -> x isa Some, data.locals)
     res = gensym()
     eval_expr = Expr(:let,
-        Expr(:block, map(x->Expr(:(=), x...), [(v.name, maybe_quote(v.value isa Core.Box ? v.value.contents : v.value)) for v in vars])...),
+                     Expr(:block, map(x->Expr(:(=), x...), [(v.name, maybe_quote(v.value isa Core.Box ? v.value.contents : v.value)) for v in vars])...,
+                     map(x->Expr(:(=), x...), [(Symbol("%$i"), data.ssavalues[i]) for i in defined_ssa])...,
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -508,9 +508,13 @@ function eval_code(frame::Frame, expr)
...
     # see https://github.com/JuliaLang/julia/issues/31255 for the Symbol("") check
     vars = filter(v -> v.name != Symbol(""), locals(frame))
+    defined_ssa = findall(!=(0), [isassigned(data.ssavalues, i) for i in 1:length(data.ssavalues)])
+    defined_locals = findall(x -> x isa Some, data.locals)
     res = gensym()
     eval_expr = Expr(:let,
-        Expr(:block, map(x->Expr(:(=), x...), [(v.name, maybe_quote(v.value isa Core.Box ? v.value.contents : v.value)) for v in vars])...),
+                     Expr(:block, map(x->Expr(:(=), x...), [(v.name, maybe_quote(v.value isa Core.Box ? v.value.contents : v.value)) for v in vars])...,
+                     map(x->Expr(:(=), x...), [(Symbol("%$i"), data.ssavalues[i]) for i in defined_ssa])...,
+                     map(x->Expr(:(=), x...), [(Symbol("@_$i"), data.locals[i].value) for i in defined_locals])...),
```
timholy, line 517: IIUC this makes slots available by their number but not their name, right?

```julia
julia> function summer(A)
           s = zero(eltype(A))
           for i in eachindex(A)
               s += A[i]
           end
           return s
       end
summer (generic function with 1 method)

julia> src = Base.uncompressed_ast(@which summer([1,2,3]))
CodeInfo(
    @ REPL[10]:2 within `summer'
1 ─ %1  = Main.eltype(A)
│         s = Main.zero(%1)
│   @ REPL[10]:3 within `summer'
│   %3  = Main.eachindex(A)
│         @_4 = Base.iterate(%3)
│   %5  = @_4 === nothing
│   %6  = Base.not_int(%5)
└──       goto #4 if not %6
2 ┄ %8  = @_4
│         i = Core.getfield(%8, 1)
│   %10 = Core.getfield(%8, 2)
│   @ REPL[10]:4 within `summer'
│   %11 = s
│   %12 = Base.getindex(A, i)
│         s = %11 + %12
│         @_4 = Base.iterate(%3, %10)
│   %15 = @_4 === nothing
│   %16 = Base.not_int(%15)
└──       goto #4 if not %16
3 ─       goto #2
    @ REPL[10]:6 within `summer'
4 ┄       return s
)

julia> src.slotnames
5-element Array{Symbol,1}:
 Symbol("#self#")
 :A
 :s
 Symbol("")
 :i
```

KristofferC: Yes, for their name you just evaluate with the local variable name, right? I was a bit unsure if I should include this but it seems easy enough and when you see e.g. `@_4 = Base.iterate(%3, %10)` in the lowered code, perhaps you just want to look at what `@_4` contains.

timholy: Right, duh, you're saying this gets covered the other branch. Carry on :smile:

PR #396, opened 2020-05-27
isdefined check in locals()

pfitzseb:
~~Not sure how to best test this though.~~ Now with tests.

### DIFF
--- a/test/interpret.jl
+++ b/test/interpret.jl
@@ -642,10 +642,10 @@ end
 end
 
 @testset "constructor locals" begin
-    struct A
+    struct A396
         a::Int
     end
 
-    frame = JuliaInterpreter.enter_call(A, 3)
+    frame = JuliaInterpreter.enter_call(A396, 3)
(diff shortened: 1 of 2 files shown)


timholy:
Presumably this fixed an error, do you have a reproducer?

pfitzseb:
Yeah. Added a test...

test/interpret.jl
```diff
@@ -640,3 +640,12 @@ end
     D = Diagonal([1.0, 2.0])
     @test @interpret(f(D)) === f(D)
 end
+
+@testset "constructor locals" begin
+    struct A
+        a::Int
+    end
+
+    frame = JuliaInterpreter.enter_call(A, 3)
```
timholy, line 649:
```suggestion
    struct A396
        a::Int
    end

    frame = JuliaInterpreter.enter_call(A396, 3)
```

timholy:
Looks like you've got a Julia 1.0-specific bug, otherwise this is good to go. (If you edit it note I added a commit that fixed a different problem.)

PR #401, opened 2020-06-03
Move code around for easier VS Code integration

davidanthoff:
These changes would allow us to ship Revise.jl out of the box with the VS Code extension and run it integrated with the debugger in the Julia REPL of the extension.

All this does is move all the code from JuliaInterpreter.jl into core.jl, except for any `using` statements that load non stdlib packages.

See also https://github.com/julia-vscode/julia-vscode/pull/1283.

CC @timholy


timholy:
I approve, with the following caveats:
- To me, the name `core.jl` suggests the "inner guts" rather than the entire package definition. Maybe call it `packagedef.jl` or something? 
- Might want to add a comment in JuliaInterpreter.jl explaining this unusual organization

davidanthoff:
Yes, agreed on both suggestions, I made those changes!

Merging this PR now would actually help quite a bit already, even if we delay the Revise.jl integration, because it would make the debugging code in VS Code simpler.

davidanthoff:
@timholy do you think you could merge this PR now? It would significantly simplify the code loading story in the VS Code extension for the debugger now (i.e. independently of whether we are going to ship Revise.jl or not).

timholy:
I support the goal but it is an ugly hack. On balance I think we should merge it if @KristofferC is OK with it---it is only a single file indirection, after all.

davidanthoff:
Yes, it is ugly, but it would replace an even uglier hack [here](https://github.com/julia-vscode/julia-vscode/blob/master/scripts/debugger/debugger.jl#L10) :) And the net benefits on the VS Code side of this approach are huge: no package versioning conflicts due to the debugger loaded in the REPL process, ever :)

It would be great if there was a less hacky approach for this at some point, but given that we always want to support the latest LTS Julia release in VS Code, even if some future Julia version offered a more elegant solution for this scenario, we would probably have to stick with a hack for a couple of years, until the elegant solution has made its way into a LTS release.

davidanthoff:
Bump, I think @timholy is waiting for @KristofferC to indicate whether he is ok with this?

KristofferC:
Let's go with this for now then but let's keep thinking about a better way to do it :)

PR #402, opened 2020-06-04
Use gensyms when extracting variables

timholy:
The `local` mechanism only works inside `macro`
Fixes

```julia
julia> args = (1,)
(1,)

julia> kwargs = ()
()

julia> @interpret complicated_keyword_stuff(args...; kwargs...)
ERROR: UndefVarError: kwargs not defined
Stacktrace:
 [1] top-level scope at /home/tim/.julia/dev/JuliaInterpreter/src/construct.jl:624
```

whereas the call works if you name `args` and `kwargs` something different.


timholy:
Any objections?

PR #403, opened 2020-06-04
Detect splatting kwarg prep

timholy:
We were failing to detect kw functions that accepted vararg positional args.


timholy:
Any objections?

KristofferC:
Didn't look too carefully but the test shows it works so LGTM. It is a bit unfortunate that we need to have so much complicated code to match but it is what it is. Perhaps in the future we could think about if the compiler could emit som helpful markers in the lowered code or something to help with stuff like this.

timholy:
I think the ultimate answer is https://github.com/JuliaLang/julia/issues/31162, but perhaps in its absence we could come up with something else?

PR #406, opened 2020-06-07
Support incremental evaluation in `split_expressions`

timholy:
https://github.com/timholy/Revise.jl/issues/475 involves a couple of issues. With https://github.com/timholy/Revise.jl/pull/497 on the way, the big obstacle is how Revise organizes its information: it first parses a file, splits the expressions out into modules, and then inserts them into a Dict as keys. The actual evaluation from `includet` (currently) iterates over the Dict keys.

The problems with this design are that (1) when an expression is repeated exactly, you still only get one key/value pair in the Dict, and (2) order information is lost (other than the fact that these are SortedDicts) in terms of the order in which expressions in different modules should be executed. In a way it's kind of amazing that the old system worked as well as it did.

To fix this properly, it seems that the best strategy is to evaluate as you go, mimicking `include`. To me it seemed that the best place to put this is `split_expressions`, so now there's a new keyword `eval::Bool`. I also took the occasion to clean up how `:toplevel` expressions are handled and be more expansive in saving LineNumberNodes, thinking that given issues like https://github.com/timholy/Revise.jl/issues/437 and similar that we'd rather not risk losing location information.

Technically, changing the output of `split_expressions!` this way is a breaking change, so perhaps we should go to version 0.8. How do folks feel about that? There's always a certain amount of Pkg-churn when we make incompatible releases, but for this package I don't think it's too bad.

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -44,7 +44,7 @@ updated _after_ the update is applied.
 !!! warning
     This feature is experimental, and may be modified or removed in a minor release.
 """
-function firehooks(hooked_fun, bp::AbstractBreakpoint)
+function firehooks(@nospecialize(hooked_fun), bp::AbstractBreakpoint)
     for hook in breakpoint_update_hooks
         try
             hook(hooked_fun, bp)
@@ -111,7 +111,7 @@ end
 breakpoint(radius2, Tuple{Int,Int}, :(y > x))
 ```
 """
-function breakpoint(f::Union{Method, Function}, sig=nothing, line::Integer=0, condition::Condition=nothing)
+function breakpoint(@nospecialize(f::Union{Method, Function}), sig=nothing, line::Integer=0, condition::Condition=nothing)
     if sig !== nothing && f isa Function
         sig = Base.to_tuple_type(sig)
         sig = Tuple{typeof(f), sig.parameters...}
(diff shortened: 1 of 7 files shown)


timholy:
Fascinating. Turns a `@test_broken` into a `@test`, though this was not my intention.

KristofferC:
> Technically, changing the output of split_expressions! this way is a breaking change, so perhaps we should go to version 0.8. How do folks feel about that? 

It looks like the only packages using it are Atom and Revise. So unless it breaks existing versions of Atom / Revise I would personally just make it a patch release.

timholy:
I hadn't checked explicitly (good idea, thanks), but sure enough it's badly breaking for Revise.

timholy:
I'm more interested in doing this right than rushing it, so I plan to leave this open until @KristofferC has had a moment to catch his breath and think about it. (I am sure your task list is long and includes big-ticket items like, "get Julia 1.5 out" which I strongly support.)

When there's time to think about it, here are a few more thoughts:

- there *are* other ways Revise could fix https://github.com/timholy/Revise.jl/issues/475. For instance, `includet` could `include` the file and then parse it for signatures retrospectively like it does with packages. That involves parsing the file twice but that's not typically the dominant contribution to the total time. If we really dislike the idea of going to 0.8, this is probably the best option.
- I think the strongest argument to change things is probably the view that https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/480983e1edc96e67912308ab2fe4ffdfe89bf988/src/construct.jl#L472-L479 is a pretty bad hack. It's designed so that https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/480983e1edc96e67912308ab2fe4ffdfe89bf988/src/construct.jl#L365 won't throw an error, but I'm not sure it's comprehensive or complete. The new `must_be_toplevel` here is an attempt to do better, but while at it I decided to try to have `split_expressions` itself not have to make such decisions, and just preserve whatever setting the original expression has.
- there are quite a few other things here; the controversial commit is ecca5e9. If it's easier to review, I'd be happy to split the rest out into a separate PR.

timholy:
I'm reconsidering this (and it may take a while), so for the time being don't merge.

PR #411, opened 2020-06-20
Add an "egal" variant of `is_quotenode`

timholy:
To prevent invalidations one wants to avoid `==` and use `===` in circumstances where types can't be inferred.

I should have thought of this earlier, as it would have saved some uglification.


timholy:
Failure on master fixed by #412

KristofferC:
Is there a reason `is_quotenode` cannot just be removed or changed to the definition of `is_quotenode_egal`?

timholy:
Fair point, there may not be any users. Nothing prevents one from stashing, say, `[1 2; 3 4]` in an `QuoteNode` but it doesn't seem important in practice.

PR #412, opened 2020-06-20
Support ReturnNode and GotoIfNot

timholy:
These new IR types were added in https://github.com/JuliaLang/julia/pull/36318. Also fixes a couple of other issues noticed along the way.


timholy:
Since this passes tests and there's an open bug at Revise (https://github.com/timholy/Revise.jl/issues/509) I'm going to merge this now, and we can tweak later if this has problems.

PR #419, opened 2020-08-09
fixes maybe problematic condition check of `prepare_thunk`

aviatesk:
I just looked at the code base and found this condition check is maybe problematic -- we should "escape" (i.e. "just evaluate") if `thunk` is not `isexpr(thunk, :thunk)`.
I don't have a MRE which causes actual problem by this though (and so I didn't add test on this)

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -360,8 +360,8 @@ function prepare_thunk(mod::Module, thunk::Expr, recursive::Bool=false; eval::Bo
         error("lowering returned an error, ", thunk)
     elseif recursive
         thunk = Meta.lower(mod, thunk)
-        if isa(thunk, Expr)
-            # If on 2nd attempt to lower it's still an Expr, just evaluate it
+        if !isexpr(thunk, :thunk)
+            # If on 2nd attempt to lower it's still an non-thunk Expr, just evaluate it
             eval && Core.eval(mod, thunk)
             return nothing
         end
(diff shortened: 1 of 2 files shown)


timholy:
```julia
julia> ex = Expr(:quote, true)
:($(Expr(:quote, true)))

julia> Meta.lower(Main, ex)
true
```

aviatesk:
Well actually the case will be handled by this condition check:
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/e42045c39f6363aa5035da5c320587b5264ca644/src/construct.jl#L371
and so `prepare_thunk(Main, Expr(:quote, true))` just returns `nothing` as is:
```julia
julia> prepare_thunk(Main, Expr(:quote, true))
```

We can still cause the error by manually specifying `recursive = true` though:
```julia
julia> prepare_thunk(Main, Expr(:quote, true), true) # error on the current master
ERROR: type Bool has no field args
Stacktrace:
 [1] getproperty(x::Bool, f::Symbol)
   @ Base ./Base.jl:33
 [2] prepare_thunk(mod::Module, thunk::Expr, recursive::Bool; eval::Bool)
   @ JuliaInterpreter ~/julia/packages/JuliaInterpreter/src/construct.jl:368
 [3] prepare_thunk(mod::Module, thunk::Expr, recursive::Bool)
   @ JuliaInterpreter ~/julia/packages/JuliaInterpreter/src/construct.jl:357
 [4] top-level scope
   @ none:1
```

Do we want to add a test for this case ?

timholy:
I think it would make sense to have a case where the current implementation is known to fail. By my eye the `recursive` input seems to handle all the issues that you're worried about? That is to say, I don't really understand what motivates this change.

aviatesk:
Okay, I added the test case.

> By my eye the recursive input seems to handle all the issues that you're worried about? 

I think so (as far as it's used by JuliaInterpreter). I still believe the change in this PR is good to be merged -- it will make the function more robust.

timholy:
Another option is to just get rid of the recursive trick altogether, if you're worried about folks calling it with `recursive=true` from the outset. Would you support that option?

PR #420, opened 2020-08-12
Use `mapany` to avoid some broadcasting invalidations

timholy:
These were triggered by loading FillArrays


timholy:
Test failure seems unrelated

PR #425, opened 2020-09-04
Add ExprSplitter

timholy:
This implements an "expression-splitting iterator" that returns individual expressions ready for evaluation to the caller. This allows the caller to decide what to do with them (e.g., evaluation, signature analysis, whatever). It's basically a proposed replacement for `prepare_thunk` and `split_expressions`; by placing the consumer in charge, we get rid of any of the weird "should I eval if I can't lower?" questions that have dogged the previous framework.

~~For the moment this is parallel to our older framework; I've not yet actually rebased JuliaInterpreter on this, although the `Frame(::Module, ::CodeInfo)` constructor is a step in this direction.~~ My local copy of `Revise#revise3` is now based on this and it seems to work well.

@pfitzseb and @aviatesk, are there callers of these functions outside JuliaInterpreter, Debugger, and Revise? If so we'd want to think about a careful deprecation path. Indeed we might want to do that anyway since `split_expressions`, even though not exported, is pretty essential in some contexts (so there may be unknown users).

Closes #406
Closes #419


pfitzseb:
I'm pretty sure we're using `split_expressions` in VSCode and Juno, yes.

KristofferC:
Instead of working hard on deprecation and then just have everyone immediately update to the new one (effectively making the work on the deprecation pointless) we could also release a new breaking version and have people just update to that. Of course, this depends on how hard it is to do the deprecation and how hard it is to do the upgrade.

timholy:
Since these are officially internal (though well-used) functions, I decided to go with the breaking change. The transition is pretty straightforward as can be seen by the tests. The main thing I'm unsure of is whether we want to export a utility for detecting and stripping out `@doc` exprs, or whether that was relevant only for Revise.

test/interpret.jl
```diff
@@ -320,9 +320,9 @@ end
 f113(;x) = x
 @test @interpret(f113(;x=[1,2,3])) == f113(;x=[1,2,3])
 
-# Some expression can appear nontrivial but lower to nothing
-@test isa(JuliaInterpreter.prepare_thunk(Main, :(@static if ccall(:jl_get_UNAME, Any, ()) == :NoOS 1+1 end)), Nothing)
-@test isa(JuliaInterpreter.prepare_thunk(Main, :(Base.BaseDocs.@kw_str "using")), Nothing)
+# Some expressions can appear nontrivial but lower to nothing
+# @test isa(Frame(Main, :(@static if ccall(:jl_get_UNAME, Any, ()) == :NoOS 1+1 end)), Nothing)
+# @test isa(Frame(Main, :(Base.BaseDocs.@kw_str "using")), Nothing)
```
timholy, line 325: I wasn't quite sure what to do about these.

timholy:
Ugh, this is ugly getting it over the finish line. The various bugs that different Julia versions have wrt LineNumberNodes are really annoying.

Are the remaining errors anything to worry about? Travis had a failure on Windows 1.4 (a strange string-indexing error), appveyor on 1.2 (a problem with `stat`). The second seems more worrisome, since it seems to have happened on both 32- and 64-bit. A few older PRs were not triggering this.

src/utils.jl
```diff
@@ -129,6 +129,23 @@ _scopename(sym) = sym
...
+function must_be_toplevel(mod::Module, ex::Expr; expand::Bool=true)
+    head = ex.head
+    if head === :toplevel || head === :block
+        for a in ex.args
+            if isa(a, Expr)
+                must_be_toplevel(mod, a) && return true
+            end
+        end
+    elseif head === :export || head === :import || head === :using || head === :global || head === :const
+        return true
```
aviatesk, line 141: It feels like `must_be_toplevel` actually means `must_be_just_evaluated` or something ?
Because it seems like what should always be in toplevel is actually only `:const`, and the others can work within local scopes (like `let`), like:
```julia
julia> let 
           i = 0
           struct A
               val
               A(v = i) = new(v)
           end
           export A
       end

julia> A()
A(0)
```

aviatesk:
> Since these are officially internal (though well-used) functions, I decided to go with the breaking change. The transition is pretty straightforward as can be seen by the tests. The main thing I'm unsure of is whether we want to export a utility for detecting and stripping out @doc exprs, or whether that was relevant only for Revise.

as far as I know, the only consumer who uses the utility for `at-doc` exprs is [only Revise](https://juliahub.com/ui/CodeSearch?q=split_expressions&u=all&t=all), so I would like to say it's your call :)

src/utils.jl
```diff
@@ -129,6 +129,23 @@ _scopename(sym) = sym
...
+    elseif expand && head === :macrocall
+        exex = macroexpand(mod, ex)
```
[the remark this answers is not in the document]
timholy: *early* expansion (like in the first lines of the `ExprSplitter` constructor) triggers failures in [this test](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/64d719fd202fedd41d9b0400340708187dd38f8e/test/toplevel.jl#L249-L289), which is explicitly designed to catch flaws like that. But "last minute" expansion, in `queuenext!`, might be possible. But of course the user can do it too: if the returned expr has a `:macroexpand` then expand and re-queue. I was taking the viewpoint of "do no harm" and so didn't want to transform expressions more than necessary.

src/utils.jl
timholy: Yeah, in a way I guess it's really `doesnt_lower`.

timholy:
Thanks! If Revise is the only one, I've already got that covered from within Revise. There's isn't much to it, really, just grabbing the 4th argument of the expression so it wouldn't be much of a utiliy anyway.

KristofferC:
> Are the remaining errors anything to worry about? Travis had a failure on Windows 1.4 (a strange string-indexing error), appveyor on 1.2 (a problem with stat)

I would not really worry about unmaintained julia versions (1.1 - 1.4). Might be time to drop compat from them completely so they will keep working at the state they are in but not get updates.

timholy:
Sounds good. Looks like https://github.com/JuliaLang/julia/issues/26685. I just pushed a debug commit; let's see if that reveals an easy fix, and if not I'll just make 1.3 the lowest Julia version supported.

timholy:
That is one weird filename :smile:.https://ci.appveyor.com/project/KristofferC/juliainterpreter-jl/builds/35055810/job/ubmbpu36otugvd9e

timholy:
A JuliaHub search indicates that the only known users of `prepare_thunk` and `split_expressions` are the ones we know about. So this crew (I should have also CCed @davidanthoff) can decide.

src/construct.jl
```diff
@@ -348,152 +348,201 @@ function prepare_frame_caller(caller::Frame, framecode::FrameCode, args::Vector{
...
+        while idx <= length(ex.args)
+            a = ex.args[idx]
+            if isa(a, LineNumberNode)
+                iter.lnn = a
+            elseif isa(a, Expr)
+                iter.index[end] = idx + 1
+                push_modex!(iter, mod, a)
+                return queuenext!(iter)
+            else
+                # error("unhandled argument ", a)
```
KristofferC, line 516: Should this be uncommented or removed?

KristofferC:
> That is one weird filename

Wow, where did that come from?

src/construct.jl
timholy: It could be useful for future debugging, but I will remove it. Frequently when I develop code I first write a brittle implementation and then remove the error condition when I become convinced I've handled the necessary cases. But this will throw on many things, so there is little point any more.

KristofferC: > Frequently when I develop code I first write a brittle implementation and then remove the error condition when I become convinced I've handled the necessary cases. 

Yeah me too, but without any comment or such it might become a point of uncertainty for a future reader.

src/utils.jl
timholy: Thanks for pushing on this. It turns out that if you wrap in a `:block` together with a `LineNumberNode`, then even `export`, `import`, and `using` do lower to a `CodeInfo`. And `:const` already did, I'm not sure why I added it here.

That leaves only `:global` expressions. Does anyone know how to make them lower to a CodeInfo?

Given that it's just `:global`, I've reworked this to simply not deliver a `:block`/`:toplevel` wrapper when it's a `:global` expr.

timholy:
I am not sure. Seems likely to be a memory bug or something.

timholy:
`Revise@3.0.0` is officially ready to be announced, pending the resolution of this PR.

src/utils.jl
aviatesk: > And `:const` already did

I never know that, so sneaky.

> Given that it's just `:global`, I've reworked this to simply not deliver a `:block`/`:toplevel` wrapper when it's a `:global` expr.

Yeah, the docstring is kind enough, I think.

src/utils.jl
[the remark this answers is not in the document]
aviatesk: I see. Then let's go this way.

aviatesk (approved):
lgtm

timholy:
Seems like folks are willing to do the work of updating for the breaking change. Many thanks, all!

timholy:
Before merging I'm going to back out the commit disabling the ambiguity test. That means the test will fail, but we've at least seen the tests pass.

Issue #433, opened 2020-09-16
AssertionError: arg isa SSAValue with at-ccall


fredrikekre:
```
julia> using JuliaInterpreter

julia> function f()
           s = "test"
           @ccall strlen(s::Cstring)::Csize_t
       end
f (generic function with 1 method)

julia> @interpret f()
ERROR: AssertionError: arg isa SSAValue
Stacktrace:
 [1] build_compiled_call!(::Expr, ::Symbol, ::Core.CodeInfo, ::Int64, ::Int64, ::Array{Symbol,1}, ::Module) at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/optimize.jl:273
 [2] optimize!(::Core.CodeInfo, ::Method) at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/optimize.jl:186
 [3] JuliaInterpreter.FrameCode(::Method, ::Core.CodeInfo; generator::Bool, optimize::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/types.jl:101
 [4] prepare_framecode(::Method, ::Any; enter_generated::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/construct.jl:190
 [5] prepare_call(::Any, ::Array{Any,1}; enter_generated::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/construct.jl:261
 [6] determine_method_for_expr(::Expr; enter_generated::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/construct.jl:582
 [7] enter_call_expr(::Expr; enter_generated::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/construct.jl:623
 [8] enter_call_expr(::Expr) at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/construct.jl:622
 [9] top-level scope at /home/fredrik/.julia/packages/JuliaInterpreter/muncv/src/construct.jl:756
```

Note that `ccall(:strlen, Csize_t, (Cstring,), s)` works.

JuliaInterpreter v0.8.1

KristofferC:
I guess this is on master with the new non-const ccall?

fredrikekre:
No this was with Julia 1.5.1.

KristofferC:
Oh, nvm, I got confused with the new `@ccall` syntax.

PR #434, opened 2020-09-16
fix at-ccall?

pfitzseb:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/433.

I feel like this is probably not the correct fix, but it works, so... ¯\\\_(ツ)\_/¯


timholy:
I've not followed `@ccall`, but note the test failures that prevent this from being merged.

pfitzseb:
So I think this actually works now, instead of just accidentally for the one test function I had.

pfitzseb:
Remaining test errors are unrelated.

timholy:
Thanks!

PR #440, opened 2020-11-19
check if symbol is defined before trying to access it

pfitzseb:
Fixes https://discourse.julialang.org/t/debugger-stops-without-hitting-a-breakpoint/50378.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -47,7 +47,7 @@ macro lookup(args...)
     nodetmp = gensym(:node)  # used to hoist, e.g., args[4]
     if havemod
         fallback = quote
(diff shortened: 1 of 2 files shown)


aviatesk (approved):
Why aren't these symbols wrapped into `QuoteNode`s ... ?

timholy (dismissed):

timholy:
Actually, this is more subtle than I initially thought. This essentially kicks the can down the road; suppose there is an object named `bar` but someone also uses `:bar` as an argument in the method. Is that supposed to refer to the object or is the symbol to be taken literally? In this case, the fix is to see if there is an object by that name, if so return it, otherwise assume that the symbol is to be taken literally.

I almost wonder if we should delete those Symbol lookup methods (that breaks a lot of tests, but maybe they can all be addressed individually).

PR #443, opened 2020-11-21
Respect namespace in method definitions

timholy:
The symbol-lookup in `evaluate_methoddef` was not careful about
avoiding extending methods in Base when the method was unqualified.

Fixes https://github.com/timholy/Revise.jl/issues/579

-------------------

Unfortunately, this doesn't currently work:
```julia
julia> using JuliaInterpreter, Test

julia> module Namespace end
Main.Namespace

julia> @testset "Namespace" begin
           frame = Frame(Namespace, :(sin(::Int) = 10))
           while true
               JuliaInterpreter.through_methoddef_or_done!(frame) === nothing && break
           end
           @test Namespace.sin(0) == 10
           @test Base.sin(0) == 0
       end
f = :sin
typeof(f) = Symbol
moduleof(frame) = Main.Namespace
Namespace: Error During Test at REPL[3]:1
  Got exception outside of a @test
  error in method definition: function Base.sin must be explicitly imported to be extended
  Stacktrace:
    [1] top-level scope
      @ none:0
    [2] top-level scope
      @ none:1
    [3] eval
      @ ./boot.jl:360 [inlined]
    [4] evaluate_methoddef(frame::Frame, node::Expr)
      @ JuliaInterpreter ~/.julia/dev/JuliaInterpreter/src/interpret.jl:278
...
```

What I don't understand is, this does:
```julia
julia> module Namespace end
Main.Namespace

julia> module DefineSin
       createsin() = Core.eval(Main.Namespace, Expr(:function, :sin))
       defsin() = Core.eval(Main.Namespace, :(sin(::Int) = 10))
       end
Main.DefineSin

julia> DefineSin.createsin()
sin (generic function with 0 methods)

julia> DefineSin.defsin()
sin (generic function with 1 method)

julia> Base.sin(0)
0.0

julia> Namespace.sin(0)
10
```
and this basically mimics what's going on in `evaluate_methoddef`. Any ideas?

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -264,25 +264,19 @@ function evaluate_methoddef(frame, node)
     f = node.args[1]
     if isa(f, Symbol)
         mod = moduleof(frame)
-        createnew = true
-        if isdefined(mod, f)
-            flkup = getfield(mod, f)
-            fmod = typeof(flkup).name.module
-            if fmod === scopeof(frame)
-                f = flkup
-                createnew = false
-            end
-        end
-        if createnew
-            @show f typeof(f) moduleof(frame)
+        if Base.isbindingresolved(mod, f)  # `isdefined` accesses the binding, making it impossible to create a new one
+            f = getfield(mod, f)
+        else
             f = Core.eval(moduleof(frame), Expr(:function, f))  # create a new function
         end
+    elseif isa(f, GlobalRef)
+        f = getfield(f.mod, f.name)
     end
     length(node.args) == 1 && return f
     sig = @lookup(frame, node.args[2])::SimpleVector
     body = @lookup(frame, node.args[3])
     ccall(:jl_method_def, Cvoid, (Any, Any, Any), sig, body, moduleof(frame))
-    return nothing
+    return f
 end
 
 function structname(frame, node)
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -264,7 +264,7 @@ function evaluate_methoddef(frame, node)
     f = node.args[1]
     if isa(f, Symbol)
         mod = moduleof(frame)
-        if Base.isbindingresolved(mod, f)  # `isdefined` accesses the binding, making it impossible to create a new one
+        if Base.isbindingresolved(mod, f) && isdefined(mod, f)  # `isdefined` accesses the binding, making it impossible to create a new one
             f = getfield(mod, f)
         else
             f = Core.eval(moduleof(frame), Expr(:function, f))  # create a new function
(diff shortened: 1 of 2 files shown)


timholy:
The error comes from https://github.com/JuliaLang/julia/blob/c36cf8cbc1480962b8d6b110f37ac8c87090da72/src/module.c#L233

src/interpret.jl
```diff
@@ -264,7 +264,19 @@ function evaluate_methoddef(frame, node)
...
+            flkup = getfield(mod, f)
+            fmod = typeof(flkup).name.module
+            if fmod === scopeof(frame)
+                f = flkup
+                createnew = false
+            end
+        end
+        if createnew
+            @show f typeof(f) moduleof(frame)
+            f = Core.eval(moduleof(frame), Expr(:function, f))  # create a new function
```
cstjean, line 278:
```suggestion
            f = Core.eval(moduleof(frame), Expr(:function, f))  # create a new function
            @show Core.eval(moduleof(frame), f)
```

cstjean:
The only other thing that comes to mind is that maybe it's world-age related? But if you're using `eval` for both the `function $f end` and the method definition, I don't see how it would come up.

timholy:
I figured it out: the error comes only if you've previously *accessed* the binding. It's mimicked by this:
```julia
julia> module Namespace end
Main.Namespace

julia> Namespace.sin
sin (generic function with 13 methods)

julia> Core.eval(Main.Namespace, Expr(:function, :sin))
ERROR: error in method definition: function Base.sin must be explicitly imported to be extended
Stacktrace:
 [1] top-level scope
   @ none:0
 [2] top-level scope
   @ none:1
 [3] eval(m::Module, e::Any)
   @ Core ./boot.jl:360
 [4] top-level scope
   @ REPL[3]:1
```

But if you omit the `Namespace.sin`, then it works.

Unfortunately, `isdefined` accesses the binding, which is why the original version of this PR triggered the error. But fortunately `Base.isbindingresolved` does not access it, making this solvable.

It looks like this won't work properly on Julia 1.0, but it doesn't error either, so I've allowed the "noncontamination" test to fail for Julia < 1.5.

cstjean:
Interesting, do you know why 

```julia
julia> module A
       sin
       @static if isdefined(A, :sin)
           sin(::Int) = 10
       end
       end
WARNING: replacing module A.
ERROR: error in method definition: function Base.sin must be explicitly imported to be extended
Stacktrace:
```
fails then? Feels like I'm accessing the bindings.

timholy:
For further insight, try that demo with `baremodule` rather than `module`.

cstjean:
Something feels a bit off about `isdefined` storing the binding. Couldn't there be a macro like

```julia
@bunglify function time(x) ... end
```

where the macro expansion uses `isdefined(:time)` and the user unintentionally adds a Base method? My imagination is failing me at the moment.

timholy:
Dunno, but I posted the question here: https://julialang.slack.com/archives/C688QKS7Q/p1606042345009700

PR #446, opened 2020-12-11
minor code quality improvements

aviatesk:
eliminates some inferrability issues, and also removes some real errors around `GlobalRef`

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -282,7 +282,7 @@ end
 function structname(frame, node)
     name = node.args[1]
     if isa(name, GlobalRef)
-        mod = name.module
+        mod = name.mod
         name = name.name
     else
         mod = moduleof(frame)
@@ -489,7 +489,7 @@ function step_expr!(@nospecialize(recurse), frame, @nospecialize(node), istoplev
                 elseif node.head === :const
                     g = node.args[1]
                     if isa(g, GlobalRef)
-                        mod, name = g.module, g.name
+                        mod, name = g.mod, g.name
                     else
                         mod, name = moduleof(frame), g::Symbol
                     end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -299,7 +299,8 @@ function lineoffset(framecode::FrameCode)
     return offset
 end
 
-getline(ln) = Int(isexpr(ln, :line) ? ln.args[1] : ln.line)::Int
+getline(ln::LineTypes) = ln.line::Int
+getline(ln::Expr)      = ln.args[1]::Int # assuming ln.head === :line
 # work around compiler error on 1.2
 @static if v"1.2.0" <= VERSION < v"1.3"
     getfile(ln) = begin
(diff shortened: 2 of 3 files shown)


src/construct.jl
```diff
@@ -483,7 +483,8 @@ function queuenext!(iter::ExprSplitter)
             newmod isa Module || throw(ErrorException("invalid redefinition of constant $(newname)"))
             mod = newmod
         else
-            if (id = Base.identify_package(mod, String(newname))) !== nothing && haskey(Base.loaded_modules, id)
+            id = Base.identify_package(mod, String(newname))
+            if id !== nothing && haskey(Base.loaded_modules, id)
```
aviatesk, line 487: just a note for you: inference can't (back-)propagate type constraints  from `===` or `isa` in this kind of case:
```julia
if (x = y) isa T
    # some code assuming x::T
end
```

src/interpret.jl
```diff
@@ -282,7 +282,7 @@ end
 function structname(frame, node)
     name = node.args[1]
     if isa(name, GlobalRef)
-        mod = name.module
+        mod = name.mod
```
timholy, line 285: :exploding_head: 
I would almost expect a parsing error for this.

timholy (approved):
Great stuff!

src/interpret.jl
pfitzseb: Well, it is a valid fieldname (as are all keywords, I think):
```
julia> x = Symbol("module")
:module

julia> @eval struct Foo; $x; end

julia> Foo(2).module
2
```

aviatesk:
CI failures on nightlies are not related to this PR; #449 will fix them anyway.

src/utils.jl
```diff
@@ -313,7 +314,11 @@ getline(ln) = Int(isexpr(ln, :line) ? ln.args[1] : ln.line)::Int
         CodeTracking.maybe_fixup_stdlib_path(path)
     end
 else
-    getfile(ln) = CodeTracking.maybe_fixup_stdlib_path(String(isexpr(ln, :line) ? ln.args[2] : ln.file)::String)
+    function getfile(ln::Union{LineTypes,Expr})
+        _getfile(ln::LineTypes) = ln.file::Symbol
+        _getfile(ln::Expr)      = ln.args[2]::Symbol # assuming ln.head === :line
+        return CodeTracking.maybe_fixup_stdlib_path(String(_getfile(ln)))
+    end
```
aviatesk, line 321: Sorry for pushing such minor optimizations, these changes just circumvents inference limitations around https://github.com/JuliaLang/julia/issues/37342.

src/utils.jl
```diff
@@ -295,7 +295,8 @@ function lineoffset(framecode::FrameCode)
     return offset
 end
 
-getline(ln) = Int(isexpr(ln, :line) ? ln.args[1] : ln.line)::Int
+getline(ln::LineTypes) = ln.line::Int
+getline(ln::Expr)      = ln.args[1]::Int # assuming ln.head === :line
```
timholy, line 299: The original tries to convert to an `Int` and then asserts it succeeds; this change requires that they start out as `Int`. It also doesn't check that `ln` is a `:line` Expr/ Are you sure these changes are safe?

src/utils.jl
timholy: Getting rid of invalidation risk is very worthwhile, no apologies needed!

My only concern is again the loss of validation that this is really an Expr of the expected type.

src/utils.jl
aviatesk: I found the `:line` field of `LineInfoNode` and `LineInfoNode` [are `Int`](https://github.com/JuliaLang/julia/blob/9921e8c1e5934e308f18e748dfd405503bf2b34c/src/jltypes.c#L2089-L2098), but I didn't track how `:line` expression is constructed... 
I think making explicit conversion here is way more robust and doesn't hurt anything, so let me do that again.

---

As for the validation for `:line` expression, I excluded that because I believe we're actually only calling this function on `:line` expressions, and non-`:line` `Expr`s will throw anyway (since it never has `.line` field). We can do something like `getline(ln::Expr) = isexpr(ln, :line) ? ln.args[1] : throw("expected :line expression")`, but I feel it's too much ?

timholy:
Thanks!

PR #447, opened 2020-12-11
Try harder to find __toplevel__ stdlibs in ExprSplitter

timholy:
Revise has to be able to analyze code in modules that are not in
the namespace. It turns out that differences between how discovery
works in interactive mode (which implicitly widens the Manifest to
include all the stdlibs) and test-mode (which does not) have caused
trouble for the discovery of toplevel modules for some time:
it turns out we've been constructing "shadow" modules of the same names
as these stdlibs. Some of the exclusions in Revise's `sigtest.jl` test
file reflect this.

This PR makes ExprSplitter try harder to find modules from
`Base.__toplevel__`. When we can't resolve the name in the manifest,
we also check `Base.loaded_modules` and see if we can find something with
that name.

This fixes a bug that just arose (but whose roots are old) for
Revise on nightly (https://github.com/timholy/Revise.jl/pull/489/checks?check_run_id=1537104080).

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -483,7 +483,19 @@ function queuenext!(iter::ExprSplitter)
             newmod isa Module || throw(ErrorException("invalid redefinition of constant $(newname)"))
             mod = newmod
         else
-            if (id = Base.identify_package(mod, String(newname))) !== nothing && haskey(Base.loaded_modules, id)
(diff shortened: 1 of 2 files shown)


timholy:
There have not been any objections, so let's do this.

PR #437, opened 2020-10-29
Adjust to upcoming Base Vararg change

Keno:

PR #457, opened 2020-12-17
fix failures on nightly 2

simeonschaub:
commits taken from #437 and #449, so this should probably not be squashed. Just fixes the tests and addresses some warnings on top of those PRs.

closes #437
closes #449

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -248,7 +248,7 @@ end
 is_leaf(frame::Frame) = frame.callee === nothing
 
 function is_vararg_type(x)
-    if isa(Vararg, Type)
+    @static if isa(Vararg, Type)
         if isa(x, Type)
             (x <: Vararg && !(x <: Union{})) && return true
             if isa(x, UnionAll)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -248,12 +248,16 @@ end
 is_leaf(frame::Frame) = frame.callee === nothing
 
 function is_vararg_type(x)
-    if isa(x, Type)
-        (x <: Vararg && !(x <: Union{})) && return true
-        if isa(x, UnionAll)
-            x = Base.unwrap_unionall(x)
+    if isa(Vararg, Type)
+        if isa(x, Type)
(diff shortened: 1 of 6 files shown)


timholy:
Since you've done this, let's do it this way! Thanks again for getting this going.

simeonschaub:
I can also just rebase this after those two PRs are merged, if you'd prefer that

timholy:
Just extra work. No need for that.

timholy:
Is the email you made these commits with not registered with your GitHub account?

simeonschaub:
> Is the email you made these commits with not registered with your GitHub account?

Yes, those were made on a different computer, and apparently, I haven't set up git properly yet there.

src/utils.jl
```diff
@@ -248,12 +248,16 @@ end
 is_leaf(frame::Frame) = frame.callee === nothing
 
 function is_vararg_type(x)
-    if isa(x, Type)
-        (x <: Vararg && !(x <: Union{})) && return true
-        if isa(x, UnionAll)
-            x = Base.unwrap_unionall(x)
+    if isa(Vararg, Type)
```
aviatesk, line 251:
```suggestion
    @static if isa(Vararg, Type)
```

simeonschaub: Any functional reason to prefer `@static` here, or is it mostly stylistic?

aviatesk: yeah, it also may make it clearer that this branching because of version compatibility when we read this code later.

simeonschaub:
Any clue what's going on with the Windows failure? Looks unrelated to this PR though.

pfitzseb:
Looks like that `expanduser` related path issue that Tim tracked down, I think.

timholy:
I'm good with this. Any objections if we merge? (Non-squash.)

aviatesk:
I'm good too, I would appreciate it if we could have a new version that works on the current master soonish :p

timholy:
For your Ctrl-F5 pleasure: https://github.com/JuliaRegistries/General/pull/26534

PR #461, opened 2020-12-19
Minor inference and specialization improvements

timholy:


src/utils.jl
```diff
@@ -123,9 +123,12 @@ end
...
 function scopename(tn::TypeName)
     modpath = Base.fullname(tn.module)
-    return Expr(:., _scopename(modpath...), QuoteNode(tn.name))
+    if isa(modpath, Tuple{Symbol})
+        return Expr(:., modpath[1], QuoteNode(tn.name))
+    else
+        return Expr(:., _scopename(modpath...), QuoteNode(tn.name))
+    end
 end
-_scopename(sym) = sym
```
aviatesk, line 128: Just to check my understanding, does this change reduce a chance of (unused) specialization, right ? (maybe otherwise `_scopename(parent, child, rest...) = Expr(:., parent, _scopename(child, rest...))` can recur back into this method ?)

aviatesk (approved):

src/utils.jl
timholy: It was more that attempts to call `_scopename(sym)` were being implemented by runtime-dispatch. Now that I think about it, that would also be true of longer module paths: the length of `modpath` is not inferrable. I should rewrite this so it never dispatches to anything. (later, busy with other things now)

timholy:
Woot, with these changes here and a couple in LoweredCodeUtils & Revise, I've gotten the time for first revision down to 1.6s, about half of where it was a week ago. (xref https://github.com/JuliaLang/julia/pull/38906) A lot of this is just me playing with the new SnoopCompile tools and giving a workout, wanting to make sure that they actually work to discover things that help latency. But certainly nice to have on their own.

PR #463, opened 2020-12-20
Push a few ccalling methods onto compiled_methods

### LINKED ISSUE #462: Precache a few compiledcalls?
`Revise.revise()` needs at least the following (just a dump from me monitoring activity during the first revision):
```julia
scope = nameof(m::Module) in Base at reflection.jl:16
sparams = Symbol[]
methname = Symbol("##compiledcall#245")
stmt = :($(Expr(:foreigncall, :(:jl_module_name), Ref{Symbol}, svec(Any), 0, :(:ccall), _J2)))
  0.180130 seconds (265.73 k allocations: 15.778 MiB, 3.75% gc time, 99.12% compilation time)

scope = binding_module(m::Module, s::Symbol) in Base at reflection.jl:105
sparams = Symbol[]
methname = Symbol("##compiledcall#246")
stmt = :($(Expr(:foreigncall, :(:jl_get_module_of_binding), Ptr{Nothing}, svec(Any, Any), 0, :(:ccall), _J2, _J3)))
  0.001159 seconds (349 allocations: 42.334 KiB)

scope = unsafe_pointer_to_objref(x::Ptr) in Base at pointer.jl:130
sparams = Symbol[]
methname = Symbol("##compiledcall#247")
stmt = :($(Expr(:foreigncall, :(:jl_value_ptr), Any, svec(Ptr{Nothing}), 0, :(:ccall), %J4, %J2)))
  0.004961 seconds (2.32 k allocations: 188.361 KiB, 74.81% compilation time)

scope = Vector{T}(::UndefInitializer, m::Int64) where T in Core at boot.jl:448
sparams = [:T]
methname = Symbol("##compiledcall#248")
stmt = :($(Expr(:foreigncall, :(:jl_alloc_array_1d), Array{T, 1}, svec(Any, Int64), 0, :(:ccall), %J2, %J3, %J1)))
  0.013158 seconds (38.09 k allocations: 2.310 MiB, 92.92% compilation time)

scope = fill!(a::Union{Array{Int8, N} where N, Array{UInt8, N} where N}, x::Integer) in Base at array.jl:405
sparams = Symbol[]
methname = Symbol("##compiledcall#249")
stmt = :($(Expr(:foreigncall, :(:memset), Ptr{Nothing}, svec(Ptr{Nothing}, Int32, UInt64), 0, :(:ccall), %J9, %J10, %J11, %J7, %J5, %J2)))
  0.001593 seconds (422 allocations: 68.047 KiB)
```

Note the first one takes ~0.18s, so there could be considerable savings to having this. But the cache-checking function would have to delay compilation for the longer, real `build_compiled_call!`, so when lookup of a precached method failed it would have to call it via an `invokelatest`. Not sure if it's worth it, but I wanted to record the results somewhere while I had them.

timholy:
These methods trigger compilation of `build_compiled_call!` during
Revise.revise(). Since they are essentially `ccall`-wrappers already,
there's not a lot of value in interpreting them.

Saves another ~100ms off the latency.

Closes #462

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -182,14 +182,16 @@ function optimize!(code::CodeInfo, scope)
                 arg1 = stmt.args[1]
                 if (arg1 === :llvmcall || lookup_stmt(code.code, arg1) === Base.llvmcall) && isempty(sparams) && scope isa Method
                     nargs = length(stmt.args)-4
-                    delete_idx = build_compiled_call!(stmt, Base.llvmcall, code, idx, nargs, sparams, evalmod)
+                    # Call via `invokelatest` to avoid compiling it until we need it
+                    delete_idx = Base.invokelatest(build_compiled_call!, stmt, Base.llvmcall, code, idx, nargs, sparams, evalmod)
                     delete_idx === nothing && error("llvmcall must be compiled, but exited early from build_compiled_call!")
                     push!(foreigncalls_idx, idx)
                     append!(delete_idxs, delete_idx)
                 end
             elseif stmt.head === :foreigncall && scope isa Method
                 nargs = foreigncall_version == 0 ? stmt.args[5]::Int : length(stmt.args[3]::SimpleVector)
-                delete_idx = build_compiled_call!(stmt, :ccall, code, idx, nargs, sparams, evalmod)
+                # Call via `invokelatest` to avoid compiling it until we need it
+                delete_idx = Base.invokelatest(build_compiled_call!, stmt, :ccall, code, idx, nargs, sparams, evalmod)
                 if delete_idx !== nothing
                     push!(foreigncalls_idx, idx)
                     append!(delete_idxs, delete_idx)
--- a/src/packagedef.jl
+++ b/src/packagedef.jl
@@ -106,6 +106,13 @@ function set_compiled_methods()
         end
     end
 
+    # This is about performance, not safety (issue #462)
+    push!(compiled_methods, which(nameof, (Module,)))
+    push!(compiled_methods, which(Base.binding_module, (Module, Symbol)))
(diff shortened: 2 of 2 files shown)


timholy:
This is a much easier way of doing what I was proposing in #462.

timholy:
Ooh, interesting; the second commit here shaves another 100ms off! Does that seem OK?

timholy:
One concern is whether this could lead to world age problems. Might be worth using `invoke_in_world` and doing this only on Julia 1.6+.

KristofferC:
Could Revise itself could push these methods on load time so at least when you aren't using Revise, you can step into them?

timholy:
Definitely. But, if you `edit` all those methods, you'll see they are pretty much one-liners already. Does that change your thinking in any way?

KristofferC:
I guess. If they just directly dispatch to the ccall, it's probably not of huge value (except if you are actually interested in how stuff like that are implemented). If it is just these it is probably fine but if the list gets too big then maybe there should be some options to give to the interpreter to chose between sort of a "latency mode" and an "accuracy mode"?

aviatesk:
Maybe we can move forward with some TODO comments about configurable interpretation of these methods ?
I think, for now, there's not much value in interpreting those methods, and we can hold off the discussion until we get an issue complaining this.

KristofferC:
Sounds fine.

pfitzseb:
Oh, whoops. Merged this on accident because I had the wrong tab focused. Sorry for that!

Do we want to revert or keep this in for now and see whether anything breaks?

KristofferC:
Might as well keep it.

timholy:
Sorry, I set you up for this by leaving it in indeterminate state. I was on the fence, but now we can see how we like it.

PR #467, opened 2021-02-20
WIP: handle toplevel anonymous functions

aviatesk:
While debugging https://github.com/aviatesk/JET.jl/issues/104, I found
the same could happen for JuliaInterpreter.

This issue doesn't happen when interpreting an local frame, so ideally
we want to use `Base.invokelatest` only when interpreting a toplevel
frame. But it needs some refactoring on the whole interpret.jl code
base, and so I want to leave it as a future work if you're okay.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -194,7 +194,8 @@ function evaluate_call_compiled!(::Compiled, frame::Frame, call_expr::Expr; ente
     fargs = collect_args(frame, call_expr)
     f = fargs[1]
     popfirst!(fargs)  # now it's really just `args`
-    return f(fargs...)
+    # TODO `invokelatest` is only needed for toplevel functions, don't use it when interpreting local code
+    return Base.invokelatest(f, fargs...)
 end
 
 function evaluate_call_recurse!(@nospecialize(recurse), frame::Frame, call_expr::Expr; enter_generated::Bool=false)
--- a/test/debug.jl
+++ b/test/debug.jl
@@ -509,4 +509,12 @@ end
         frame, pc = JuliaInterpreter.debug_command(frame, :n)
         @test pc isa BreakpointRef
     end
+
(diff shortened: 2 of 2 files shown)


timholy:
In general we've tried to get people to split their expressions before constructing frames. https://juliadebug.github.io/JuliaInterpreter.jl/stable/internals/#Toplevel-code-and-world-age-1. But if you think this is necessary we might do it anyway. Since `invokelatest` has performance overhead I'd prefer to avoid it unless necessary, though.

aviatesk:
> In general we've tried to get people to split their expressions before constructing frames.

Ah, I forgot it !

But still some complicated cases are not gracefully handled by `ExprSplitter` and can cause world age issue...

```julia
es = ExprSplitter(@__MODULE__, quote
    using Test
    @testset begin
        a = @inferred(ones(Int,ntuple(d->1,1)))
    end
end)
for (m, x) in es
    frame = Frame(m, x)
    debug_command(Compiled(), frame, :c, true) => world age ...
end
```

I've not debugged in detail where `ExprSpliter` fails to split the definitions yet though. 
Let me debug more for now.

aviatesk:
MRE is reduced to:
```julia
es = ExprSplitter(@__MODULE__, quote
    using Test
    @inferred(ones(Int,ntuple(d->1,1)))
end);
for (m, x) in es
    frame = Frame(m, x)
    debug_command(Compiled(), frame, :c, true) # => world age
end
```

So we need to find a way to handle toplevel anonymous functions in general ?

aviatesk:
Sorry, so it's just that I didn't understand correctly what `through_methoddef_or_done` does and how JuliaInterpreter handles these world ages in general.
Sorry for the noise.

timholy:
No worries, it was no trouble and it's a pretty subtle issue. It took a long time to wrap my head around and I still forget sometimes.

PR #468, opened 2021-02-24
Interpreted methods

pfitzseb:
This adds `JuliaInterpreter.interpreted_methods` as an easy way to compile all methods in a module except for a few of them (this takes precedence over both `compiled_methods` and `compiled_modules`)

Makes the implementation of [this proposal](https://discourse.julialang.org/t/debugging-in-juno-extremely-slow/53801/10?u=pfitzseb) much easier.

### DIFF
--- a/docs/src/dev_reference.md
+++ b/docs/src/dev_reference.md
@@ -91,6 +91,7 @@ JuliaInterpreter.framedict
 JuliaInterpreter.genframedict
 JuliaInterpreter.compiled_methods
 JuliaInterpreter.compiled_modules
+JuliaInterpreter.interpreted_methods
 ```
 
 ## Utilities
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -21,6 +21,13 @@ rather than recursed into via the interpreter.
 const compiled_methods = Set{Method}()
 
 """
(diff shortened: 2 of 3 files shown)


timholy:
I'm in favor. What's the plan for how you add the methods, though? Can you use LanguageServer to identify all methods in a module?

pfitzseb:
See https://github.com/julia-vscode/julia-vscode/pull/1888#issuecomment-785847683 for the UX.

If you have a running process we give you a list of names you can force to be compiled/interpreted, but there's also an input field that takes e.g. `Base.gcd`/`-Base.gcd` to make sure `gcd` is compiled/interpreted and `Base`/`Base.` to make sure `Base`/`Base` and all submodules are compiled.

The default is now to treat all of Base and submodules as compiled, but whitelist certain higher order functions (which I didn't think too hard about, to be honest -- the list is [here](https://github.com/julia-vscode/julia-vscode/pull/1888/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R916-R965)). The defaults set by JuliaInterpreter are also always in place (even though they aren't reflected in the UI); it's possible to get rid of them with the functionality in this PR though.

timholy (approved):

PR #469, opened 2021-03-13
always use QuoteNode in eval_code

simeonschaub:
This fixes a pretty nasty segfault in Debugger.jl for me (see https://gist.github.com/simeonschaub/33313623fd14ecbaa147b6e77946154c). @KristofferC Was there a particular reason for introducing `maybe_quote` here and not just quote everything in a QuoteNode?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JuliaInterpreter"
 uuid = "aa1ae85d-cabe-5617-a682-6adf51b2e16a"
-version = "0.8.10"
+version = "0.8.11"
 
 [deps]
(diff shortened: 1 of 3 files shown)


KristofferC:
>  Was there a particular reason for introducing maybe_quote here and not just quote everything in a QuoteNode?

I think it was just carried over from the old ASTInterpreter, but not sure. No particular reason from what I recall, at least.

timholy (approved):
Thanks for the test case, too!

simeonschaub:
Opened an issue for the failing test cases: #470. These were failing before as well, so not related to this PR.

Ok if I bump the patch version and tag a new release right away?

timholy:
That would be great.

Issue #472, opened 2021-04-13
Interpretation does not respect world ages


maleadt:
CUDA.jl 3.0 for Julia 1.6 uses a hack (awaiting https://github.com/JuliaLang/julia/pull/39697) that registers methods in an unreachable world age for the purpose of device overrides:

```julia
julia> const override_world = typemax(Csize_t) - 1
0xfffffffffffffffe

julia> macro unreachable(ex)
           quote
               world_counter = cglobal(:jl_world_counter, Csize_t)
               regular_world = unsafe_load(world_counter)

               $(Expr(:tryfinally, # don't introduce scope
                   quote
                       unsafe_store!(world_counter, $(override_world-1))
                       $(esc(ex))
                   end,
                   quote
                       unsafe_store!(world_counter, regular_world)
                   end
               ))
           end
       end
@unreachable (macro with 1 method)

julia> foobar() = 42
foobar (generic function with 1 method)

julia> @unreachable foobar() = error("unreachable")
foobar (generic function with 1 method)

julia> foobar()
42
```

However, when interpreting this code we do execute the unreachable version:

```
julia> @interpret foobar()
ERROR: unreachable
Stacktrace:
  [1] maybe_evaluate_builtin(frame::Frame, call_expr::Expr, expand::Bool)
    @ JuliaInterpreter ~/Julia/depot/packages/JuliaInterpreter/lObtQ/src/builtins.jl:225
  [2] evaluate_call_recurse!(recurse::Any, frame::Frame, call_expr::Expr; enter_generated::Bool)
    @ JuliaInterpreter ~/Julia/depot/packages/JuliaInterpreter/lObtQ/src/interpret.jl:204
  [3] evaluate_call_recurse!
    @ ~/Julia/depot/packages/JuliaInterpreter/lObtQ/src/interpret.jl:201 [inlined]
  [4] eval_rhs(recurse::Any, frame::Frame, node::Expr)
    @ JuliaInterpreter ~/Julia/depot/packages/JuliaInterpreter/lObtQ/src/interpret.jl:395
  [5] step_expr!(recurse::Any, frame::Frame, node::Any, istoplevel::Bool)
    @ JuliaInterpreter ~/Julia/depot/packages/JuliaInterpreter/lObtQ/src/interpret.jl:538
  [6] step_expr!(recurse::Any, frame::Frame, istoplevel::Bool)
[...]

PR #478, opened 2021-04-26
Always interpret in current world

pfitzseb:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/472.

This seems strictly more useful than the current behavior, but maybe I'm missing something.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -25,6 +25,12 @@ function to_function(@nospecialize(x))
     isa(x, GlobalRef) ? getfield(x.mod, x.name) : x
 end
 
+@static if isdefined(Base, :get_world_counter)
+    import Base: get_world_counter
+else
+    get_world_counter() = ccall(:jl_get_world_counter, UInt, ())
+end
+
 """
     method = whichtt(tt)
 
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -32,7 +32,7 @@ Like `which` except it operates on the complete tuple-type `tt`.
 """
 function whichtt(@nospecialize(tt))
     # TODO: provide explicit control over world age? In case we ever need to call "old" methods.
-    m = ccall(:jl_gf_invoke_lookup, Any, (Any, UInt), tt, typemax(UInt))
+    m = ccall(:jl_gf_invoke_lookup, Any, (Any, UInt), tt, Base.get_world_counter())
```
aviatesk, line 35: and we need something like below for v1.0 compat ...
```julia
@static if isdefined(Base, :get_world_counter)
    import Base: get_world_counter
else
    get_world_counter() = ccall(:jl_get_world_counter, UInt, ())
end
```

aviatesk (changes requested):
Sounds reasonable to me.
I guess maybe the previous implementation referred to [the implementation of `Base.which`](https://github.com/JuliaLang/julia/blob/75c4f5562a0cd83f47c5695caee33c45630e76fc/base/reflection.jl#L1300), which certainly uses `typemax(UInt)`, but I still don't think there is any need to use that instead of `get_world_counter` in both cases.

aviatesk (approved):

PR #480, opened 2021-05-05
fix #479: buggy handling of ccall

### LINKED ISSUE #479: TypeError: in typeassert, expected Type, got a value of type Ptr{Nothing}
Julia 1.6
JuliaInterpreter v0.8.14

MWE: 

```julia
using JuliaInterpreter, GLFW, ModernGL

function gen_buffers()
    result = GLuint[0]
    ModernGL.glGenFramebuffers(1, result)
    id = result[1]
    if id <= 0
        error("glGenFramebuffers returned invalid id. OpenGL Context active?")
    end
    return id
end

function test()
    w = GLFW.Window()
    gen_buffers()
end

@interpret test()
```

simeonschaub:
This was quite annoying to track down. The issue here was that
`replace_ssa!` modifies the passed expression, but it is possible for
there to be multiple references to the same copy of an expression in
`stmts`, so the replacement would be accidentally applied twice
generating wrong code.

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -25,18 +25,18 @@ function extract_inner_call!(stmt::Expr, idx, once::Bool=false)
     return nothing
 end
 
-function replace_ssa!(@nospecialize(stmt), ssalookup)
-    isa(stmt, Expr) || return nothing
-    for (i, a) in enumerate(stmt.args)
+function replace_ssa(@nospecialize(stmt), ssalookup)
+    isa(stmt, Expr) || return stmt
+    return Expr(stmt.head, Any[
         if isa(a, SSAValue)
-            stmt.args[i] = SSAValue(ssalookup[a.id])
+            SSAValue(ssalookup[a.id])
         elseif isa(a, NewSSAValue)
-            stmt.args[i] = SSAValue(a.id)
+            SSAValue(a.id)
         else
-            replace_ssa!(a, ssalookup)
+            replace_ssa(a, ssalookup)
         end
-    end
-    return nothing
+        for a in stmt.args
+    ]...)
 end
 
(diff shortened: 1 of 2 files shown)


src/optimize.jl
```diff
@@ -25,18 +25,18 @@ function extract_inner_call!(stmt::Expr, idx, once::Bool=false)
...
-            stmt.args[i] = SSAValue(a.id)
+            SSAValue(a.id)
         else
-            replace_ssa!(a, ssalookup)
+            replace_ssa(a, ssalookup)
         end
-    end
-    return nothing
+        for a in stmt.args
+    ]...)
```
pfitzseb, line 39: Isn't splatting a potentially very long vector considered bad form?

pfitzseb (approved):
LGTM (except for that splat, but not sure if that actually matters).

src/optimize.jl
simeonschaub: This vector should typically be relatively short, since exprs don't usually have that many arguments. There isn't currently a nice way to create an expression like this without splatting unfortunately, but since the constructor is a builtin there shouldn't be any unnecessary specialization, so this should be fine.

KristofferC (approved):

src/optimize.jl
pfitzseb: Couldn't you write this as
```
ex = Expr(stmt.head)
ex.args = Any[
    if isa(a, SSAValue)
        SSAValue(ssalookup[a.id])
    elseif isa(a, NewSSAValue)
        SSAValue(a.id)
    else
         replace_ssa(a, ssalookup)
    end
    for a in stmt.args
]
return ex
```
But yeah, as you said, doesn't matter much ;)

PR #482, opened 2021-05-12
setup `_INACTIVE_EXCEPTION`, which can be distinguished from thrown objects

aviatesk:
Previously `FrameData.last_exception` is initialized with `nothing`,
which can't be distinguished from thrown objects, which technically can
be `nothing` as well.
This PR setups `_INACTIVE_EXCEPTION` singleton type and make
`FrameData.last_exception` initialized with it.


KristofferC (approved):

aviatesk:
Fixed documentation build. Going to merge this once CI gets passed and then tag a new patch release.

PR #484, opened 2021-05-28
make next command less skipping

bicycle1885:
This PR makes the next command less skipping. The motivation is described here: https://github.com/JuliaDebug/Debugger.jl/issues/291.

~~My approach is to add a keyword argument (`sameline`) to the `maybe_next_call!` function so that it tires to find a call or a return statement without stepping forward to the next lines. If it fails, it stops at the current program counter.~~

With this patch (and a minor modification to Debugger.jl), assignments and branches are no longer skipped and now it works like debuggers of other languages (such as Python's ipdb):
```
julia> function foo(x)
           #println(x)
           y = 0
           if x
               y = 1
           end
           return y
       end

julia> @enter foo(true)
In foo(x) at REPL[1]:1
 1  function foo(x)
 2      #println(x)
>3      y = 0
 4      if x
 5          y = 1
 6      end
 7      return y
 8  end

About to run: 0
1|debug> n
In foo(x) at REPL[1]:1
 1  function foo(x)
 2      #println(x)
 3      y = 0
>4      if x
 5          y = 1
 6      end
 7      return y
 8  end

About to run: goto %4 if not _J2
1|debug>
In foo(x) at REPL[1]:1
 1  function foo(x)
 2      #println(x)
 3      y = 0
 4      if x
>5          y = 1
 6      end
 7      return y
 8  end

About to run: 1
1|debug>
In foo(x) at REPL[1]:1
 3      y = 0
 4      if x
 5          y = 1
 6      end
>7      return y
 8  end

About to run: return 1
1|debug>
1
```

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -132,9 +132,19 @@ next_call!(frame::Frame, istoplevel::Bool=false) = next_call!(finish_and_return!
 Return the current program counter of `frame` if it is a `:return` or `:call` expression.
 Otherwise, step through the statements of `frame` until the next `:return` or `:call` expression.
 """
-maybe_next_call!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false) =
-    maybe_next_until!(frame -> is_call_or_return(pc_expr(frame)), recurse, frame, istoplevel)
-maybe_next_call!(frame::Frame, istoplevel::Bool=false) = maybe_next_call!(finish_and_return!, frame, istoplevel)
+function maybe_next_call!(@nospecialize(recurse), frame::Frame, istoplevel::Bool=false; sameline::Bool=false)
+    sameline || return maybe_next_until!(frame -> is_call_or_return(pc_expr(frame)), recurse, frame, istoplevel)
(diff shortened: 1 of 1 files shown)


KristofferC:
> `About to run: goto %4 if not _J2`

I don't think it is good to stop at these internal statements.

pfitzseb:
Although that seems to be more of a printing issue? Stopping at branch conditions makes sense to me.

bicycle1885:
How to print assignments and branch conditions is a separated problem, and it should be fixed in Debugger.jl.

NOTE: I found the current impl is completely broken. So, there is no need to check the patch at the moment.

bicycle1885:
I've changed the design: `next_line!` continues to step forward until it finds a statement which is one of:
- function call
- assignment (both local and global)
- return
- goto (both conditional and unconditional)

Here, "assignment" excludes assignments to compiler-generated variables. I'm not perfectly sure how it can be achieved, but it now checks if `slotnames` corresponding to the variable is `Symbol("")` or not. Any better ideas are welcomed.

PR #486, opened 2021-06-02
Improve handling of duplicate slotnames

pfitzseb:
Turns out that we can sometimes end up with duplicate slotnames, which breaks the current logic in `prepare_slotfunction`.

E.g. 
```
function debug(x)
    for iter in Iterators.CartesianIndices(x)
        i = iter[1]
        c = i
        a, b, c, d = tmp()
    end
    return x
end
```
results in 
```
julia> c = @code_lowered debug((1,2,3));

julia> c.slotnames
10-element Vector{Symbol}:
 Symbol("#self#")
 :x
 Symbol("")
 Symbol("")
 :iter
 :d
 :b
 :a
 :c
 :i
 ```
which means that we never insert the proper assignment for `i` (because `Symbol("")` is in there twice).

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -188,16 +188,10 @@ function shouldbreak(frame::Frame, pc::Int)
 end
 
 function prepare_slotfunction(framecode::FrameCode, body::Union{Symbol,Expr})
-    uslotnames = Set{Symbol}()
     framename, dataname = gensym("frame"), gensym("data")
     assignments = Expr[:($dataname = $framename.framedata)]
     default = Unassigned()
-    for slotname in framecode.src.slotnames
-        if slotname ∉ uslotnames
-            push!(uslotnames, slotname)
-        else
-            continue
-        end
+    for slotname in unique(framecode.src.slotnames)
         list = framecode.slotnamelists[slotname]
         if length(list) == 1
             maxexpr = :($dataname.last_reference[$(list[1])] > 0 ? $(list[1]) : 0)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/breakpoints.jl [hunk 1]
```diff
@@ -189,19 +189,18 @@ end
...
-        end
-    end
     framename, dataname = gensym("frame"), gensym("data")
     assignments = Expr[:($dataname = $framename.framedata)]
     default = Unassigned()
-    for i = 1:length(slotnames)
-        slotname = framecode.src.slotnames[i]
-        qslotname = QuoteNode(slotname)
+    for slotname in framecode.src.slotnames
+        if slotname === Symbol("")
```
pfitzseb, line 196: Not entirely sure whether this is correct, but our tests at least say that it's fine.

KristofferC (approved):
Looks like CI want some special handling for Julia 1.0.

src/breakpoints.jl
KristofferC: Would things pass even if this is removed?

pfitzseb: Yep. Removing the check ;)

src/breakpoints.jl
```diff
@@ -193,9 +193,6 @@ function prepare_slotfunction(framecode::FrameCode, body::Union{Symbol,Expr})
     assignments = Expr[:($dataname = $framename.framedata)]
     default = Unassigned()
     for slotname in framecode.src.slotnames
```
KristofferC, line 195: I guess `uslotnames` could be removed and make the loop go over `unique(framecode.src.slotnames)` instead?

pfitzseb: Ye.

PR #487, opened 2021-06-08
update to https://github.com/JuliaLang/julia/pull/41018

aviatesk:
Now `DataType`s don't have `hasfreetypevars` field.
We can use `Base.has_free_typevars` instead while keeping backward compat.


simeonschaub (approved):

KristofferC (approved):

aviatesk:
Thanks for your quick reviews !

PR #491, opened 2021-08-08
allow interpreting TypedSlots

simeonschaub:
This comes up when trying to interpret the output of `code_warntype` for
example. Is this reasonable to support?

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -419,7 +419,7 @@ function replace_coretypes_list!(list::AbstractVector; rev::Bool)
         end
         if isa(x, Core.SSAValue)
             return SSAValue(x.id)
-        elseif isa(x, Core.SlotNumber)
+        elseif isa(x, Core.SlotNumber) || isa(x, Core.TypedSlot)
             return SlotNumber(x.id)
         end
         return x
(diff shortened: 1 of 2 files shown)


timholy:
Cool. Here's one big issue:
```julia
julia> const objs = Any['x', 3.2, "hello", false, which(+, (Int,Int))]
5-element Vector{Any}:
      'x': ASCII/Unicode U+0078 (category Ll: Letter, lowercase)
     3.2
      "hello"
 false
      +(x::T, y::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8} in Base at int.jl:87

julia> f(x) = x isa Float64

julia> g(x) = x isa AbstractFloat

julia> callf() = f(rand(objs))
callf (generic function with 1 method)

julia> callg() = g(rand(objs))
callg (generic function with 1 method)

julia> using BenchmarkTools

julia> @btime callf()
  3.943 ns (0 allocations: 0 bytes)
false

julia> @btime callg()
  21.442 ns (0 allocations: 0 bytes)
false
```
This package has to deal with rampant non-inferrability, and is extremely performance-sensitive. So to have this you'd have to write it purely in terms of concrete-type `isa` checks.

And here's the other big issue: see #242, #257, #277

simeonschaub:
Ok, I have updated this with a much simpler version where we just always normalize `Core.TypedSlot` to `JuliaInterpreter.SlotNumber` instead. If we wanted to preserve the type information, we could also introduce a new type for them instead, but this works fine for the Cthulhu use case at least.

timholy (commented):
If we can get a test, this can be merged.

PR #494, opened 2021-08-13
minor inferrability improvements for `Base.StackTraces.StackFrame(::Frame)`

aviatesk:


src/utils.jl
```diff
@@ -709,8 +710,8 @@ function Base.StackTraces.StackFrame(frame::Frame)
     end
     Base.StackFrame(
         fname,
-        Symbol(JuliaInterpreter.getfile(frame)),
-        JuliaInterpreter.linenumber(frame),
+        Symbol(getfile(frame)),
+        @something(linenumber(frame), getline(linetable(frame, 1))),
```
aviatesk, line 714: I couldn't reason that `linenumber(frame)` always returns `Int` here, so added a fallback, which may not be very accurate information but not lead to runtime error at least.

src/packagedef.jl
```diff
@@ -46,6 +46,12 @@ else
     end
 end
 
+if !@isdefined(var"@something")
```
KristofferC, line 49:
```suggestion
if !isdefined(Base, Symbol("@something"))
```

src/packagedef.jl
```diff
@@ -46,6 +46,12 @@ else
     end
 end
 
+if !isdefined(Base, Symbol("@something"))
+    macro something(x...)
+        :($(GlobalRef(Base, :something))($(map(esc, x)...)))
```
simeonschaub, line 51: Just curious, why the `GlobalRef` here?

aviatesk: I'm convinced by Jeff's comment here: https://github.com/JuliaLang/julia/pull/37219#discussion_r478520332 :p

simeonschaub: I meant instead of `:(something($(map(esc, x)...))`, since macro hygiene should take care of the rest. The `GlobalRef` certainly won't hurt though either.

aviatesk: I thought your alternative doesn't work when there is no `something` is defined in macro callsite, but it does ?
```julia
macro something2(x...); :(something($(map(esc, x)...))); end
baremodule bar
Main.@something2(nothing, 1) # `something` not defined here ?
end
```
I'm honestly not too sure why `something2` captures `something` there.

simeonschaub: This is what macro hygiene does. All names that aren't escaped with `esc` are resolved inside the module the macro is defined in.

aviatesk: So is my understanding correct that we really don't need interpolations for any global references within macro ?

simeonschaub: Yes, unless you are afraid `something` might be shadowed inside the module the macro was defined in, that won't be needed.

aviatesk: Thanks, then I think it would be better in terms of readability if we use simple textual presentation rather than explicit `GlobalRef` most of cases. Will change it here as well.

simeonschaub (approved):

Issue #44, opened 2019-02-17
Compiled resumers


timholy:
#41 introduced the idea of using `Compiled` mode to finish tests (or preparatory steps) that take too long, so that an `Abort` doesn't nix later tests depending on the results. However, the implementation has several deficiencies:
- it starts again from the beginning of the highest-level statement; instead, it should start at the lowest level on the stack and work its way back up. That will prevent it from repeating work (which is what it does now).
- in some cases it may not fix the performance problem: it still has to interpret the frame that threw the `Abort`. `Compiled` means it won't need to recurse anymore, but if that frame happens to have an insanely long loop you'll still be stuck for a long time.

What we need instead is the notion of a compiled "resumer" function: starting from the method `body`, build a (temporary) function that schematically looks like this (would be built from lowered expressions in reality):

```julia
function #foo_resumer#34(slotvalues...)
    $(slotname[1]) = slotvalues[1]
...
    $(slotname[n]) = slotvalues[n]
    goto pc_where_it_quit
    $body
end
```

Then this could be called, in which case it would finish the computations in a truly compiled mode.

It has not escaped my attention that together with `evaluate_limited!`, this implements a possible means of addressing Julia's compiler-latency problems. Once this last piece is in place, I believe it  would be a complete implementation of "interpret by default, then fall back gracefully to compiled mode for 'hot' portions of the code."

Issue #23, opened 2019-02-13
Plan for tags + release


KristofferC:
@timholy What are your thoughts about tagging + releasing and making a bit of noise so that people start trying things out. Do you feel the interpreter needs some more maturing based on the julia tests you are running or is it in a stable enough state for external use?

timholy:
Just having people try stuff probably doesn't gain us much, given that we know (see #13) that many things are broken. But more contributors would be golden. TBH given my experience with Revise I am a little skeptical that many will jump on the bandwagon, but you never know.

From the standpoint of making noise, I also confess a small temptation to wait until we have something really good so that people don't have a bad first impression. From the standpoint of "internals" (aka, leaving aside the issue of Juno integration) I think the two big tasks left are (1) improving the correctness of JuliaInterpreter, and (2) integrating with Revise. Maybe I'm naive, but I am guessing fancy features like breakpoints will be pretty easy.

If I were to design a dream path, given the improvements in #24 so that #13 is more useful, I'd shift my own emphasis: I'd switch to tackling (2) while others slog away at (1) (I'm happy to provide advice, of course). Then when we've gotten somewhere I or someone would add the breakpoint stuff here, someone (not me) would do the Juno side. Then we make a big announcement, and then hope that a flood of excited people help out with all the difficult correctness issues (1).

Thoughts?

pfitzseb:
I'd be happy to take a look at implementing breakpoints (and obviously integrate this with Juno). I can also try to help with (1), but I'd probably not be as immediately useful there (although the docs are going to help a lot, so who knows).

davidanthoff:
I'd love to see the package registered and tagged, just so that it is a bit easier to use. From my point of view the making noise part could then still come later :)

timholy:
I think we're now getting close; as of this morning the prospect for robust signature detection is looking very good, so I'm increasingly confident we aren't going to need a massive API break in the very near future. (The world-age issues did require pretty massive breakage, so it was good we waited.)

Before we make this public, we need to decide our tolerance for errors. In case it's not clear to folks, here's why  #13 is so important:  we're going to implement breakpoints (including "break on error") by running all code in the interpreter. That means JuliaInterpret has to be competent to run code all the way from high-level user code down to the ccalls, intrinsic functions, and builtins that lie at the bottom-most level of the call chain. If there are any errors on the way, the user can't even step through code.

[...]

PR #495, opened 2021-08-14
[Do Not Merge] Run Julia Base Tests

simeonschaub:
This is just to see what the current state here is, since I didn't feel
like running this on my Laptop.


oxinabox:
Maybe we could have a GHA that it linked only to manual (or schedule).
That runs and pushes results into a issue.

simeonschaub:
Yes, that was my eventual goal, but lets see how this goes first.

timholy:
In case a little more history helps, see #13

In case anyone is interested in investing in this package for the purpose of latency reduction, see:
- #44 
- #309

Having sunk most of January-March of 2019 into this package, I then had to return to other duties and never really "came back." I do think there are some interesting possibilities, so I'm pretty jazzed to see people playing with this.

And #23 was a fun trip down memory lane!

simeonschaub:
Can anyone see the actual logs?

pfitzseb:
Nope. Seems like something *really* breaks (or maybe there's just too much being printed?).

simeonschaub:
I ran this on my machine and it eventually froze, so I suspect this might be an oversubscription issue.

KristofferC:
Not sure if this should be open anymore, https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/754 made a lot of progress on julia testing

PR #497, opened 2021-08-18
report coverage when interpreting

simeonschaub:
Still need to write some tests and perhaps run some benchmarks, but
seems to work, at least for simple cases.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -433,6 +433,20 @@ function check_isdefined(frame, @nospecialize(node))
     end
 end
 
+function coverage_visit_line!(frame::Frame)
+    pc, code = frame.pc, frame.framecode
+    code.report_coverage || return
+    src = code.src
+    codeloc = src.codelocs[pc]
+    if codeloc != frame.last_codeloc
+        linetable = src.linetable::Vector{Any}
+        lineinfo = linetable[codeloc]::Core.LineInfoNode
+        file, line = String(lineinfo.file), lineinfo.line
+        ccall(:jl_coverage_visit_line, Cvoid, (Cstring, Csize_t, Cint), file, sizeof(file), line)
+        frame.last_codeloc = codeloc
+    end
+end
+
 # For "profiling" where JuliaInterpreter spends its time. See the commented-out block
 # in `step_expr!`
 const _location = Dict{Tuple{Method,Int},Int}()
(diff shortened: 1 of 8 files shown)


src/interpret.jl
```diff
@@ -433,6 +433,20 @@ function check_isdefined(frame, @nospecialize(node))
     end
 end
 
+function coverage_visit_line!(frame::Frame)
+    pc, code = frame.pc, frame.framecode
+    code.report_coverage || return
+    src = code.src
+    codeloc = src.codelocs[pc]
```
KristofferC, line 440: Can use https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/22d1914cb4fcd08de6f27a8cfe97a5f5205cfe06/src/utils.jl#L360?

simeonschaub: Seems like that would be less efficient, since we don't need to actually look up `codeloc` in the linetable unless the location has changed. Since this function is probably quite hot, it's probably worth doing this way.

KristofferC: 👍

KristofferC:
Should this be done in the "optimize" stage (https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/19757b45fa438cf8335601099a21525a88c9bb39/src/optimize.jl#L148) by inserting these only at places where the line location changes. I think that is how Base does it.

simeonschaub:
Maybe, but wouldn't that be unwanted when debugging for example, since you would need to step over more lines if coverage is enabled? I think the reason base does it that way is because everything gets compiled, so it actually needs to insert the appropriate instructions into the generated code.

KristofferC:
I guess, the interpreter might be slow enough that this will work jus as well.

timholy (approved):
This looks great! Excited to have this.

One question: how expensive is the `ccall`? Would it be worth keeping track of a `BitVector`/`Vector{Bool}`whatever that tracks whether a given `pc` has been previously covered? The interpreter could conceivably work differently, rather than counting hits it could just track yes/no. (I think that's all Codecov shows anyway.) But if the `ccall` isn't particularly expensive, it's surely better to work the same way Julia itself does.

Needs tests.

simeonschaub:
It seems to take about 8µs, but doesn't look like the overhead makes much of a difference in real-world examples. Running the JuliaInterpreter test suite with `--code-coverage=user`, I can't measure any significant overhead, with `--code-coverage=all`, it takes 1'40 instead of 1'34, but that might not even come from JuliaInterpreter alone. I even ran the entire ChainRules test suite with this in https://github.com/JuliaDiff/ChainRules.jl/pull/505 and while GH actions can generally be a bit noisy performance-wise it only added about 2-3 minutes over the 25 minutes or so it takes otherwise. That seems acceptable to me.

PR #498, opened 2021-10-27
fix coverage test on nightly

simeonschaub:
Depends on JuliaLang/julia#42810 to pass. I did also observe some other
test failures on nightly I couldn't really explain.

### DIFF
--- a/test/interpret.jl
+++ b/test/interpret.jl
@@ -645,7 +645,14 @@ module VecTest
     end
     f() = 1.0 * one(Tensor{2,3})
 end
-let a = (VecElement{Float64}(1.0), VecElement{Float64}(2.0))
+let
+    # NOTE we need to make sure this code block is compiled, since vecadd is generated function,
+    # but currently `@interpret` doesn't handle a call to generated functions very well
+    @static if isdefined(Base, :Experimental) &&
+               isdefined(Base.Experimental, Symbol("@force_compile"))
+        Base.Experimental.@force_compile
+    end
+    a = (VecElement{Float64}(1.0), VecElement{Float64}(2.0))
     @test @interpret(VecTest.vecadd(a, a)) == VecTest.vecadd(a, a)
 end
 @test @interpret(VecTest.f()) == [1 0 0; 0 1 0; 0 0 1]
--- a/test/limits.jl
+++ b/test/limits.jl
@@ -98,7 +98,7 @@ module EvalLimited end
     end
     @test 8 < EvalLimited.s < 50  # with Compiled(), 9 statements per iteration
     @test length(aborts) == 1
-    @test aborts[1].at.line ∈ (2, 3, 4)  # 2 corresponds to lowering of the for loop
+    @test aborts[1].at.line ∈ (2, 3, 4, 5)  # 2 corresponds to lowering of the for loop
 
     # Now try again with recursive stack
     empty!(aborts)
@@ -117,8 +117,8 @@ module EvalLimited end
     @test length(aborts) == 1
     lin = aborts[1].at
     if lin.file == Symbol("fake.jl")
-        @test lin.line ∈ (2, 3, 4)
+        @test lin.line ∈ (2, 3, 4, 5)
     else
-        @test lin.file == Symbol("range.jl")  # if it aborts in `iterate`
+        @test lin.method === :iterate || lin.method === :getproperty
     end
 end
--- a/Project.toml
+++ b/Project.toml
@@ -15,7 +15,6 @@ julia = "1"
 [extras]
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
-DiffUtils = "8294860b-85a6-42f8-8c35-d911f667b5f6"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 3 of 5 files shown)


test/limits.jl
```diff
@@ -40,6 +40,9 @@ end
     frame = Frame(modexs[5]...)
     if VERSION < v"1.4.0-DEV.475"
         @test Aborted(frame, 1).at.file == Symbol("util.jl")
+    elseif isdefined(Base, :Experimental) &&
+           isdefined(Base.Experimental, Symbol("@force_compile"))
+        @test Aborted(frame, 1).at.file == Symbol("timing.jl")
```
timholy, line 45: Ooh, that's unfortunate. I don't immediately understand why it's different from the old heuristic, but I haven't poked at it. Can you file a Julia issue? I regard line-number issues as bugs.

test/interpret.jl
```diff
@@ -645,7 +645,14 @@ module VecTest
...
 end
-let a = (VecElement{Float64}(1.0), VecElement{Float64}(2.0))
+let
+    # NOTE we need to make sure this code block is compiled, since vecadd is generated function,
+    # but currently `@interpret` doesn't handle a call to generated functions very well
+    @static if isdefined(Base, :Experimental) &&
+               isdefined(Base.Experimental, Symbol("@force_compile"))
+        Base.Experimental.@force_compile
+    end
+    a = (VecElement{Float64}(1.0), VecElement{Float64}(2.0))
```
timholy, line 655: You sure you need this? It didn't fail for me on nightly.

aviatesk: Yeah, I couldn't reproduce the error within REPL, but I also confirmed `Pkg.test` will hit [this line](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/398721d10be2edf40883a11b98e210ccda5393e3/src/construct.jl#L271) consistently without this change.
I think the root problem here is that `@interpret` doesn't configure `enter_generated` when calling `enter_call_expr`.

test/limits.jl
aviatesk: Is this really bug? The line numbers appearing in lowered code seems just identical:
> v1.7
```julia
julia> ex = Main, Base.parse_input_line("""
           x = 1
           for i = 1:10
               x += 1
           end
           let y = 0
               z = 5
           end
           if 2 > 1
               println("Hello, world!")
           end
           @elapsed sum(rand(5))
           """; filename="fake.jl"); Meta.lower(Main, ex)
:($(Expr(:toplevel, :(#= fake.jl:1 =#), :(x = 1), :(#= fake.jl:2 =#), :(for i = 1:10
      #= fake.jl:3 =#
      x += 1
  end), :(#= fake.jl:5 =#), :(let y = 0
      #= fake.jl:6 =#
      z = 5
  end), :(#= fake.jl:8 =#), :(if 2 > 1
      #= fake.jl:9 =#
      println("Hello, world!")
  end), :(#= fake.jl:11 =#), :($(Expr(Symbol("hygienic-scope"), quote
    #= timing.jl:297 =#
    while false
        #= timing.jl:297 =#
    end
    #= timing.jl:298 =#
    local t0 = time_ns()
    #= timing.jl:299 =#
    $(Expr(:escape, :(sum(rand(5)))))
    #= timing.jl:300 =#
    (time_ns() - t0) / 1.0e9
end, Base))))))
```
> nightly
```julia
julia> ex = Base.parse_input_line("""
           x = 1
           for i = 1:10
               x += 1
           end
           let y = 0
               z = 5
           end
           if 2 > 1
               println("Hello, world!")
           end
           @elapsed sum(rand(5))
           """; filename="fake.jl"); Meta.lower(Main, ex)
:($(Expr(:toplevel, :(#= fake.jl:1 =#), :(x = 1), :(#= fake.jl:2 =#), :(for i = 1:10
      #= fake.jl:3 =#
      x += 1
      #= fake.jl:4 =#
  end), :(#= fake.jl:5 =#), :(let y = 0
      #= fake.jl:6 =#
      z = 5
  end), :(#= fake.jl:8 =#), :(if 2 > 1
      #= fake.jl:9 =#
      println("Hello, world!")
  end), :(#= fake.jl:11 =#), :($(Expr(Symbol("hygienic-scope"), quote
    #= timing.jl:356 =#
    $(Expr(Symbol("hygienic-scope"), :($(Expr(:meta, :force_compile))), Base.Experimental))
    #= timing.jl:357 =#
    local t0 = time_ns()
    #= timing.jl:358 =#
    $(Expr(:escape, :(sum(rand(5)))))
    #= timing.jl:359 =#
    (time_ns() - t0) / 1.0e9
end, Base))))))
```

The actual difference seems to be that v1.7 produces `NewvarNode` while nightly doesn't:
> v1.7
```julia
julia> modexs = collect(ExprSplitter(Main, ex)); frame = Frame(modexs[5]...); frame.framecode.src
CodeInfo(
    @ fake.jl:11 within `top-level scope`
1 ─     Core.NewvarNode(:(#310#t0))
   ┌ @ timing.jl:297 within `macro expansion`
2 ┄│     goto #4 if not false
3 ─│     goto #2
   └
   ┌ @ timing.jl:298 within `macro expansion`
4 ─│     _J1 = ($(QuoteNode(time_ns)))()
│  └
│  ┌ @ timing.jl:299 within `macro expansion`
│  │     rand(5)
│  │     sum(%J5)
│  └
│  ┌ @ timing.jl:300 within `macro expansion`
│  │     ($(QuoteNode(time_ns)))()
│  │     ($(QuoteNode(-)))(%J7, _J1)
│  │     ($(QuoteNode(/)))(%J8, 1.0e9)
│  └
└──     return %J9
)
```

> nightly
```julia
julia> modexs = collect(ExprSplitter(Main, ex)); frame = Frame(modexs[5]...); frame.framecode.src
CodeInfo(
    @ fake.jl:11 within `top-level scope`
   ┌ @ timing.jl:356 within `macro expansion`
1 ─│     $(Expr(:meta, :force_compile))
│  │ @ timing.jl:357 within `macro expansion`
│  │     _J1 = ($(QuoteNode(time_ns)))()
│  │ @ timing.jl:358 within `macro expansion`
│  │     rand(5)
│  │     sum(%J3)
│  │ @ timing.jl:359 within `macro expansion`
│  │     ($(QuoteNode(time_ns)))()
│  │     ($(QuoteNode(-)))(%J5, _J1)
│  │     ($(QuoteNode(/)))(%J6, 1.0e9)
│  └
└──     return %J7
)
```

So I'd say the first line number of that chunk was `fake.jl` sort of accidentally.

timholy: xref #500

test/interpret.jl
timholy: I'm OK with this if it's necessary.

timholy:
I think one of the failures should be fixed by #500 instead, but otherwise I'm good with this.

timholy:
Sorry for the rebase. Once that's done let's merge this.

PR #499, opened 2021-12-05
update LTS version, drop v1.0 support

aviatesk:
Let's live on the cutting edge :)


simeonschaub:
Will need to update the Julia compat as well as the minor version

pfitzseb:
FWIW, I'd kinda like to keep 1.0 compat for the VS Code extension. 
It's not _super_ important though since I suppose we can disable some features for old Julia versions.

KristofferC:
VSCode could just still use JuliaInterpreter 0.8.21 on 1.0? Or is the worry that future features won't be available?

pfitzseb:
We're currently vendoring and side-loading all packages. Shipping multiple versions of a package would be feasible but fairly annoying. 
Then again, only supporting a debugger on LTS and newer doesn't seem too bad.

aviatesk:
It's painful to try to keep v1.0 compat imho mostly because this package heavily interacts with Julia internals, as we just see yet another example at https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/500#issuecomment-986774575.
So I'd like to ask you to drop the debugger support for older Julia versions.

pfitzseb:
Alright, :shipit: :)

Project.toml
```diff
@@ -10,21 +10,4 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
...
-PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
-SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
-SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-TableReader = "70df011a-6618-58d7-8e16-3cf9e384cb47"
-Tensors = "48a634ad-e948-5137-8d70-aa71f2a747f4"
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
-
-[targets]
-test = ["Test", "Dates", "Distributed", "HTTP", "LinearAlgebra", "Mmap", "PyCall", "SHA", "SparseArrays", "Tensors", "TableReader", "DataFrames"]
+julia = "1.6"
```
KristofferC, line 13: Can we keep this? I don't see any advantage to the `test/Project.toml` approach. In my opinion the other way is very strange right now since it isn't a proper project.

aviatesk: Well, isn't `test/Project.toml` more "up-to-date" way to describe test-specific dependencies as described in: <https://pkgdocs.julialang.org/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above>?
I'm actually very indifferent between those two though.

> In my opinion the other way is very strange right now since it isn't a proper project.
Hm, I couldn't get the point. Any directory is a proper project as far as it contains `Project.toml`?

KristofferC: It gives you a false sense that you can activate the test project and then include the `runtests.jl`. That's why I'm saying it is but a true project since you would never activate it. It is purely a list of test dependencies, so what's the advantage of putting that in a separate file?

aviatesk: I see. Then if we use `test/Project.toml` approach, you think it should include the package itself (and thus `test/Manifest.toml` also needs to be checked-in), right?
Now I'd like to just keep test dependencies in `Project.toml`, and simply wonder what situations `test/Project.toml` is preferred.

KristofferC: > Now I'd like to just keep test dependencies in Project.toml, and simply wonder what situations test/Project.toml is preferred

Right now, I would say never. Something like https://github.com/JuliaLang/Pkg.jl/issues/1233 is needed to make these first-class to be useful. In my opinion.

aviatesk: Okay, thanks for your explanation!

pfitzseb (approved):

aviatesk:
@timholy @pfitzseb @KristofferC Can we tag 0.9? It's quite big change, but since it is a minor bump, it shouldn't break any existing system and I think we can gradually update dependents in the ecosystem.

PR #500, opened 2021-12-06
Improve file/line attribution for macros

timholy:
This makes it easier to identify the caller of a macro.
This fixes a test failure on nightly.

xref #498


aviatesk (approved):

aviatesk:
Hm, but this PR seems to cause a failure in v1.0 CI. I'd like to drop the support anyway, so I'm just fine to merge this as is.

timholy:
In this case, this change ended up fixing a failure on older Julia versions, and we were effectively doing a `@test_broken`. Nice when it works out that way.

Issue #153, opened 2019-03-15
MWE for Channel failure


macd:
It looks like ```take!``` is not getting all that was ```put!``` into the Channel
```
julia (v1.2)> include("chantest.jl")
Test Failed at /home/macd/jlang/dev/JuliaInterpreter.jl/test/chantest.jl:15
  Expression: pvals == tvals
   Evaluated: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10  …  999991, 999992, 999993, 999994, 999995, 999996, 999997, 999998, 999999, 1000000] == [1, 2, 3, 4, 5, 6, 7, 8, 9, 10  …  999735, 999736, 999737, 999738, 999739, 999740, 999741, 999742, 999743, 999744]
ERROR: Error while loading expression starting at /home/macd/jlang/dev/JuliaInterpreter.jl/test/chantest.jl:21
caused by [exception 1]
There was an error during testing
```
Here is the MWE
```
include("utils.jl")

module ChanTest
using Random
using Test
end

m = ChanTest

ex = quote
    c = Channel{Int}(Inf)
    pvals = map(i->put!(c,i), 1:10^6)
    tvals = Int[take!(c) for i in 1:10^6]
    @test pvals == tvals
end

nstmts = 100000

modexs, _ = JuliaInterpreter.split_expressions(m, ex)
for modex in modexs
    frame = JuliaInterpreter.prepare_thunk(modex)
    nstmtsleft = nstmts
    while true
        ret, nstmtsleft = evaluate_limited!(frame, nstmtsleft, true)
        if isa(ret, Aborted)
            JuliaInterpreter.finish_stack!(Compiled(), frame, true)
            break
        elseif isa(ret, Some)
            break
        end
    end
end
```

goerch:
For me the critical test case in channels is

```
@testset "multiple for loops waiting on the same channel" begin
    # Test multiple "for" loops waiting on the same channel which
    # is closed after adding a few elements.
    c = Channel(32)
    results = []
    @sync begin
        for i in 1:20
            @async for ii in c
                @show i ii
                push!(results, ii)
            end
        end
        sleep(1.0)
        for i in 1:5
            @show i
            put!(c,i)
        end
        close(c)
    end
    @test sum(results) == 15
end
```

Trying an interpretation with

```
using JuliaInterpreter, Test

@interpret (
quote
    # Test multiple "for" loops waiting on the same channel which
    # is closed after adding a few elements.
[...]

PR #508, opened 2021-12-10
Fixing Julia tests

goerch:
This commit contains the changes needed to make the tests running locally against the installed Julia tests.


goerch:
These are the tests which seem to be most time consuming or not terminating:

```
interrupting channels
interrupting precompile
interrupting subtype
interrupting worlds
```

KristofferC:
Nice to see this getting worked on again!

goerch:
The current critical problem with channels is reported in #153. For subtype I hit AssertionError("is_leaf(frame)") in step_expr!. In which of my branches do you want me to try to fix this?

goerch:
> Nice to see this getting worked on again!

I'm not sure how to proceed with this patch. In its original form it intended to make the tests runnable on Windows (BTW, is the repository version working on other platforms?). I'd be willing to revert it to this initial form in order to keep the changes small and establish a baseline for further improvements regarding #13. Is this the way to go?

PR #509, opened 2021-12-16
remove more legacy code

aviatesk:
We can eliminate this code since `:struct_type`/`:abstract_type`/`:primitive_type`
expressions no longer exist in v1.6 and higher.


simeonschaub (approved):
LGTM

PR #511, opened 2022-01-12
Fix toplevel-global behaviour in ExprSplitter

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2571.

This seems to be correct, but I don't quite understand the original reasoning for including the `:global` case in `push_modex!`.

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -436,7 +436,7 @@ function push_modex!(iter::ExprSplitter, mod::Module, ex::Expr)
         # Issue #427
         modifies_scope = false
         for a in ex.args
-            if isa(a, Expr) && a.head ∈ (:local, :global)
+            if Meta.isexpr(a, :local)
                 modifies_scope = true
                 break
             end
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -435,10 +435,12 @@ function push_modex!(iter::ExprSplitter, mod::Module, ex::Expr)
     if ex.head === :toplevel || ex.head === :block
         # Issue #427
         modifies_scope = false
-        for a in ex.args
-            if Meta.isexpr(a, :local)
-                modifies_scope = true
-                break
+        if ex.head === :block
+            for a in ex.args
+                if isa(a, Expr) && a.head ∈ (:local, :global)
+                    modifies_scope = true
+                    break
+                end
             end
         end
         push!(iter.index, modifies_scope ? 0 : 1)
--- a/test/toplevel.jl
+++ b/test/toplevel.jl
@@ -540,3 +540,15 @@ end
     modexs = collect(ExprSplitter(@__MODULE__, ex))
     @test length(modexs) == 3
 end
+
+@testset "toplevel global" begin
+    ex = :(begin
+        global foo_g = 10
+        sin(foo_g)
+    end)
+    modexs = collect(ExprSplitter(@__MODULE__, ex))
+    for (mod, ex) in modexs
+        @test JuliaInterpreter.finish!(Frame(mod, ex), true) === nothing
+    end
+    @test length(modexs) == 2
+end
--- a/test/toplevel.jl
+++ b/test/toplevel.jl
@@ -550,5 +550,5 @@ end
     for (mod, ex) in modexs
         @test JuliaInterpreter.finish!(Frame(mod, ex), true) === nothing
     end
-    @test length(modexs) == 2
+    @test length(modexs) == 1
 end
--- a/test/toplevel.jl
+++ b/test/toplevel.jl
@@ -541,14 +541,24 @@ end
     @test length(modexs) == 3
 end
 
-@testset "toplevel global" begin
-    ex = :(begin
-        global foo_g = 10
-        sin(foo_g)
-    end)
+@testset "toplevel scope annotation" begin
+    ex = Base.parse_input_line("""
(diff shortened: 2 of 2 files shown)


aviatesk (commented):
I think we need to return entire block expression only when it contains `:local` declaration, but I'm not sure what's the problem with interpreting a block expression with `:global` declaration since `:global` should work anywhere in a same way?

pfitzseb:
On master:
```
julia> expr = Base.parse_input_line("""global i = 0
       function increment_global()
           global i += 1
       end

       increment_global()""")
:($(Expr(:toplevel, :(#= none:1 =#), :(global i = 0), :(#= none:2 =#), :(function increment_global()
      #= none:2 =#
      #= none:3 =#
      global i += 1
  end), :(#= none:6 =#), :(increment_global()))))

julia> for (mod, ex) in ExprSplitter(Main, expr)
           @show Frame(mod, ex)
       end
ERROR: ArgumentError: lowering did not return a `:thunk` expression, got $(Expr(:toplevel, :(#= none:1 =#), :(global i = 0), :(#= none:2 =#), :(function increment_global()
      #= none:2 =#
      #= none:3 =#
      global i += 1
  end), :(#= none:6 =#), :(increment_global())))
Stacktrace:
 [1] Frame(mod::Module, ex::Expr)
   @ JuliaInterpreter ~/.julia/packages/JuliaInterpreter/6pP9P/src/types.jl:275
 [2] macro expansion
   @ ./show.jl:1047 [inlined]
 [3] top-level scope
   @ ./REPL[11]:2
```
With this PR:
```
julia> expr = Base.parse_input_line("""global i = 0
       function increment_global()
           global i += 1
       end

       increment_global()""")
:($(Expr(:toplevel, :(#= none:1 =#), :(global i = 0), :(#= none:2 =#), :(function increment_global()
      #= none:2 =#
      #= none:3 =#
      global i += 1
  end), :(#= none:6 =#), :(increment_global()))))

julia> for (mod, ex) in ExprSplitter(Main, expr)
           @show Frame(mod, ex)
       end
Frame(mod, ex) = Frame for Main
  1* 0  1 ─     global i
  2  0  │       i = 0
  3  0  └──     return 0
Frame(mod, ex) = Frame for Main
  1* 2  1 ─      $(Expr(:thunk, CodeInfo(
    @ none within `top-level scope`
1 ─     return $(Expr(:method, :increment_global))
)))
  2  2  │        $(Expr(:method, :increment_global))
  3  2  │   %3 = Core.Typeof(increment_global)
  4  2  │   %4 = Core.svec(%3)
  5  2  │   %5 = Core.svec()
  6  2  │   %6 = Core.svec(%4, %5, $(QuoteNode(:(#= none:2 =#))))
  7  2  │        $(Expr(:method, :increment_global, :(%6), CodeInfo(
    @ none:3 within `none`
1 ─      global i
│   %2 = i + 1
│        i = %2
└──      return %2
)))
  8  2  └──      return increment_global
Frame(mod, ex) = Frame for Main
  1* 6  1 ─ %1 = increment_global()
  2  6  └──      return %1

File: src/construct.jl [hunk 1]
```diff
@@ -436,7 +436,7 @@ function push_modex!(iter::ExprSplitter, mod::Module, ex::Expr)
         # Issue #427
         modifies_scope = false
         for a in ex.args
-            if isa(a, Expr) && a.head ∈ (:local, :global)
+            if Meta.isexpr(a, :local)
                 modifies_scope = true
                 break
             end
```
aviatesk, line 442: Probably the right fix would be to add this `ex.head === :block` condition here.
```julia
        if ex.head === :block
            for a in ex.args
                if isa(a, Expr) && a.head ∈ (:local, :global)
                    modifies_scope = true
                    break
                end
            end
        end
```
Otherwise the same problem would still happen for `:toplevel` expression with `:local` declaration:
```julia
ex = Base.parse_input_line("""
local i
i = 42
println(i)
""")
```
, but we can safely ignore those expression in `:toplevel`.

test/toplevel.jl
```diff
@@ -540,3 +540,15 @@ end
...
+    ex = :(begin
+        global foo_g = 10
+        sin(foo_g)
+    end)
+    modexs = collect(ExprSplitter(@__MODULE__, ex))
+    for (mod, ex) in modexs
+        @test JuliaInterpreter.finish!(Frame(mod, ex), true) === nothing
+    end
+    @test length(modexs) == 1
+end
```
aviatesk, line 554:
```suggestion
@testset "toplevel scope annotation" begin
    ex = Base.parse_input_line("""
    global foo_g = 10
    sin(foo_g)
    """)
    modexs = collect(ExprSplitter(@__MODULE__, ex))
    for (mod, ex) in modexs
        @test JuliaInterpreter.finish!(Frame(mod, ex), true) === nothing
    end
    @test length(modexs) == 2

    ex = Base.parse_input_line("""
    local foo = 10
    sin(42)
    """)
    modexs = collect(ExprSplitter(@__MODULE__, ex))
    for (mod, ex) in modexs
        @test JuliaInterpreter.finish!(Frame(mod, ex), true) === nothing
    end
    @test length(modexs) == 2
end
```
since the previous version passes on master.

timholy:
I can't remember either. Thanks for fixing this!

PR #512, opened 2022-01-22
Adjust to upcoming effects change

Keno:
This is for https://github.com/JuliaLang/julia/pull/43852.

Probably hold off on merging until the review phase is through,
but the PR is here for people who want to work on the branch
and need Revise to work.


Keno:
The representation in Base was changed. This will no longer be required.

Issue #435, opened 2020-09-23
ERROR: type Union has no field name


zenna:
I don't have a MWE but Debugger is failing on some of my code due to the fact that in `commands.jl`, in

`is_kw = isa(scope, Method) && comp(String(Base.unwrap_unionall(Base.unwrap_unionall(scope.sig).parameters[1]).name.name), "#kw")`

`Base.unwrap_unionall(Base.unwrap_unionall(scope.sig).parameters[1])` is a Union

```
julia> @enter x(ω)
scope = (f::Union{Distribution, Mv, Variable})(ω::Ω) where Ω<:AbstractΩ in OmegaCore.Var at /home/zenna/repos/OmegaCore.jl/src/var/dispatch.jl:14
Base.unwrap_unionall(scope.sig) = Tuple{Union{Distribution, Mv, Variable},Ω<:AbstractΩ}
(Base.unwrap_unionall(scope.sig)).parameters = svec(Union{Distribution, Mv, Variable}, Ω<:AbstractΩ)
(Base.unwrap_unionall(scope.sig)).parameters[1] = Union{Distribution, Mv, Variable}
ERROR: type Union has no field name
```

I suspect it's something to do with the fact that I'm providing a call method for a union in `f`

timholy:
Thanks for the report and the diagnosis of the problem. Can you perhaps uncomment https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/2f5f80034bc287a60fe77c4e3b5a49a087e38f8b/src/interpret.jl#L435-L436 and isolate a test case? Fixes that come with tests are oh-so-much-more-valuable than bare fixes, especially since in the long run this package is likely to be heavily refactored to execute a more efficient format than Julia's current lowered IR.

That said, if you can't trace it down, we should just fix this.

melonedo:
I have a MWE with Debugger.jl. Simply execute `s` once on `0x0 == UInt8('\t')` will trigger this issue.
```julia
julia> versioninfo()
Julia Version 1.7.1
Commit ac5cc99908 (2021-12-22 19:35 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)
Environment:
[...]

PR #513, opened 2022-01-25
Fix kwarg wrapper detection for Unions

pfitzseb:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/435.

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -221,8 +221,10 @@ function maybe_step_through_wrapper!(@nospecialize(recurse), frame::Frame)
 
     is_kw = false
     if isa(scope, Method)
-        param1 = Base.unwrap_unionall(Base.unwrap_unionall(scope.sig).parameters[1])
-        if !(param1 isa Union)
+        unwrap1 = Base.unwrap_unionall(scope.sig)
+        if isa(unwrap1, DataType)
+            param1 = Base.unwrap_unionall(unwrap1.parameters[1])
+            if isa(param1, DataType)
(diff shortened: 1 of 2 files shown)


src/commands.jl
```diff
@@ -218,7 +218,15 @@ function maybe_step_through_wrapper!(@nospecialize(recurse), frame::Frame)
     length(stmts) < 2 && return frame
     last = stmts[end-1]
     isexpr(last, :(=)) && (last = last.args[2])
-    is_kw = isa(scope, Method) && endswith(String(Base.unwrap_unionall(Base.unwrap_unionall(scope.sig).parameters[1]).name.name), "#kw")
+
+    is_kw = false
+    if isa(scope, Method)
+        param1 = Base.unwrap_unionall(Base.unwrap_unionall(scope.sig).parameters[1])
+        if !(param1 isa Union)
```
pfitzseb, line 225: Not a fan of this, but :shrug:

aviatesk, line 225: Maybe something like this would be more robust? 😂 
```suggestion
        unwrap1 = Base.unwrap_unionall(scope.sig)
        if isa(unwrap1, DataType)
            param1 = Base.unwrap_unionall(unwrap1.parameters[1])
            if isa(param1, DataType)
```

pfitzseb: Yeah I was thinking about that, but then again I wasn't sure if this can be anything else :D

aviatesk (approved):

PR #514, opened 2022-02-03
Simplify precompiles

timholy:
This switches to "do a little work" for precompilation,
rather than explicitly spelling out the specific signatures.
This should be more maintainable and potentially more
exhaustive (or could be made to be so) since it
precompiles across runtime dispatch boundaries.

There are also a couple of improvements in `step_expr!`
inference that I made while trying to track down some
invalidations. These weren't the cause, but they don't
hurt.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -457,14 +457,14 @@ function step_expr!(@nospecialize(recurse), frame, @nospecialize(node), istoplev
                     return (frame.pc = node.args[2]::Int)
                 end
             elseif node.head === :enter
-                rhs = node.args[1]
+                rhs = node.args[1]::Int
                 push!(data.exception_frames, rhs)
             elseif node.head === :leave
                 for _ = 1:node.args[1]::Int
                     pop!(data.exception_frames)
                 end
             elseif node.head === :pop_exception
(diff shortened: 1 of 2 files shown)


timholy:
The main thing to think about here is whether we're OK with `__JIInternal__` or whether a junk module is too high a price for the simplification here.

KristofferC:
> The main thing to think about here is whether we're OK with __JIInternal__ or whether a junk module is too high a price for the simplification here.

Doesn't feel that costly but is it not possible to use an "anonymous module" á la:

```
let
m = Module()
...
```

?

timholy:
That's the first thing I tried. Here's what happens:
```julia
julia> using Anon
[ Info: Precompiling Anon [412b51d8-6eb2-47c2-9dc2-53277b3423fe]
ERROR: LoadError: Evaluation into the closed module `anonymous` breaks incremental compilation because the side effects will not be permanent. This is likely due to some other module mutating `anonymous` with `eval` during precompilation - don't do this.
```

timholy:
I like that suggestion. I've updated this. It doesn't sound like there are any major concerns, so I'll merge. No need to tag immediately, so there's time to revisit this if someone objects.

PR #515, opened 2022-02-04
fix builtin detection

KristofferC:
There is a shortcut that says that any builtin/intrinsic has an empty method table but this is not true on nightly for `Core._apply_iterate`:

```
julia> typeof(Core._apply_iterate).name.mt
# 1 method for builtin function "_apply_iterate":
[1] _apply_iterate(...) in Core
```

This takes the number of failures on nightly from 9 to 1 and makes Debugger.jl pass again on nightly.

### DIFF
--- a/src/builtins.jl
+++ b/src/builtins.jl
@@ -22,8 +22,6 @@ Otherwise, return `call_expr`.
 If `expand` is true, `Core._apply` calls will be resolved as a call to the applied function.
 """
 function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
-    # By having each call appearing statically in the "switch" block below,
-    # each gets call-site optimized.
     args = call_expr.args
     nargs = length(args) - 1
     fex = args[1]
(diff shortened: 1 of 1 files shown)


KristofferC:
@aviatesk, maybe you know what changed w.r.t this?

aviatesk:
Yeah, specifically, now we have non-empty `MethodTable` after https://github.com/JuliaLang/julia/pull/43865.

src/builtins.jl
```diff
@@ -35,7 +35,8 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
     # Builtins and intrinsics have empty method tables. We can circumvent
     # a long "switch" check by looking for this.
     mt = typeof(f).name.mt
-    if isa(mt, Core.MethodTable)
+    # For some reason Core._apply_iterate does not have an empty MT
+    if isa(mt, Core.MethodTable) && f !== Core._apply_iterate
         isempty(mt) || return call_expr
     end
```
aviatesk, line 41: I think we can just replace these lines with `isa(f, Core.Builtin) || isa(f, Core.IntrinsicFunction) || return call_expr`.

Issue #517, opened 2022-02-21
Evaluating julia code using `eval_code` can be slow if the number of local variables is very large


KristofferC:
In some cases with very big functions with lots of local variables, this code in `eval_code`:

https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/5ce00482840b1a644523aac6193458e643e27aec/src/utils.jl#L606-L609

can become quite slow (on the order of 10 seconds). There might be a smarter way to do this than to generate a big expression with all the locals (and slots and SSA values) and `eval` it so that the time of execution scales with the size of the input expression and not by the size of the frame. Interpreting the expression in the given frame (non-recursively) might work.

KristofferC:
It turns out that the `let` statement in 

https://github.com/JuliaDebug/Infiltrator.jl/blob/6e91ec59ae75bea94dc0a0579bc340c839d7a3c8/src/Infiltrator.jl#L462-L493

makes this way slower than without it, probably because the block then gets compiled instead of interpreted. Without the `let` even for the big function that made me open this issue, the code evaluates in ~0.01 seconds which is acceptable. So the question is, what's a good way of shielding the evaluation in a scope but still allow the values to be GCed. Infiltrator uses a new module every time but I am not sure modules ever gets GCed so that could rack up quite a lot of memory if the function is big and one does many evaluations.

PR #518, opened 2022-02-21
do not generate code for unused SSA values, slots and local variables in `eval_code`

KristofferC:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/517

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -589,24 +589,43 @@ respectively) can be evaluated using the syntax `var"%3"` and `var"@_4"` respect
 """
 function eval_code end
 
+function extract_usage!(s::Set{Symbol}, expr)
+    if expr isa Expr
+        for arg in expr.args
+            if arg isa Symbol
+                push!(s, arg)
+            elseif arg isa Expr
+                extract_usage!(s, arg)
+            end
+        end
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -589,6 +589,15 @@ respectively) can be evaluated using the syntax `var"%3"` and `var"@_4"` respect
...
 function eval_code end
 
+function extract_usage!(s::Set{Symbol}, expr::Expr)
+    for arg in expr.args
+        extract_usage!(s, arg)
+    end
+    return s
+end
+extract_usage!(s::Set{Symbol}, sym::Symbol) = push!(s, sym)
+extract_usage!(s::Set{Symbol}, ::Any) = s
```
simeonschaub, line 599: This should avoid some dynamic dispatch:

```suggestion
function extract_usage!(s::Set{Symbol}, expr::Expr)
    for arg in expr.args
        if arg isa Symbol
            push!(s, arg)
        elseif arg isa Expr
            extract_usage!(s, arg)
        end
    end
    return s
end
```

simeonschaub (approved):
Seems reasonable to me

pfitzseb (approved):

PR #519, opened 2022-02-23
recurse in builtin expansion if we get a new builtin when expanding

KristofferC:
There are cases (arguably a bit artificial but still) where after we expand a builtin, the function that we are going to run is another builtin, for example:

```
About to run: <(Core._apply_iterate)(iterate, Core._apply_iterate, (iterate, ...)
```

Currently, we cannot step through this since the second `_apply_iterate` will call into C. Solve this by recursing into the expander in the case where the expanded function is another builtin.


aviatesk (approved):
Sounds reasonable to have this.

PR #522, opened 2022-03-07
clean up coverage files generated by test

aviatesk:
I found generated coverage files corresponding to source files a bit noisy, e.g. in a source tree view in my local editor.
I think we can just clean them up after test?

### DIFF
--- a/test/code_coverage/code_coverage.jl
+++ b/test/code_coverage/code_coverage.jl
@@ -1,5 +1,13 @@
-# delete any present coverage files
-let
+function cleanup_coverage_files()
+    # clean up coverage files for source code
+    dir, _, files = first(walkdir(normpath(@__DIR__, "..", "..", "src")))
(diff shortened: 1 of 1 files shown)


simeonschaub:
I wonder whether there's a way to get the exact extension of the coverage files generated by this specific spawned Julia session. Just might be a little bit cleaner since there might already be coverage data from previous profiles in the src folder which we don't necessarily want to delete if anyone runs the tests.

simeonschaub:
Ok, I looked into it and it seems to just use the pid of the process, so you could do something like:

```julia
p = run(`julia -E 'Libc.getpid()'`; wait=false)
pid = Libc.getpid(p)
wait(p)
```

and then only delete files ending in `.$pid.cov`.

aviatesk:
Thanks! Now I think this deletes coverage files more nicely.

Issue #526, opened 2022-03-07
Linetables are not always sorted


KristofferC:
In presence of e.g. macro expansion, line numbers in line tables are not required to be sorted. For example in

```jl
function f()
    x = 1
    @info "foo"
    x = 2
end
```

the line table looks like

```jl
julia> JuliaInterpreter.linetable(JuliaInterpreter.get_source(@which f()))
21-element Vector{Any}:
 Core.LineInfoNode(Main, :f, Symbol("REPL[6]"), 2, 0)
 Core.LineInfoNode(Main, :f, Symbol("REPL[6]"), 3, 0)
 Core.LineInfoNode(Main, Symbol("macro expansion"), Symbol("logging.jl"), 370, 2)
 Core.LineInfoNode(Main, Symbol("macro expansion"), Symbol("logging.jl"), 371, 2)
 ...
 Core.LineInfoNode(Main, :f, Symbol("REPL[6]"), 4, 0)
```

This means that the `searchsorted` at https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/19757b45fa438cf8335601099a21525a88c9bb39/src/utils.jl#L427-L429 gives garbage values.

PR #527, opened 2022-03-11
fix some cases of setting breakpoints in functions that contain macro expansions

KristofferC:
The current breakpoint code assumed:

- That all lineinfo for a method has the same file as the definition of the method.
- That the lines in the lineinfo was sorted.
- That a given `(file, line)` would only exist once in a method.

None of these are true in the presence of macro expansion. This PR makes us be a bit more careful about not doing any of the incorrect assumptions above. Now, for example, if there is a function like

```jl
function f(x)
    x += 1
    @info "..."
    x += 1
    @info "..."
    return x
end
```

and a breakpoint is added at `logging.jl:370` (where the `@info` macro is expanded) it will actually stop twice there. Also, if you happen to add a breakpoint to a method at line 370 and the method contains an `@info` statement, it will no longer stop in `logging.jl:370`.

Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/526

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -38,7 +38,9 @@ function whichtt(@nospecialize(tt))
         # for now, actual code execution doesn't ever need to consider overlayed method table
         result = Core.Compiler._findsup(tt, nothing, get_world_counter())
         result === nothing && return nothing
-        return first(result).method
+        fresult = first(result)
+        fresult === nothing && return nothing
+        return fresult.method
     else
         m = ccall(:jl_gf_invoke_lookup, Any, (Any, UInt), tt, get_world_counter())
         m === nothing && return nothing
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -372,18 +372,16 @@ function codelocation(code::CodeInfo, idx::Int)
     return codeloc
 end
 
-function compute_corrected_linerange(framecode)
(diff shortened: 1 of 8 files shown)


KristofferC:
This removes the `statementnumber` function since it is fundamentally flawed so it is technically breaking but I don't know anyone using it. @pfitzseb does VSCode use that at all?

pfitzseb:
>  does VSCode use that at all?

Nope.

src/utils.jl
```diff
@@ -370,50 +370,83 @@ function codelocation(code::CodeInfo, idx::Int)
...
+        method = framecode.scope
+        _, line1 = whereis(method)
+        offset = line1 - method.line
+        src = JuliaInterpreter.get_source(method)
+        lastline = linetable(src)[end]::LineTypes
+        return line1:getline(lastline) + offset
+    else
+        getline(linetable(framecode, 1)):getline(last(linetable(framecode)))
+    end
 end
```
pfitzseb, line 385: Can we make this multiple dispatch instead? We have some code in DebugAdapter that calls `compute_corrected_linerange` with a method argument.

KristofferC: Should be back now.

PR #450, opened 2020-12-15
re-automate builtins generations

aviatesk:
Now JuliaInterpreter will automatically pick up the latest compatible
builtins file under `src/builtins` and `include` it (with the fallback
`0.0.0` file freeze at #449).
So with this PR we no longer need to manually update builtin files under
within `src`, and can focus on just updating `bin/generate_builtins.jl`
on future changes in Julia.

What we lose with this PR is, updating already generated files later can 
be tedious; say if we have 10 more generated files under `src/builtins` 
in the future, and then manually applying something like #442 to all of 
them can be really annoying..

NOTE: to check the CI results, this PR is built on #449, so please review this after #449

PR #532, opened 2022-04-04
update builtins, cleanup pre-1.6 code paths

simeonschaub:
This makes `src/builtins.jl` in sync with `bin/generate_builtins.jl`
again. A lot of this code was also out-of-date, now that we only support
1.6, so this includes some cleanup.


aviatesk (approved):
LGTM. I'd like to ask a review from @KristofferC before merging, as I remember we decided to turn off the automated builtin file generation for good reasons (see https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/450).

Having said that, the discussion I liked above may be unrelated to this PR as this one doesn't enable the installation time generation.

KristofferC:
> I'd like to ask a review from @KristofferC before merging, as I remember we decided to turn off the automated builtin file generation for good reasons (see https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/450).

We turned off running it during compile time and having one builtin file per version but just updating the generation file seems uncontroversial. So LGTM (in principle, didn't look through carefully the exact changes but if tests pass I'm sure it's good).

bin/generate_builtins.jl
```diff
@@ -137,47 +138,79 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
...
+        end
+        return maybe_recurse_expanded_builtin(frame, new_expr)
+""")
         end
 
         id = findfirst(isequal(f), Core.Compiler.T_FFUNC_KEY)
         fcall = generate_fcall(f, Core.Compiler.T_FFUNC_VAL, id)
-        print(io,
+        if f in Core.Builtin[
+            Core._call_in_world_total, Core.donotdelete,
```
KristofferC, line 197: This requires Julia 1.9 to run the file now. Maybe that is ok.

simeonschaub: Yeah, I don't think there really is a way around that if we want the list to be up-to-date. I can maybe add a comment to that regard though

simeonschaub:
Seems like https://github.com/JuliaLang/julia/pull/44855 hasn't arrived on nightly yet, that might be part of the reason nightly CI is so slow currently

PR #533, opened 2022-04-04
fix various fma issues

### LINKED ISSUE #531: The interpreter does not use fma even if it would be used for compiled functions
```jl
julia> using JuliaInterpreter

julia> a, b, c = (1.0585073227945125, -0.00040303348596386557, 1.5051263504758005e-16)
(1.0585073227945125, -0.00040303348596386557, 1.5051263504758005e-16)

julia> muladd(a, b, c)
-0.00042661389622400056

julia> @interpret muladd(a, b, c)
-0.0004266138962240006
```

This can be a bit confusing when debugging. It would be good if the interpreter gave the same results as what it would when compiled.

I think we need to override the `Core.Intrinsics.have_fma` function:

```
julia> f() = Base.have_fma(Float64);

julia> f()
true

julia> @interpret f()
false
```

KristofferC:
Fixes #531


src/packagedef.jl
```diff
@@ -116,6 +116,15 @@ function set_compiled_methods()
     push!(compiled_modules, Base.Threads)
 end
 
+function _have_fma_compiled(T)
+    Base.Experimental.@force_compile
+    Core.Intrinsics.have_fma(T)
+end
```
simeonschaub, line 122: I think you want to explicitly force specialization here, I don't think `@force_compile` does much:
```suggestion
_have_fma_compiled(::Type{T}) where {T} = Core.Intrinsics.have_fma(T)
```

src/builtins.jl
```diff
@@ -263,6 +263,32 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
     end
     if isa(f, Core.IntrinsicFunction)
         cargs = getargs(args, frame)
+        @static if isdefined(Core.Intrinsics, :have_fma)
```
simeonschaub, line 266: Can you rebase on #532 and update `bin/generate_builtins.jl` instead?

test/interpret.jl
```diff
@@ -895,3 +895,12 @@ end
...
 end
+
+f_fma() = Base.have_fma(Float64)
+@testset "fma" begin
+    @test (@interpret f_fma()) == f_fma()
+    a, b, c = (1.0585073227945125, -0.00040303348596386557, 1.5051263504758005e-16)
+    @test (@interpret muladd(a, b, c)) === muladd(a,b,c)
+    a = 1.0883740903666346; b = 2/3
+    @test (@interpret a^b) === a^b
+end
```
simeonschaub, line 906: Will need to be inside a version check

simeonschaub (commented):
Some comments

KristofferC:
@simeonschaub, updated

PR #534, opened 2022-04-13
eval = is not functionally equivalent to global assignment

vtjnash:
Caused [crash](https://github.com/JuliaLang/julia/issues/44954) because of the incorrect semantics


simeonschaub (approved):
Thanks!

vtjnash:
I can't merge

aviatesk (approved):
It might be worth to have test cases for these code paths.

Issue #538, opened 2022-07-14
Wrong assumption about `@doc` arguments breaks Revise


mgkurtz:
When using Revise, I got the following error

    ┌ Error: Failed to revise /some/path/to/file.jl
    │   exception =
    │    BoundsError: attempt to access 3-element Vector{Any} at index [4]
    │    Stacktrace:
    │     [1] getindex
    │       @ ./array.jl:861 [inlined]
    │     [2] iterate(iter::JuliaInterpreter.ExprSplitter, state::Nothing)
    │       @ JuliaInterpreter ~/.julia/packages/JuliaInterpreter/Q7nmG/src/construct.jl:526

which was generated by https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/7fe78fa6df11e573beb682467c2f929c5892208d/src/construct.jl#L526

There it is assumed, that if `is_doc_expr(ex)` holds, `ex.args` must have length 4. But I have foolishly added a comment between my docstring and the function I documented, like

```julia
@doc "foo"
# TODO: implement this correctly ;)
foo() = 42
```

which resulted in `@doc` having only one argument and `ex.args` = `[:@doc, LineNumberNode(...), "TODO ..."]` thus has only length 3. So maybe adding this comment was foolish from me, but my follys should not break Revise from working. To settle this, perhaps `is_doc_expr` may get an extra line `length(ex.args) == 4 || return false`?

simeonschaub:
> To settle this, perhaps `is_doc_expr` may get an extra line `length(ex.args) == 4 || return false`?

This is probably reasonable either way, but that comments in the line after `@doc` parse this way seems like an upstream bug to me. Could you open an issue for this on the Julia repo?

pfitzseb:
There's a typo in the OP; the comment doesn't show up anywhere after parsing:
```
julia> Meta.parseall("""
       @doc "foo"
       # TODO: implement this correctly ;)
       foo() = 42 
       """) |> dump
Expr
  head: Symbol toplevel
  args: Array{Any}((4,))
    1: LineNumberNode
      line: Int64 1
      file: Symbol none
    2: Expr
      head: Symbol macrocall
      args: Array{Any}((3,))
        1: Symbol @doc
        2: LineNumberNode
          line: Int64 1
          file: Symbol none
        3: String "foo"
    3: LineNumberNode
      line: Int64 3
      file: Symbol none
    4: Expr
      head: Symbol =
[...]

PR #539, opened 2022-07-17
Fix `is_doc_expr` by constraining number of args

mgkurtz:
Solution as discussed in issue #538.

Without this change the first added test fails and the `collect` from the second test throws the `BoundsError` from #538 by calling `iterate(::ExprSplitter)`.


pfitzseb (approved):
lgtm.

timholy:
Thanks @mgkurtz !

PR #540, opened 2022-07-24
Re-impose "no double lookup" for invoke

### LINKED ISSUE #535: Can not step into body of `@error` because "UndefVarError: REPL[1] not defined"
MWE:

```julia
using JuliaInterpreter
f() = @error "ex"
frame = JuliaInterpreter.enter_call(f)
debug_command(frame, :c)
#=
ERROR: UndefVarError: REPL[17] not defined
Stacktrace:
 [1] collect_args(recurse::Any, frame::Frame, call_expr::Expr; isfc::Bool)
   @ JuliaInterpreter C:\Users\melonedo\.julia\packages\JuliaInterpreter\Q7nmG\src\interpret.jl:60
=#
```

I expect it to log some message instead of crashing.

The issue is identical to #441, but since I am using 0.9.13, it should have already been fixed.

timholy:
This was added in #442 and inadvertently removed in #532.
However, the test in #442 was not stringent enough; the test added here
comes from #535.

Fixes #535
Fixes https://github.com/timholy/Revise.jl/issues/695

### DIFF
--- a/bin/generate_builtins.jl
+++ b/bin/generate_builtins.jl
@@ -8,7 +8,7 @@ const RECENTLY_ADDED = Core.Builtin[
     Core.get_binding_type, Core.set_binding_type!,
     Core.getglobal, Core.setglobal!,
     Core.modifyfield!, Core.replacefield!, Core.swapfield!,
-] 
+]
 const kwinvoke = Core.kwfunc(Core.invoke)
 
 function scopedname(f)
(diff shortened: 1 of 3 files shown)


pfitzseb (approved):

simeonschaub (changes requested):
Could you update `bin/generate_builtins.jl` correspondingly, so it's not out of sync with this file?

simeonschaub (approved):

PR #541, opened 2022-07-25
Add action to check builtin consistency

pfitzseb:


KristofferC:
Why should this not be a normal test?

pfitzseb:
Because `generate_builtins.jl` only runs on ~nightly, but that fails tests currently and we'd never notice the failure like that. So now I'm running this as a separate action and as part of runtests.jl.

KristofferC (approved):

PR #545, opened 2022-08-02
NFC: normalize `x == ::Symbol` to `x === ::Symbol`

aviatesk:
It would be better to use a consistent style in our code base.
This is also aligned with the convention we're using for Julia base.


aviatesk:
All tests run green (thanks so much @simeonschaub for fixing up the nightly failure), going to merge.

PR #546, opened 2022-08-18
very slight support for `OpaqueClosure`s

oscardssmith:
this will prevent them from crashing, but will not allow stepping into them.


pfitzseb:
Will also need to be added to https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/master/bin/generate_builtins.jl.

KristofferC:
Also, a test would be good.

oscardssmith:
does this look right?

oscardssmith:
@pfitzseb thanks for fixing the broken-ness!

pfitzseb (approved):

PR #547, opened 2022-08-31
Fix invalidations from StaticArrays' `map!`

thchr:
This fixes a source of invalidations for StaticArrays when loaded together with Revise.

image

Previously, the now-annotated call site was inferred as `AbstractVector`, and the call to `mapany` then went through a completely uinferred call to `map!` (e.g., since `length(AbstractVector)` infers to `Any`).

I don't know if this is the ideal fix, but it at least seems correct and fixes the invalidation.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -727,7 +727,7 @@ function Base.StackTraces.StackFrame(frame::Frame)
     scope = scopeof(frame)
     if scope isa Method
         method = scope
-        method_args = something.(frame.framedata.locals[1:method.nargs]) :: Vector{Any}
+        method_args = [something(frame.framedata.locals[i]) for i in 1:method.nargs]
         atypes = Tuple{mapany(_Typeof, method_args)...}
         sig = method.sig
         sparams = Core.svec(frame.framedata.sparams...)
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -727,7 +727,7 @@ function Base.StackTraces.StackFrame(frame::Frame)
     scope = scopeof(frame)
     if scope isa Method
         method = scope
-        method_args = something.(frame.framedata.locals[1:method.nargs])
+        method_args = something.(frame.framedata.locals[1:method.nargs]) :: Vector{Any}
```
KristofferC, line 730:
```suggestion
        method_args = [something(x) for x in frame.framedata.locals[1:method.nargs]]
```

maybe also works to make it understand the result is a `Vector`?

thchr: That works too, yeah: done.

Also noticed that `fname` further down didn't get inferred as `Symbol`, so fixed that too - I suppose it couldn't hurt.

thchr:
Doctest failures seem related to code movement in Base. I don't know what the "builtins.jl consistency" failure is due to?

aviatesk (approved):
Nice improvements! And don't bother with the test failures -- we will fix up them soonish.

PR #548, opened 2022-09-24
eval_code: don't error on empty input.

fredrikekre:


pfitzseb (approved):

aviatesk (approved):
For absolute correctness, we may want to wrap the successfully evaluated result into
```julia
struct EvalResult
    val
    EvalResult(@nospecialize val) = new(val)
end
```
allowing it to be distinguished from the empty input case.

Issue #552, opened 2022-11-29
foreigncall handling is not semantically correct


Keno:
The foreigncall code here: https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/master/src/optimize.jl#L281-L314 is not semantically correct. It assumes that all foreigncall expressions are generated from frontend syntax and follow a predictable structure. This is not necessarily true. Cassette-style overlaying passes can modify the structure and some packages generate `:foreigncall` directly.

Keno:
Also causes https://github.com/JuliaDebug/CassetteOverlay.jl/issues/17, but that part of the code needs to be re-thought in general beyond the particular issue.

PR #554, opened 2022-12-06
simplify foreigncall handling

KristofferC:
Instead of reconstructing the surface syntax we can just pass along the arguments to a `Expr(:foreigncall, ...)`.

Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/552


aviatesk (approved):
The failures in nightly CI are known and unrelated. Thanks so much for working on this!

Issue #549, opened 2022-10-09
generate_builtins should not write to package directory


maleadt:
PkgEval currently fails to test JuliaInterpreter.jl because the package writes to its package directory (IIUC): https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/8725668c4c10e5942fc554fcc314bfa45ee6b59e/bin/generate_builtins.jl#L313

```
Check builtin.jl consistency: Error During Test at /home/pkgeval/.julia/packages/JuliaInterpreter/ScElF/test/check_builtins.jl:4
  Got exception outside of a @test
  LoadError: SystemError: opening file "/home/pkgeval/.julia/packages/JuliaInterpreter/ScElF/bin/../src/builtins.jl": Permission denied
  Stacktrace:
    [1] systemerror(p::String, errno::Int32; extrainfo::Nothing)
      @ Base ./error.jl:176
    [2] #systemerror#82
      @ ./error.jl:175 [inlined]
    [3] systemerror
      @ ./error.jl:175 [inlined]
    [4] open(fname::String; lock::Bool, read::Nothing, write::Nothing, create::Nothing, truncate::Bool, append::Nothing)
      @ Base ./iostream.jl:293
    [5] open(fname::String, mode::String; lock::Bool)
      @ Base ./iostream.jl:356
    [6] open(fname::String, mode::String)
      @ Base ./iostream.jl:355
    [7] open(::var"#3#4", ::String, ::Vararg{String}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ Base ./io.jl:393
    [8] open
      @ ./io.jl:392 [inlined]
    [9] generate_builtins(file::String)
      @ Main ~/.julia/packages/JuliaInterpreter/ScElF/bin/generate_builtins.jl:83
   [10] top-level scope
      @ ~/.julia/packages/JuliaInterpreter/ScElF/bin/generate_builtins.jl:313
   [11] include(fname::String)
      @ Base.MainInclude ./client.jl:478
   [12] macro expansion
      @ ~/.julia/packages/JuliaInterpreter/ScElF/test/check_builtins.jl:7 [inlined]
```

JuliaInterpreter.jl should probably use Scratch.jl (if these files are required for execution), or a temporary directory (if it's only for testing).

PR #559, opened 2022-12-29
Make generate_builtins path customizable for testing. Update builtins

IanButterworth:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/549
Also updates builtins

@maleadt

### DIFF
--- a/bin/generate_builtins.jl
+++ b/bin/generate_builtins.jl
@@ -8,7 +8,8 @@ const RECENTLY_ADDED = Core.Builtin[
     Core.get_binding_type, Core.set_binding_type!,
     Core.getglobal, Core.setglobal!,
     Core.modifyfield!, Core.replacefield!, Core.swapfield!,
-    Core.finalizer
+    Core.finalizer, Core._compute_sparams, Core._svec_ref,
+    Core.compilerbarrier
 ]
 const kwinvoke = Core.kwfunc(Core.invoke)
 
--- a/bin/generate_builtins.jl
+++ b/bin/generate_builtins.jl
@@ -317,5 +317,5 @@ end
 """)
 end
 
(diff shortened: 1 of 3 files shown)


test/check_builtins.jl
```diff
@@ -5,8 +5,11 @@ using Test, DeepDiffs
     @testset "Check builtin.jl consistency" begin
         builtins_path = joinpath(@__DIR__, "..", "src", "builtins.jl")
         old_builtins = read(builtins_path, String)
-        include("../bin/generate_builtins.jl")
-        new_builtins = read(builtins_path, String)
+        new_builtins_path = tempname()
```
maleadt, line 8: Maybe it's better to use `mktempdir()` and `$BUILTINS_DIR/builtins.jl` instead?

maleadt:
LGTM, thanks!

maleadt:
I'm not a member of JuliaDebug either. cc @KristofferC

KristofferC:
This seems to fail everywhere :P. I think there might be some `@static isdefined` checks put into the builtin-generator?

IanButterworth:
Spent too long there trying to engineer what `RECENTLY_ADDED` does. Should really read files from the top-down

bin/generate_builtins.jl
```diff
@@ -316,4 +317,5 @@ end
 """)
 end
 
-generate_builtins(joinpath(@__DIR__, "..", "src", "builtins.jl"))
+builtins_dir = get(ENV, "BUILTINS_DIR", joinpath(@__DIR__, "..", "src"))
```
KristofferC, line 320: Include `JULIAINTERPRETER`? Or `JULIA_JULIAINTERPRETER_BUILTINS_DIR` maybe.

PR #564, opened 2023-02-04
fix one test failure on nightly

aviatesk:


aviatesk:
There seems to be unrelated segfault also...

test/interpret.jl
```diff
@@ -856,25 +860,25 @@ end
...
-    ci = code_typed(foo, NTuple{2, Int}; optimize=false)[][1]
-    mi = Core.Compiler.method_instances(foo, NTuple{2, Int})[]
-
-    frameargs = Any[foo, 1, 2]
-    framecode = JuliaInterpreter.FrameCode(mi.def, ci)
-    frame = JuliaInterpreter.prepare_frame(framecode, frameargs, mi.sparam_vals)
-    @test JuliaInterpreter.finish_and_return!(frame) === 8
-end
+# @testset "TypedSlots" begin
+#     function foo(x, y)
```
KristofferC, line 864: Could you explain why this is commented out?

aviatesk: This testset causes segfault on nightly. Just wanted to check if there are other test failures.

PR #565, opened 2023-02-06
Protect `error` calls from invalidation

timholy:
These get called in the absence of knowledge about the second input,
so protecting them with `invokelatest` avoids invalidation risk.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -99,7 +99,7 @@ function lookup_or_eval(@nospecialize(recurse), frame, @nospecialize(node))
             elseif f === Base.getproperty
                 return Base.getproperty(ex.args[2], ex.args[3])
             else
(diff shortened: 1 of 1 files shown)


src/interpret.jl
```diff
@@ -99,7 +99,7 @@ function lookup_or_eval(@nospecialize(recurse), frame, @nospecialize(node))
             elseif f === Base.getproperty
                 return Base.getproperty(ex.args[2], ex.args[3])
             else
-                error("unknown call f ", f)
+                Base.invokelatest(error, "unknown call f ", f)
```
aviatesk, line 102: If we are on Julia 1.8 and higher, we can do something like this alternatively:
```suggestion
                error(lazy"unknown call f $f")
```

aviatesk (approved):

PR #566, opened 2023-02-17
adjust to JuliaLang/julia#48693

aviatesk:


oscardssmith:
mentioning Slot cleanup for searchability.

PR #568, opened 2023-03-24
Fix kw pattern matching, other changes on 1.9+

timholy:
With the introduction of Core.kwcall, we need to adjust our
pattern-matching for the statements that prepare to dispatch on
kw-handling functions.

There are still a few broken tests on 1.9+, but this fixes the majority.


timholy:
The scary segfault doesn't happen on `backports-release-1.9`. Perhaps among others, "fix incorrect static parameter definedness check (#49097)" is critical here.

timholy:
I went ahead and fixed the remaining 4 failures (on nightly), so we can be sure they're not related to any of the kwprep changes.

aviatesk:
Thanks for updating this!

PR #579, opened 2023-08-10
Update some failing doctests

timholy:
The main change seems to be the `@` notation.

I was motivated by decreasing annoying noise, cf https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/575#issuecomment-1672030152

### DIFF
--- a/.github/workflows/Documenter.yml
+++ b/.github/workflows/Documenter.yml
@@ -10,6 +10,10 @@ jobs:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
+      - uses: julia-actions/setup-julia@v1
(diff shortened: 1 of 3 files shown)


timholy:
[1ecb156](https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/579/commits/1ecb156096b45ad3daf09037032c495cb61e7fb5) is a little weird: on my machine it prints as `d` (disabled) but the right answer surely is `c`. I've left it `c` in the docs but added a filter so it isn't picky. Feel free to suggest an alternative.

timholy:
Seems good now.

PR #581, opened 2023-08-11
Ignore `:aliasscope` and `:popaliasscope`

### LINKED ISSUE #574: Error only while using the debugger in VSCode
hello - I'm able to run a script in VSCode, but get an error only while using the VSCode debugger - I have shared the details along with a reproducer on the Julia VSCode extension repo here:

https://github.com/julia-vscode/julia-vscode/issues/3349

and one of the contributors was kind enough to look into the error and has ID'ed the root cause as an issue with JuliaInterpreter - the stack trace of the error are in the comment here:

https://github.com/julia-vscode/julia-vscode/issues/3349#issuecomment-1641634133

it'd help us out greatly if we could get some feedback on resolving the issue - thank you so much!

timholy:
Fixes #574

@vchuravy, is just ignoring it the correct thing to do? It's just a compiler annotation?

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -401,7 +401,8 @@ function eval_rhs(@nospecialize(recurse), frame, node::Expr)
     elseif head === :boundscheck
         return true
     elseif head === :meta || head === :inbounds || head === :loopinfo ||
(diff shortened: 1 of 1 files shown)


vchuravy:
Yes it's okay to ignore

PR #583, opened 2023-08-19
align arguments number in breakpoints hook docstring

disberd:
Simply update the first line of the docstring to specify that the hook function must accept two arguments as explained further down in the docstring


timholy:
Thanks!

PR #584, opened 2023-08-25
adjust tests for latest Julia master

aviatesk:


aviatesk:
Tests are failing due to sketchy segfaults, but it looks like they are happening on master too. I would like to move forward this for now and try to fix the segfault in a separate PR.

KristofferC:
Sounds good

PR #585, opened 2023-08-25
adjust to JuliaLang/julia#50943

aviatesk:


aviatesk:
I confirmed the tests are passing on nightly on my local machine (mac with apple m2 silicon).

PR #587, opened 2023-09-03
Add usage documentation

ShalokShalom:


timholy:
This seems OK, but it does duplicate the docs. I go back and forth in how I feel about that. Other views?

ShalokShalom:
> This seems OK, but it does duplicate the docs. I go back and forth in how I feel about that. Other views?

I felt the same. 

To be honest, the docs seem pretty good. 

The Readme on the other hand, seems like this app is not maintained and got its last update in 1960. 

So I thought I pretty it up a bit. 

But if you like to keep it short and tidy, I would recommend placing the documentation link a bit obvious. 

It might be just me, but I really thought this is not documented at all first, only containing the instruction to install it, and that's it. 

I mean, there are projects like this, so I was quick to jump to that conclusion.

PR #592, opened 2023-10-09
Adjust to upcoming julia lowering change

Keno:
See https://github.com/JuliaLang/julia/pull/51590.


vtjnash (approved):

aviatesk:
We've observed the intermittent errors in the v1 and nightly CI, which we're aware of but haven't investigated into yet. Given this, I think we can just proceed with merging this PR.

aviatesk (approved):

PR #594, opened 2023-10-19
investigate segfaults on CI

simeonschaub:
The segfaults on CI seem to point to these two tests in particular.
Let's see if disabling them fixes CI or whether that might have just
been a red herring


simeonschaub:
Well, that certainly answers the question... 😅

aviatesk:
😱

simeonschaub:
Hmm, at least nightly doesn't segfault anymore, the tests just fail... :thinking:

aviatesk:
How would you like to proceed with this issue?  I'm of the view that for healthy development, it's acceptable to disable tests that are challenging to correct at the moment. My preference is to focus on restoring the CI as a first step.

simeonschaub:
Yeah, unfortunately this PR sh

aviatesk:
Yeah, I will take a look on it later today.

PR #596, opened 2023-11-25
adapt to array changes in Julia base

simeonschaub:
https://github.com/JuliaLang/julia/pull/51319 added and removed a lot of builtins. Adapt `generate_builtins` accordingly

### DIFF
--- a/bin/generate_builtins.jl
+++ b/bin/generate_builtins.jl
@@ -9,8 +9,13 @@ const RECENTLY_ADDED = Core.Builtin[
     Core.getglobal, Core.setglobal!,
     Core.modifyfield!, Core.replacefield!, Core.swapfield!,
     Core.finalizer, Core._compute_sparams, Core._svec_ref,
-    Core.compilerbarrier
+    Core.compilerbarrier,
+    Core.memoryref, Core.memoryref_isassigned, Core.memoryrefget, Core.memoryrefoffset, Core.memoryrefset!,
 ]
+# Builtins present in 1.6, not builtins (potentially still normal functions) anymore
+const RECENTLY_REMOVED = GlobalRef.(Ref(Core), [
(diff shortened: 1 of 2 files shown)


simeonschaub:
CI is a bit of a :wastebasket::fire: rn, so review with care :grinning:

bin/generate_builtins.jl
```diff
@@ -246,6 +250,24 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
...
+        elseif name === :arrayset
+            maxarg = 6
+        elseif name === :arraysize
+            maxarg = 2
+        end
+        fcall = generate_fcall_nargs(name, minarg, maxarg)
+        name = repr(name)
+        print(io,
+"""
+    elseif @static isdefined($mod, $name) && f === $name
```
aviatesk, line 267:
```suggestion
        rname = repr(name)
        print(io,
"""
    elseif @static isdefined($mod, $rname) && f === $name
```
? Otherwise `f === :arrayref` does never seem to be true?

aviatesk: Also, they are now generic functions defined within `Core`, so e.g. `isdefined(Core, :arrayref)` will be `true` on the latest master. Don't we want to have `Core.arrayref isa Core.Builtin` instead?

simeonschaub: > Also, they are now generic functions defined within Core, so e.g. isdefined(Core, :arrayref) will be true on the latest master. Don't we want to have Core.arrayref isa Core.Builtin instead?

My thought process was that even though they're not technically builtins anymore there isn't really any point in stepping into them and we might as well evaluate them eagerly

aviatesk: That sounds fine. Let's move this forward as is.

PR #598, opened 2023-11-28
remove old compats

aviatesk:
The version compat is >=1.6, so we can expect `ReturnNode` and `GotoIfNot` are defined in `Core` always.
This should be good to go as far as 1.6 CI passes.


aviatesk:
Going to merge as the 1.6 CI went successfully on this branch.

PR #599, opened 2023-11-28
adjust to the `:enter` IR changes made in JuliaLang/julia#52300

aviatesk:
I don't expect this commit to fix the CI on :fire:, but since these changes are straightforward, we should be able to go ahead and merge this as is.


simeonschaub (approved):
LGTM

PR #600, opened 2023-11-28
follow up #596

aviatesk:


simeonschaub (approved):
Oops! 😄

bin/generate_builtins.jl
```diff
@@ -260,11 +260,12 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
...
             maxarg = 2
         end
-        fcall = generate_fcall_nargs(name, minarg, maxarg)
+        _scopedname = "$mod.$name"
+        fcall = generate_fcall_nargs(_scopedname, minarg, maxarg)
         rname = repr(name)
         print(io,
 """
-    elseif @static isdefined($mod, $rname) && f === $name
+    elseif @static (isdefined($mod, $rname) && $_scopedname isa Core.Builtin) && f === $_scopedname
```
aviatesk, line 268: It turns out that we never hit this path on the recent versions since we exclude the generic cases above (https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/9d50726a3aa25d6267825a4ec8ea19c6a5e7a8a2/src/builtins.jl#L46).

This change is necessary to make this script functional on the previous versions.

PR #604, opened 2023-12-12
Remove buggy linearization pass

vtjnash:
There is only one bug in the base linearization pass that needs to be handled explicitly, otherwise linearization is fully guaranteed for IR.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -9,22 +9,6 @@ function lookup_var(frame, slot::SlotNumber)
     throw(UndefVarError(frame.framecode.src.slotnames[slot.id]))
 end
 
-function lookup_expr(frame, e::Expr)
-    head = e.head
-    head === :the_exception && return frame.framedata.last_exception[]
-    if head === :static_parameter
-        arg = e.args[1]::Int
-        if isassigned(frame.framedata.sparams, arg)
-            return frame.framedata.sparams[arg]
-        else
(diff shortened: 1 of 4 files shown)


src/interpret.jl
```diff
@@ -91,18 +99,22 @@ function lookup_or_eval(@nospecialize(recurse), frame, @nospecialize(node))
         if ex.head === :call
             f = ex.args[1]
             if f === Core.svec
-                return Core.svec(ex.args[2:end]...)
+                return f(popfirst!(ex.args)...)
```
aviatesk, line 102:
```suggestion
                popfirst!(ex.args)
                return f(ex.args...)
```

src/interpret.jl
```diff
@@ -91,18 +99,22 @@ function lookup_or_eval(@nospecialize(recurse), frame, @nospecialize(node))
...
             elseif f === Core.apply_type
-                return Core.apply_type(ex.args[2:end]...)
-            elseif f === Core.typeof
-                return Core.typeof(ex.args[2])
-            elseif f === Base.getproperty
-                return Base.getproperty(ex.args[2], ex.args[3])
+                return f(popfirst!(ex.args)...)
```
aviatesk, line 104:
```suggestion
                popfirst!(ex.args)
                return f(ex.args...)
```

aviatesk (approved):
Added some fixes and cleanups. This is now ready to go.

Issue #591, opened 2023-10-09
JuliaInterpreter model of :pop_exception is incorrect


Keno:
JuliaInterpreter treats `:pop_exception`, as another spelling of `:leave`, but this is a misunderstanding. `:pop_exception` operates on the stack of exceptions (i.e. an exception gets pushed on the stack when we take a `catch` edge), not the stack of exception handlers

https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/a3cf18e0997f79d9caaa79242c314157b6b80578/src/interpret.jl#L490-L491

KristofferC:
Should be fixed

PR #605, opened 2023-12-22
implement support for `current_scope`

aviatesk:


Keno:
Seems reasonable, but it feels like there needs to be something that actually reflects the shadow copy of the current scope here into the runtime upon calling a native function. Not sure if that's here or in Debugger.jl

aviatesk:
This PR impacts the nightly CI only, although the nightly CI are all failing due to unrelated issues to this PR, primarily due to issues #610 and possibly #591.
Despite the failures, the tests added in this PR have been successfully executed on my local nightly build. Considering this PR also resolves the `Check builtins.jl` CI issue, I'd like to move forward with it.

PR #606, opened 2024-01-03
changes to adapt to compressed line table format

vtjnash:
This contains some of the updates needed to keep working with the optimized debuginfo format (https://github.com/JuliaLang/julia/pull/52415). This is a draft, since someone needs integrate this to maintain support for old versions, where the old code is mostly marked here with my initials (jwn) and my initials should be removed entirely and instead the implementation in those places fixed.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -258,7 +258,11 @@ end
 
 # These getters improve inference since fieldtype(CodeInfo, :linetable)
 # and fieldtype(CodeInfo, :codelocs) are both Any
-const LineTypes = Union{LineNumberNode,Core.LineInfoNode}
+@static if VERSION ≥ v"1.12.0-DEV.173"
+    const LineTypes = Union{LineNumberNode,Base.IRShow.LineInfoNode}
+else
+    const LineTypes = Union{LineNumberNode,Core.LineInfoNode}
+end
 function linetable(arg)
     if isa(arg, Frame)
         arg = arg.framecode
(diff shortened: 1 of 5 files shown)


aviatesk:
It looks like this "optimization" introduces a very bad interaction with the new lineinfo format: https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/28dd37ce0f7bcb357bfb820b1d6b53b98aa3796c/src/optimize.jl#L76-L92
```julia
julia> using JuliaInterpreter

julia> fnone1() = nothing;

julia> let m = only(methods(fnone1))
           JuliaInterpreter.FrameCode(m, JuliaInterpreter.get_source(m); optimize=true).src.debuginfo
       end
Core.DebugInfo(Symbol("REPL[2]"), nothing, svec(), "\x01\0\0\0\0\0\0\0\0\0\0\0\x01\x01\0")

julia> fnone2() = nothing;

julia> let m = only(methods(fnone2))
           JuliaInterpreter.FrameCode(m, JuliaInterpreter.get_source(m); optimize=true).src.debuginfo
       end
Core.DebugInfo(Symbol("REPL[4]"), nothing, svec(), "")
```

src/utils.jl
```diff
@@ -258,26 +258,74 @@ end
...
 function linetable(arg, i::Integer; macro_caller::Bool=false)::Union{Expr,LineTypes}
     lt = linetable(arg)
-    lineinfo = _linetable(lt, i)
-    if macro_caller
-        while lineinfo isa Core.LineInfoNode && lineinfo.method === Symbol("macro expansion") && lineinfo.inlined_at != 0
-            lineinfo = _linetable(lt, lineinfo.inlined_at)
+    @static if VERSION ≥ v"1.12.0-DEV.173"
+        # TODO: decode the linetable at this frame efficiently by reimplementing this here
+        # TODO: get the contextual name from the parent, rather than returning "n/a" (which breaks Cthulhu)
+        return Base.IRShow.buildLineInfoNode(lt, :var"n/a", i)[1] # ignore all inlining / macro expansion / etc :(
```
aviatesk, line 286: I don't understand this commit -- do we now store some macro expansion context within `debuginfo`?

src/utils.jl
```diff
@@ -258,38 +258,88 @@ end
...
+    lastline = 0
+    for k = 0:typemax(Int)
+        codeloc = Core.Compiler.getdebugidx(debuginfo, k)
+        line::Int = codeloc[1]
+        line < 0 && break
+        lastline = max(lastline, line)
+    end
+    return lastline
+end
+function codelocs(arg, i::Integer)
```
aviatesk, line 315: For `statementnumbers` to function in v1.12, there's a need for a tool that can rebuild the full `codelocs` object from `Core.DebugInfo`, similar to what we could do just by looking at `src.codelocs` previously. Or, we may have to overhaul the `statementnumbers` algorithm.

aviatesk:
This PR is obviously not finished yet, but I want to push ahead with it as is to avoid holding up progress on other packages in the ecosystem. We can tackle the specific issues that I've identified so far later on (likely when Jameson returns).

PR #607, opened 2024-01-05
update docs and convert `jldoctest` to `@repl` blocks

aviatesk:
Doctest is very fragile.
Using `@repl`, we can update documentation automatically.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+docs/Manifest.toml linguist-generated=true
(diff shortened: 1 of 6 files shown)


aviatesk:
It looks like https://github.com/JuliaDebug/JuliaInterpreter.jl/actions/runs/7418164310/job/20185719270?pr=607 saying our `DOCUMENTER_KEY` isn't setup correctly? The doc deployment is actually failing on master too.

timholy:
I updated DOCUMENTER_KEY earlier today, should be OK now.

PR #609, opened 2024-01-07
Fix revise#718

### LINKED ISSUE #427: more careful container expression flattening ?
[Container expression flattening logic in `ExprSplitter`](https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/7aef26ff8b247d361f4b5705bf1497f5c360fb65/src/construct.jl#L503-L521) can split block expression in wrong way, e.g.

```julia
ex = :(begin
    local foo = 10
    sin(foo)
end)
for (mod, ex) in ExprSplitter(@__MODULE__, ex)
    finish!(Frame(mod, ex), true)
end
```
```
ERROR: UndefVarError: foo not defined
Stacktrace:
  [1] collect_args(frame::JuliaInterpreter.Frame, call_expr::Expr; isfc::Bool)
    @ JuliaInterpreter ~/julia/packages/JuliaInterpreter/src/interpret.jl:60
  [2] collect_args
    @ ~/julia/packages/JuliaInterpreter/src/interpret.jl:135 [inlined]
  [3] evaluate_call_recurse!(recurse::Any, frame::JuliaInterpreter.Frame, call_expr::Expr; enter_generated::Bool)
    @ JuliaInterpreter ~/julia/packages/JuliaInterpreter/src/interpret.jl:208
  [4] evaluate_call_recurse!
    @ ~/julia/packages/JuliaInterpreter/src/interpret.jl:202 [inlined]
  [5] eval_rhs(recurse::Any, frame::JuliaInterpreter.Frame, node::Expr)
    @ JuliaInterpreter ~/julia/packages/JuliaInterpreter/src/interpret.jl:389
  [6] step_expr!(recurse::Any, frame::JuliaInterpreter.Frame, node::Any, istoplevel::Bool)
    @ JuliaInterpreter ~/julia/packages/JuliaInterpreter/src/interpret.jl:527
  [7] step_expr!
    @ ~/julia/packages/JuliaInterpreter/src/interpret.jl:582 [inlined]
  [8] finish!(recurse::Any, frame::JuliaInterpreter.Frame, istoplevel::Bool)
    @ JuliaInterpreter ~/julia/packages/JuliaInterpreter/src/commands.jl:14
  [9] finish!
    @ ~/julia/packages/JuliaInterpreter/src/commands.jl:19 [inlined]
 [10] top-level scope
    @ ./none:2
```

In this specific example, the expression will be split into `local foo = 10` and `sin(foo)`, and in the second expression `foo` will just be resolved as (implicitly global referring) `Symbol` rather than slots, which leads interpret to fail.

timholy:
This is a very minimal fix, reasoning that `global` annotations are common at toplevel but `local` ones are not. When I fixed #427, there was no good reason to avoid splitting on `global`, I don't know why I did that.

We might someday want to rewrite `ExprSplitter` more extensively, but that would likely be a breaking change.

Fixes https://github.com/timholy/Revise.jl/issues/718
Also adds nested docstring test

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -338,10 +338,13 @@ end
 """
     ExprSplitter(mod::Module, ex::Expr; lnn=nothing)
 
-Create an iterable that returns individual expressions together with their module of evaluation.
+Given a module `mod` and a top-level expression `ex` in `mod`, create an iterable that returns
+individual expressions together with their module of evaluation.
 Optionally supply an initial `LineNumberNode` `lnn`.
 
 # Example
+
+In a fresh session,
 
 ```
 julia> expr = quote
@@ -375,13 +378,13 @@ mod = Main
 ex = :($(Expr(:toplevel, :(#= REPL[7]:6 =#), :(const threshold = 0.1))))
 ```
 
-Note that `Main.Private` was created for you so that its internal expressions could be evaluated.
+`ExprSplitter` created `Main.Private` was created for you so that its internal expressions could be evaluated.
 `ExprSplitter` will check to see whether the module already exists and if so return it rather than
 try to create a new module with the same name.
 
 In general each returned expression is a block with two parts: a `LineNumberNode` followed by a single expression.
 In some cases the returned expression may be `:toplevel`, as shown in the `const` declaration,
-but otherwise it will be a `:block`.
+but otherwise it will preserve its parent's `head` (e.g., `expr.head`).
 
 # World age, frame creation, and evaluation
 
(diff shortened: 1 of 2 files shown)


timholy:
Wow, segfault or `ReadOnlyMemoryError` on all three Julia 1.10s. It's common locally, too, although if I run the tests enough times I get this to pass. Clearly we're going to have to fix the segfaults. I've never gotten `rr` working on my WSL instance so I can't promise it will be soon.

aviatesk (approved):
Yeah that segmentation fault is known, as it's occurring in the master too. I've attempted to fix it, but without success. Since it's not necessary to resolve it in this PR, I thinh we can go ahead without addressing it here.

timholy:
I notice there are a bunch of commits since the most recent release. Many of these commits are recent, suggesting lots of development effort lately. For that reason I won't tag a new release (I'm uncertain if we're midstream in a bunch of improvements), but certainly a new release will be welcome whenever it's a good time.

timholy:
Aww, Revise is all green and pretty again (except nightly). Thanks for the release!

PR #611, opened 2024-02-18
Adjust to upcoming compiler change

Keno:
The `.inferred` field is being removed in [1], so delete the assertion.

[1] https://github.com/JuliaLang/julia/pull/53219

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -68,7 +68,6 @@ function optimize!(code::CodeInfo, scope)
     mod = moduleof(scope)
     evalmod = mod == Core.Compiler ? Core.Compiler : CompiledCalls
     sparams = scope isa Method ? sparam_syms(scope) : Symbol[]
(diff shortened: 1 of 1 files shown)


src/optimize.jl
```diff
@@ -68,7 +68,6 @@ function optimize!(code::CodeInfo, scope)
     mod = moduleof(scope)
     evalmod = mod == Core.Compiler ? Core.Compiler : CompiledCalls
     sparams = scope isa Method ? sparam_syms(scope) : Symbol[]
-    code.inferred && error("optimization of inferred code not implemented")
```
aviatesk, line 71: I wondered if we can replace the condition of this assertion with this one:
```julia
    code.slottypes !== nothing && error("optimization of inferred code not implemented")
```

aviatesk: And the same for LoweredCodeUtils.jl too.

Issue #610, opened 2024-01-09
JuliaInterpreter hangs when handling errors on v1.11


aviatesk:
MRE:
```julia
julia> function func()
           i::Int = 0
           try
               undefcall()
           finally
               i += 1
               println("finally calause: ", i)
           end
       end

julia> @interpret func()
...
finally calause: 2461
finally calause: 2462
finally calause: 2463
finally calause: 2464
finally calause: 2465
finally calause: 2466
finally calause: 2467
finally calause: 2468
finally calause: 2469
finally calause: 2470
finally calause: 2471
...
```

We need to update our interpretation logic against JuliaLang/julia#52245.

PR #614, opened 2024-03-07
adapt to implicit leave change in Julia

vtjnash:
Fixes https://github.com/JuliaDebug/JuliaInterpreter.jl/issues/610
Refs https://github.com/JuliaLang/julia/pull/52245


JeffBezanson (approved):

aviatesk (approved):
LGTM. There are a few other CI failures, but it looks like this commit fixes the issue at least. Let's move forward with this as it would help us debug the other issues.

PR #615, opened 2024-03-11
some updates required for Julia v1.11

vtjnash:
not yet passing, but fixes many of the issues

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,9 +23,6 @@ jobs:
           - version: '~1.11.0-0' # next release
             os: ubuntu-latest
             arch: x64
-          - version: '1.6' # old LTS
-            os: ubuntu-latest
-            arch: x64
           - version: 'nightly' # dev
             os: ubuntu-latest
             arch: x64
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,17 +13,34 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version:
-          - '1.6'     # LTS
-          - '1'       # current stable
-          - '1.11-nightly' # next stable
-          - 'nightly'
-        os:
-          - ubuntu-latest
-          - macOS-latest
-          - windows-latest
-        arch:
-          - x64
+        include:
+          - version: '1.6' # old LTS
+            os: ubuntu-latest
+            arch: x64
+          - version: '1.10' # current stable
+            os: ubuntu-latest
+            arch: x64
+          - version: '~1.11.0-0' # next release
+            os: ubuntu-latest
+            arch: x64
+          - version: '1.6' # old LTS
+            os: ubuntu-latest
+            arch: x64
+          - version: 'nightly' # dev
+            os: ubuntu-latest
+            arch: x64
+          #- version: '1' # x86 ubuntu -- disabled since PyCall/conda is broken on this platform
+          #  os: ubuntu-latest
+          #  arch: x86
+          - version: '1' # x86 windows
+            os: windows-latest
+            arch: x86
+          - version: '1' # x64 windows
+            os: windows-latest
+            arch: x64
+          - version: '1' # x64 macOS
+            os: macos-latest
+            arch: x64
     env:
       PYTHON: ""
     steps:
--- a/test/debug.jl
+++ b/test/debug.jl
@@ -426,7 +426,8 @@ end
 
         frame = JuliaInterpreter.enter_call(sort, a)
         frame = stepkw!(frame)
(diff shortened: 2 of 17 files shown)


.github/workflows/CI.yml
```diff
@@ -13,16 +13,34 @@ jobs:
...
+          #  arch: x86
+          - version: '1' # x86 windows
+            os: windows-latest
+            arch: x86
+          - version: '1' # x64 windows
+            os: windows-latest
+            arch: x64
+          - version: '1' # x64 macOS
+            os: macos-latest
+            arch: x64
```
aviatesk, line 43: Keeping this list updated and maintained seems to be a real hassle. Can we go back to how things were before with using `matrix`? Adding `1.11-nightly` just for testing the next release temporarily makes sense to me. And if we need to disable certain platforms, we could use:
```yml
       exclude:
         - version: '1' # x86 ubuntu -- disable since PyCall/conda is broken on this platform
           os: ubuntu-latest
           arch: x86
```
should do the trick.

test/interpret.jl
```diff
@@ -775,29 +783,14 @@ end
...
-            quote
-                unsafe_store!(world_counter, regular_world)
-            end
-        ))
-    end
+@noinline foobar() = (GC.safepoint(); 42)
+function run_foobar()
+    @eval foobar() = "nope"
+    @test_broken @interpret(foobar()) == foobar() == 42
+    nothing
```
aviatesk, line 790:
```suggestion
    return @interpret(foobar()), foobar()
```
and add `@test_broken run_foobar() == (42, 42) TODO #617` in the `"unreachable worlds"` testset.
Since it would be easier to debug things by running `run_foobar`.

test/interpret.jl
```diff
@@ -951,7 +944,7 @@ end
     @test @interpret (Base.Experimental.@opaque x->3*x)(4) == 12
 end
 
-# CassetteOverlay, issue #552
+## CassetteOverlay, issue #552
```
aviatesk, line 947:
```suggestion
# CassetteOverlay, issue #552
```

.github/workflows/CI.yml
vtjnash: The only part of the include list to update is the number for latest stable and next release, which is the same regardless of whether it is a matrix or list. I like a list more, since we can set the list of platforms that are actually interesting sparsely selecting interesting variations on the matrix, rather than running all versions for all platforms for all OS

vtjnash: It is basically a build "T"ee instead of a Matrix. One platform runs tests in depth (ubuntu x64) and other platforms run the tests in breadth ("1.x")

test/interpret.jl
```diff
@@ -775,29 +783,14 @@ end
...
+    @test_broken @interpret(foobar()) == foobar() == 42
+    nothing
 end
-
 @testset "unreachable worlds" begin
-    foobar() = 42
-    @unreachable foobar() = "nope"
-
-    @test @interpret(foobar()) == foobar()
+    run_foobar()
```
vtjnash, line 793:
```suggestion
    @test run_foobar()
```

test/interpret.jl
```diff
@@ -775,29 +783,13 @@ end
...
+    @eval foobar() = "nope"
+    return @interpret(foobar()), foobar()
 end
-
 @testset "unreachable worlds" begin
-    foobar() = 42
-    @unreachable foobar() = "nope"
-
-    @test @interpret(foobar()) == foobar()
+    @test run_foobar()
```
vtjnash, line 792:
```suggestion
    @test ==(run_foobar()...)
```

aviatesk:
I'm trying to fix the remaining test failures by #619 (hopefully).

aviatesk:
CI is green on all versions except the nightly 🎉

test/debug.jl
```diff
@@ -427,7 +426,12 @@ end
 
         frame = JuliaInterpreter.enter_call(sort, a)
         frame = stepkw!(frame)
-        @test frame.pc == JuliaInterpreter.nstatements(frame.framecode) - 1
+        @static if VERSION ≥ v"1.7"
+            # TODO fix this broken test (@aviatesk)
```
aviatesk, line 430: This test case is broken on `≥v"1.11-"`. We need to revisit the implementation of `maybe_step_through_kwprep!`
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/eae07f9c6f8155f2d1edb7638834856da732f466/src/commands.jl#L273-L278

timholy: Since in principle one can hand-write all these statements oneself, the real resolution will presumably come once we have JuliaLowering and we can go back to the source code to see how it was written.

PR #619, opened 2024-03-20
add support for preset `current_scope`

aviatesk:
There are cases where we invoke JuliaInterpreter from a context with some scope is set (e.g. `@test_logs (...) (@interpret ...)`).


src/construct.jl
```diff
@@ -287,11 +287,15 @@ function prepare_framedata(framecode, argvals::Vector{Any}, lenv::SimpleVector=e
         ssavalues = Vector{Any}(undef, ng)
         sparams = Vector{Any}(undef, 0)
         exception_frames = Int[]
-        current_scopes = Any[]
+        current_scopes = Scope[]
         last_reference = Vector{Int}(undef, ns)
         callargs = Any[]
         last_exception = Ref{Any}(_INACTIVE_EXCEPTION.instance)
     end
+    @static if isdefined(Base, :current_scope)
```
aviatesk, line 295:
```suggestion
    @static if isdefined(Core, :current_scope)
```

PR #634, opened 2024-06-20
Don't error if we don't have line information

Keno:
`whereis` is already allowed to return `nothing`. It's possible for codeloc to be != 0, but for Core.Compiler to still be unable to produce a location. In that case, also return `nothing`.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -277,12 +277,14 @@ function linetable(arg)
     return ci.linetable::Union{Vector{Core.LineInfoNode},Vector{Any}} # issue #264
     end # @static if
 end
-function linetable(arg, i::Integer; macro_caller::Bool=false)::Union{Expr,LineTypes}
+function linetable(arg, i::Integer; macro_caller::Bool=false)::Union{Expr,Nothing,LineTypes}
     lt = linetable(arg)
     @static if VERSION ≥ v"1.12.0-DEV.173"
     # TODO: decode the linetable at this frame efficiently by reimplementing this here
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -277,12 +277,14 @@ function linetable(arg)
     return ci.linetable::Union{Vector{Core.LineInfoNode},Vector{Any}} # issue #264
     end # @static if
 end
-function linetable(arg, i::Integer; macro_caller::Bool=false)::Union{Expr,LineTypes}
+function linetable(arg, i::Integer; macro_caller::Bool=false)::Union{Expr,Nothing,LineTypes}
```
aviatesk, line 280: `linetable(arg, ::Integer)` is used in the other places too, like `linenumber(::FrameCode, pc)`, so better to handle the `nothing` case there too?

Keno: I guess the other option would be to synthesize a special LineNumber node that indicates the absence of information. That way clients that don't care don't have to worry about it

timholy: I like the `nothing` option since anything that isn't prepared will hard-error.

timholy:
I've finished off the `nothing` case (I think), please review.

Also, it sure would be nice to have a test! I know they can be hard to capture, but there's a real sense in which JuliaInterpreter/Revise/etc work only because of their test suites. If we don't keep capturing that we become vulnerable to regressions.

We also shouldn't be too optimistic that Revise will "work" once this is merged and a new LoweredCodeUtils is released; JuliaInterpreter tests still fail, and thus it's very likely that Revise will fail on certain files as a consequence.

timholy:
Thanks for the fix!

PR #635, opened 2024-07-05
More 1.12 compat

Keno:
Fixes most of the semantic issues on 1.12 - there's still some test failures related to source locations, but should be usable for day-to-day things at least.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -538,14 +538,13 @@ function step_expr!(@nospecialize(recurse), frame, @nospecialize(node), istoplev
                     error("this should have been handled by split_expressions")
                 elseif node.head === :using || node.head === :import || node.head === :export
                     Core.eval(moduleof(frame), node)
-                elseif node.head === :const
+                elseif node.head === :const || node.head === :globaldecl
                     g = node.args[1]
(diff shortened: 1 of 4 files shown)


src/interpret.jl
```diff
@@ -538,14 +538,18 @@ function step_expr!(@nospecialize(recurse), frame, @nospecialize(node), istoplev
                     error("this should have been handled by split_expressions")
                 elseif node.head === :using || node.head === :import || node.head === :export
                     Core.eval(moduleof(frame), node)
-                elseif node.head === :const
+                elseif node.head === :const || node.head === :globaldecl
                     g = node.args[1]
                     if isa(g, GlobalRef)
                         mod, name = g.mod, g.name
                     else
                         mod, name = moduleof(frame), g::Symbol
```
aviatesk, line 546: Looks like now we don't need this `mod` variable? Assuming these `GlobalRef` modules are always same as `moduleof(frame)`.

aviatesk (approved):

src/interpret.jl
Keno: We just just remove this block and use `g` directly, since we're falling back to the base interpreter anyway.

PR #645, opened 2024-09-10
restrict the signature types of method with the `recurse` argument

aviatesk:
These routines may be overloaded by the new `recurse` type, and in such cases, it is safer to restrict the method signature as much as possible (since the overloads may use `@invoke` to dispatch to the original method to reuse the original implementation).


simeonschaub (approved):
Sounds reasonable

PR #649, opened 2024-09-29
fix interpretation of opaque closures on nightly

simeonschaub:
This is the best way I could think of to determine whether an OC
contains inferred code after JuliaLang/julia#53219. Also deletes a
fastpath we had for OCs containing inferred code in
`maybe_evaluate_builtin` as it doesn't really seem necessary and we
would otherwise scan the IR twice for OCs we do want to interpret.


timholy (approved):

KristofferC:
Should be done

PR #654, opened 2024-11-14
bump the minimum supported Julia version to 1.10 (LTS)

aviatesk:
Maintaining this package that closely depends on Julia internals across multiple Julia versions is very cumbersome. Now that the LTS has been updated to 1.10, let's drop compatibility with earlier versions and simplify the code.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -14,13 +14,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - version: '1.6' # old LTS
+          - version: '1' # current stable
             os: ubuntu-latest
             arch: x64
-          - version: '1.10' # current stable
+          - version: '1.10' # lowerest version supported
             os: ubuntu-latest
             arch: x64
-          - version: '~1.11.0-0' # next release
+          - version: '1.12-nightly' # next release
             os: ubuntu-latest
             arch: x64
           - version: 'nightly' # dev
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ jobs:
           - version: '1' # current stable
             os: ubuntu-latest
             arch: x64
-          - version: '1.10' # lowerest version supported
+          - version: '1.10' # lowest version supported
             os: ubuntu-latest
             arch: x64
           - version: '1.12-nightly' # next release
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JuliaInterpreter"
 uuid = "aa1ae85d-cabe-5617-a682-6adf51b2e16a"
-version = "0.9.37"
+version = "0.10.0"
 
 [deps]
 CodeTracking = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
@@ -10,7 +10,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 CodeTracking = "0.5.9, 1"
-julia = "1.6"
+julia = "1.10"
 
 [extras]
 CassetteOverlay = "d78b62d4-37fa-4a6f-acd8-2f19986eb9ee"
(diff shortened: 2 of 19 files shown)


src/builtins.jl
```diff
@@ -152,11 +152,7 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
             return Some{Any}(Core.finalizer(getargs(args, frame)...))
         end
     elseif @static isdefined(Core, :get_binding_type) && f === Core.get_binding_type
```
simeonschaub, line 154: This `@static` can be removed as well

src/builtins.jl
```diff
@@ -256,13 +252,7 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
             return Some{Any}(getfield(getargs(args, frame)...))
         end
     elseif @static isdefined(Core, :getglobal) && f === getglobal
```
simeonschaub, line 254: Same here

src/builtins.jl
```diff
@@ -152,11 +152,7 @@ function maybe_evaluate_builtin(frame, call_expr, expand::Bool)
             return Some{Any}(Core.finalizer(getargs(args, frame)...))
         end
     elseif @static isdefined(Core, :get_binding_type) && f === Core.get_binding_type
-        if nargs == 2
-            return Some{Any}(Core.get_binding_type(@lookup(frame, args[2]), @lookup(frame, args[3])))
-        else
-            return Some{Any}(Core.get_binding_type(getargs(args, frame)...))
-        end
+        return Some{Any}(Core.get_binding_type(getargs(args, frame)...))
```
simeonschaub, line 155: I thought splatting here was bad for inference, which is why `n == 2` was special cased before?

simeonschaub (commented):
Overall SGTM, just some comments

src/builtins.jl
aviatesk: Now the compiler has the special casings for these builtin functions and it doesn't register any tfuncs for them. Maybe we can recover tfuncs that are simplified version of the special cases.

bin/generate_builtins.jl
```diff
@@ -2,16 +2,26 @@
...
-# All builtins present in 1.6
-const ALWAYS_PRESENT = Core.Builtin[
-    (<:), (===), Core._abstracttype, Core._apply_iterate, Core._apply_pure,
-    Core._call_in_world, Core._call_latest, Core._equiv_typedef, Core._expr,
-    Core._primitivetype, Core._setsuper!, Core._structtype, Core._typebody!,
-    Core._typevar, Core.apply_type, Core.ifelse, Core.sizeof, Core.svec,
-    applicable, fieldtype, getfield, invoke, isa, isdefined, nfields,
-    setfield!, throw, tuple, typeassert, typeof
+# All builtins present in 1.10
+const RECENTLY_ADDED = Core.Builtin[
```
simeonschaub, line 6: Why call this `RECENTLY_ADDED` instead of `ALWAYS_PRESENT`?

.github/workflows/CI.yml
```diff
@@ -14,13 +14,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - version: '1.6' # old LTS
+          - version: '1' # current stable
             os: ubuntu-latest
             arch: x64
-          - version: '1.10' # current stable
+          - version: '1.10' # lowerest version supported
```
KristofferC, line 20:
```suggestion
          - version: '1.10' # lowest version supported
```

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "JuliaInterpreter"
 uuid = "aa1ae85d-cabe-5617-a682-6adf51b2e16a"
-version = "0.9.37"
+version = "0.10.0"
```
KristofferC, line 3: Doesn't have to be a breaking release?

bin/generate_builtins.jl
aviatesk: The comment above was wrong. This list now contains builtins that have been added after 1.10.

Project.toml
aviatesk: As far as I recall, bumping the supported Julia version typically involves at least a minor version bump (based on my experience with Cthulhu.jl). But is that not always the case?

PR #657, opened 2025-01-19
Lookup bindings in the latest world

Keno:
Post-partitioning the executing world age matters for bindings lookup. JuliaInterpreter currently does not model world age correctly. That's a bigger issue. However, for now, just look up bindings in the latest world age, which mostly restores the unpartitioned behavior.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -2,7 +2,7 @@ isassign(frame::Frame) = isassign(frame, frame.pc)
 isassign(frame::Frame, pc::Int) = (pc in frame.framecode.used)
 
 lookup_var(frame::Frame, val::SSAValue) = frame.framedata.ssavalues[val.id]
-lookup_var(frame::Frame, ref::GlobalRef) = getfield(ref.mod, ref.name)
+lookup_var(frame::Frame, ref::GlobalRef) = invokelatest(getfield, ref.mod, ref.name)
 function lookup_var(frame::Frame, slot::SlotNumber)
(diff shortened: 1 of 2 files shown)


aviatesk (approved):
If `invokelatest` isn’t specified, in which world do these functions perform binding lookups? 

Not just in JuliaInterpreter, but for packages using `getglobal`, I assumed the default lookup world for `getglobal` would be set to the latest as I believed it would then be non-breaking.

Keno:
> If `invokelatest` isn’t specified, in which world do these functions perform binding lookups?

The current TLS age of the executing task.

> Not just in JuliaInterpreter, but for packages using getglobal, I assumed the default lookup world for getglobal would be set to the latest as I believed it would then be non-breaking.

No, those cannot be the semantics as all global accesses would then have to be inferred to `Any` in case some other task replaced the binding under your feet after you compiled it.

KristofferC:
This was a quite significant performance hit. 

Before

```
julia> function mysum(x)
           s = 0
           for v in x
               s += v
           end
           return s
       end;

# Before
julia> @time @interpret mysum(fill(1, 100000))
  0.566310 seconds (10.40 M allocations: 186.914 MiB, 1.29% gc time)
100000

# After
julia> @time @interpret mysum(fill(1, 100000))
  0.912419 seconds (16.20 M allocations: 280.050 MiB, 0.57% gc time)
100000
```

anyway to do this in a bit better way?

PR #658, opened 2025-02-05
More 1.12 world age adjustments

Keno:


src/construct.jl
```diff
@@ -479,8 +479,8 @@ function queuenext!(iter::ExprSplitter)
     if head === :module
         # Find or create the module
         newname = ex.args[2]::Symbol
-        if isdefined(mod, newname)
-            newmod = getfield(mod, newname)
+        if invokelatest(isdefined, mod, newname)
+            newmod = invokelatest(getfield, mod, newname)
```
aviatesk, line 483: What's the difference between `invokelatest` vs. `@invokelatest` here?

Keno: `invokelatest` will be marginally faster since it doesn't need to re-lookup the `getfield` binding.

Keno: Although we're still missing the optimization to make `invokelatest` on intrinsics fast, so it doesn't really make a difference right now.

PR #659, opened 2025-02-06
Correct logic for ccall symbol lookup

Keno:
The existing logic could not distinguish between ccall(foo) and ccall(:foo) and was giving wrong errors as a result.


aviatesk:
LGTM

PR #664, opened 2025-03-17
Fix builtins

timholy:
Compensates for the accidental push of
aca346cf36874a591b284c23b4adc1b194271001


timholy:
The Julia 1 failure is windows-only and appears to be in a section marked `if Sys.islinux()`, which is strange, but it does not seem related to my recent work. Since `master` is currently broken, I'll merge this now.

PR #667, opened 2025-03-31
Fix kwprep stepping in Julia 1.12 and 1.11

### LINKED ISSUE #620: fix the broken test case in `@testset "Stepping over kwfunc preparation"`
This test case is broken on `≥v"1.11-"`.
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/eae07f9c6f8155f2d1edb7638834856da732f466/test/debug.jl#L431

We need to revisit the implementation of `maybe_step_through_kwprep!`
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/eae07f9c6f8155f2d1edb7638834856da732f466/src/commands.jl#L273-L278

_Originally posted by @aviatesk in https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/615#discussion_r1533547783_

timholy:
Closes #620

(So glad I get to do this again! :laughing: )

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -273,21 +273,26 @@ function maybe_step_through_kwprep!(@nospecialize(recurse), frame::Frame, istopl
     pc, src = frame.pc, frame.framecode.src
     n = length(src.code)
     stmt = pc_expr(frame, pc)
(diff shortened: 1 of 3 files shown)


timholy:
While I know this is frankencode, it may be worth getting a review on this one (at least for the obvious stuff).

PR #671, opened 2025-04-01
Fix world age issues in test/toplevel

timholy:
With these changes we can get much of the way through the toplevel tests. This is good, because Revise is critically dependent on toplevel handling. (Despite Revise tests passing on Julia 1.12, it often fails on complex code bases.)

There are remaining world-age issues in the Non-frames test (and maybe others). My sense is that probably the only realistic way to get everything working well is to provide full support for world age in JuliaInterpreter.

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -18,9 +18,9 @@ function lookup_stmt(stmts::Vector{Any}, @nospecialize arg)
         if isa(q, QuoteNode) && isa(q.value, Symbol)
             mod = lookup_stmt(stmts, arg.args[2])
             if isa(mod, GlobalRef)
-                mod = @invokelatest getproperty(mod.mod, mod.name)
+                mod = @invokelatest getglobal(mod.mod, mod.name)
             end
-            isa(mod, Module) && return @invokelatest getproperty(mod, q.value)
+            isa(mod, Module) && return @invokelatest getglobal(mod, q.value)
         end
     end
     return arg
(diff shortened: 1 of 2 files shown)


src/optimize.jl
```diff
@@ -18,9 +18,9 @@ function lookup_stmt(stmts::Vector{Any}, @nospecialize arg)
         if isa(q, QuoteNode) && isa(q.value, Symbol)
             mod = lookup_stmt(stmts, arg.args[2])
             if isa(mod, GlobalRef)
-                mod = getproperty(mod.mod, mod.name)
+                mod = @invokelatest getproperty(mod.mod, mod.name)
```
aviatesk, line 21:
```suggestion
                mod = @invokelatest getglobal(mod.mod, mod.name)
```

src/optimize.jl
```diff
@@ -18,9 +18,9 @@ function lookup_stmt(stmts::Vector{Any}, @nospecialize arg)
...
             end
-            isa(mod, Module) && return getproperty(mod, q.value)
+            isa(mod, Module) && return @invokelatest getproperty(mod, q.value)
```
aviatesk, line 23:
```suggestion
            isa(mod, Module) && return @invokelatest getglobal(mod, q.value)
```
I don't think this change will significantly improve performance, but since we've already raised the minimum compatibility to v1.10, there's no longer a need to call `getproperty`.

Issue #666, opened 2025-03-31
`@interpret(x == 2)` and `@interpret(x) == 2` lead to different execution world ages


timholy:
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/22f186bf2fcca6ef6b5570770035011441cb3bba/test/interpret.jl#L416-L420 is broken on Julia 1.12 and can be fixed by changing the test to `@test @interpret(g_gf()) == 2`. As much sense as I think that change makes, I decided to try to understand why this happens and only appeared now. It turns out that the error is thrown at https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/22f186bf2fcca6ef6b5570770035011441cb3bba/src/construct.jl#L753, where we try to evaluate the arguments passed to the call, and for the current test the outer call is `==(g_gf(), 2)`. Thus it tries to `Core.eval(Main, :(g_gf()))`, and this fails for reasons of world age.

Fundamentally this seems to be testing that JuliaInterpreter *doesn't* respect world ages properly. Should we ditch this test?

KristofferC:
(I deleted my earlier reply because I posted it too early, sorry).

KristofferC:
Maybe the test can be changed to

```julia
q = 0

function h()
    eval(:(q = 2))
    return q
end

@test @interpret(h() == 2)
```

?

KristofferC:
I think @aviatesk has said that this whole optimization of resolving values early is invalid anyway though..

timholy:
Yes, it doesn't seem to respect world ages. That may be more important with the bindings work in 1.12. Unfortunately, getting this fixed is probably not a small change, and the performance cost will likely be significant.

timholy:
In an ideal world what we'd have is a whole invalidation framework for JuliaInterpreter now: resolve the bindings in the right world age, but when the world age updates you have to check for invalidations. Ugh.

timholy:
So the easier thing is to boot out the optimization and get things at the time you need them (in the right world age).

timholy:
> I think [@aviatesk](https://github.com/aviatesk) has said that this whole optimization of resolving values early is invalid anyway though..

It's also been deactivated (which is partly why so much is broken)

https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/822632d6f77323d338cf1e6765d1d6d5e10421c0/src/optimize.jl#L43

timholy:
[...]

PR #680, opened 2025-04-15
Support external method tables in method evaluation

serenity4:
Currently, method evaluation assumes a syntax of the form
```julia
Expr(:method, :foo, %sig, %body)
```
whereas methods defined in an external method table will have the form
```julia
Expr(:method, mt, %sig, %body)
```

where `mt` references a `Core.MethodTable`, with the function directly extracted from the signature. This form assumes that the function is already defined, so we may rely on having `typeof(foo)` defined and provided as first argument. This slight difference simplifies the evaluation logic as now implemented in `evaluate_overlayed_methoddef`.

We used to evaluate both in the same way, leading to errors such as raised in https://github.com/timholy/Revise.jl/issues/646. While further changes are required to properly revise methods defined in external method tables, this PR at least allows us to interpret any necessary overlay definitions.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,8 +27,7 @@ Mmap = "a63ad114-7e13-5084-954f-fe012c677804"
 PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
 SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-Tensors = "48a634ad-e948-5137-8d70-aa71f2a747f4"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["CassetteOverlay", "DataFrames", "Dates", "DeepDiffs", "Distributed", "FunctionWrappers", "HTTP", "LinearAlgebra", "Logging", "LoopVectorization", "Mmap", "PyCall", "SHA", "SparseArrays", "Tensors", "Test"]
+test = ["CassetteOverlay", "DataFrames", "Dates", "DeepDiffs", "Distributed", "FunctionWrappers", "HTTP", "LinearAlgebra", "Logging", "LoopVectorization", "Mmap", "PyCall", "SHA", "SparseArrays", "Test"]
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JuliaInterpreter"
 uuid = "aa1ae85d-cabe-5617-a682-6adf51b2e16a"
-version = "0.9.44"
+version = "0.9.45"
 
 [deps]
 CodeTracking = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -359,7 +359,7 @@ function extract_method_table(frame::Frame, node::Expr; eval = true)
     end
     mod, name = isa(arg, Symbol) ? (moduleof(frame), arg) : (arg.mod, arg.name)
     @invokelatest(isdefinedglobal(mod, name)) || return nothing
-    value = Core.eval(mod, name)
+    value = getglobal(mod, name)
     isa(value, MethodTable) && return value
     return nothing
 end
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -334,17 +334,20 @@ function evaluate_methoddef(frame::Frame, node::Expr)
     length(node.args) == 1 && return f
     sig = @lookup(frame, node.args[2])::SimpleVector
     body = @lookup(frame, node.args[3])::Union{CodeInfo, Expr}
-    ccall(:jl_method_def, Cvoid, (Any, Ptr{Cvoid}, Any, Any), sig, C_NULL, body, moduleof(frame)::Module)
-    return f
+    RT = method_def_return_type()
+    return ccall(:jl_method_def, Any, (Any, Ptr{Cvoid}, Any, Any), sig, C_NULL, body, moduleof(frame)::Module)::RT
 end
+
+# see https://github.com/JuliaLang/julia/pull/58076
+method_def_return_type() = @static VERSION ≥ v"1.13.0-DEV.388" ? Method : Nothing
 
 function evaluate_overlayed_methoddef(frame::Frame, node::Expr, mt::Core.MethodTable)
     # Overlaying an empty function such as `function f end` is not legal, and `f` must
     # already be defined so we don't need to do as much work as in `evaluate_methoddef`.
     sig = @lookup(frame, node.args[2])::SimpleVector
     body = @lookup(frame, node.args[3])::Union{CodeInfo, Expr}
-    ccall(:jl_method_def, Cvoid, (Any, Any, Any, Any), sig, mt, body, moduleof(frame)::Module)
-    return mt
+    RT = method_def_return_type()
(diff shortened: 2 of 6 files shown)


Project.toml
```diff
@@ -27,8 +27,7 @@ Mmap = "a63ad114-7e13-5084-954f-fe012c677804"
 PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
 SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-Tensors = "48a634ad-e948-5137-8d70-aa71f2a747f4"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["CassetteOverlay", "DataFrames", "Dates", "DeepDiffs", "Distributed", "FunctionWrappers", "HTTP", "LinearAlgebra", "Logging", "LoopVectorization", "Mmap", "PyCall", "SHA", "SparseArrays", "Tensors", "Test"]
+test = ["CassetteOverlay", "DataFrames", "Dates", "DeepDiffs", "Distributed", "FunctionWrappers", "HTTP", "LinearAlgebra", "Logging", "LoopVectorization", "Mmap", "PyCall", "SHA", "SparseArrays", "Test"]
```
timholy, line 33: See below re "vendoring"

src/interpret.jl
```diff
@@ -332,9 +334,34 @@ function evaluate_methoddef(frame::Frame, node::Expr)
...
+    isa(arg, MethodTable) && return arg
+    if !isa(arg, Symbol) && !isa(arg, GlobalRef)
+        eval || return nothing
+        value = Core.eval(moduleof(frame), arg)
+        isa(value, MethodTable) && return value
+        return nothing
+    end
+    mod, name = isa(arg, Symbol) ? (moduleof(frame), arg) : (arg.mod, arg.name)
+    @invokelatest(isdefined(mod, name)) || return nothing
+    value = Core.eval(mod, name)
```
timholy, line 362: `getglobal`?

src/utils.jl
```diff
@@ -26,16 +26,16 @@ function to_function(@nospecialize(x))
...
+    method = whichtt(tt, mt = nothing)
 
 Like `which` except it operates on the complete tuple-type `tt`,
 and doesn't throw when there is no matching method.
 """
-function whichtt(@nospecialize(tt))
+function whichtt(@nospecialize(tt), mt::Union{Nothing, MethodTable} = nothing)
     # TODO: provide explicit control over world age? In case we ever need to call "old" methods.
     # branch on https://github.com/JuliaLang/julia/pull/44515
     # for now, actual code execution doesn't ever need to consider overlayed method table
```
timholy, line 37: Seems like this line of the comment could be deleted?

test/interpret.jl
```diff
@@ -415,7 +415,7 @@ end
 
     function g_gf()
         eval(:(z = 2))
-        return z
+        return @invokelatest(@__MODULE__().z)
```
timholy, line 418: This test should either change like you suggest, switch to `@test_throws`, or go away. Currently it's testing that JuliaInterpreter is broken wrt world age, xref #666

FYI the `teh/worldage` branch is a start on proper support for world age, but it's on hold for now while I work on other package updates for 1.12.

test/interpret.jl
```diff
@@ -646,16 +645,16 @@ module VecTest
             Core.getfield(Base, :llvmcall)($exp, Vec{$N, $T}, Tuple{Vec{$N, $T}, Vec{$N, $T}}, x, y)
         end
     end
-    f() = 1.0 * one(Tensor{2,3})
+    f(a) = vecadd(a, a)
```
timholy, line 648: If this "vendors" the implementation in Tensors, I'm all in favor. CC @KristofferC to check whether there was any more specific intent for the original version.

(We should vendor more of the tests and depend less on external packages.)

timholy (commented):
Looks good, just minor nits

src/utils.jl
aviatesk: Yeah, JuliaInterpreter hasn't implemented the infrastructure to run the overlay method yet, but it's just not implemented yet, and it's easy to implement. I'm planning to implement it for the use case in JET.

src/interpret.jl
```diff
@@ -332,9 +334,34 @@ function evaluate_methoddef(frame::Frame, node::Expr)
...
+    arg = node.args[1]
+    isa(arg, MethodTable) && return arg
+    if !isa(arg, Symbol) && !isa(arg, GlobalRef)
+        eval || return nothing
+        value = Core.eval(moduleof(frame), arg)
+        isa(value, MethodTable) && return value
+        return nothing
+    end
+    mod, name = isa(arg, Symbol) ? (moduleof(frame), arg) : (arg.mod, arg.name)
+    @invokelatest(isdefined(mod, name)) || return nothing
```
aviatesk, line 361:
```suggestion
    @invokelatest(isdefinedglobal(mod, name)) || return nothing
```
If you rebase onto the latest `master`, you can use `isdefinedglobal` within the JuliaInterpreter module regardless of the version.

src/interpret.jl
```diff
@@ -332,9 +334,34 @@ function evaluate_methoddef(frame::Frame, node::Expr)
...
+
+function extract_method_table(frame::Frame, node::Expr; eval = true)
+    isexpr(node, :method, 3) || return nothing
+    arg = node.args[1]
+    isa(arg, MethodTable) && return arg
+    if !isa(arg, Symbol) && !isa(arg, GlobalRef)
+        eval || return nothing
+        value = Core.eval(moduleof(frame), arg)
+        isa(value, MethodTable) && return value
+        return nothing
```
aviatesk, line 358:
```suggestion
        try
            return Core.eval(moduleof(frame), arg)
        catch
            return nothing
        end
```

src/utils.jl
aviatesk: xref: https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/682 (You don't need this at all to complete this PR. Just for reference).

test/interpret.jl
KristofferC: I don't really recall the original intent. I guess best way would be to check where the test was intorudced and see that the new one also fails/succeeds in the same way.

src/interpret.jl
serenity4: I'll still keep the `isa(value, MethodTable)` guard, but we should indeed wrap the `eval` in a `try`/`catch`.

src/utils.jl
serenity4: I updated the comment to link to that PR, which can then be deleted there.

aviatesk (approved):

src/interpret.jl
```diff
@@ -332,9 +334,34 @@ function evaluate_methoddef(frame::Frame, node::Expr)
...
+    isa(arg, MethodTable) && return arg
+    if !isa(arg, Symbol) && !isa(arg, GlobalRef)
+        eval || return nothing
+        value = try Core.eval(moduleof(frame), arg) catch _ nothing end
+        isa(value, MethodTable) && return value
+        return nothing
+    end
+    mod, name = isa(arg, Symbol) ? (moduleof(frame), arg) : (arg.mod, arg.name)
+    @invokelatest(isdefinedglobal(mod, name)) || return nothing
+    value = getglobal(mod, name)
```
aviatesk, line 362:
```suggestion
    value = @invokelatest getglobal(mod, name)
```
Maybe we should `@invokelatest` here too?

PR #682, opened 2025-04-19
implement a system for interpreting overlay methods

aviatesk:
Implemented, being inspired by JuliaDebug/JuliaInterpreter.jl#680, thinking it might be useful in the near future for JET's use case.

The overlay interpretertation is enabled by overloading `JuliaInterpreter.method_table(recurse)`, but maybe the entire `recurse`-overload mechanism itself needs to be aligned with the `AbstractInterpreter` design and properly organized.

Since this interface will also be needed for JET,
I'll probably work on it soon.


aviatesk:
This is now built on top of JuliaDebug/JuliaInterpreter.jl#683.

PR #683, opened 2025-04-21
change the `recurse` interface to `AbstractInterpreter`-like interface

aviatesk:
Align the `recurse` argument to something like the base Compiler's `AbstractInterpreter` and make JuliaInterpreter routines overloadable properly.

This change is quite breaking (thus bumping the minor version of this package), but necessary to enhance the customizability of JI. For example, it will make it easier to add changes like JuliaDebug/JuliaInterpreter.jl#682 in a nicer way, but also should enable better designs in packages such as Revise and JET.


aviatesk:
If I have one concern about this PR, which is that the change from the old argument type declaration `@nospecialize(recurse)` to `interp::Interpreter` in this PR might lead to different code specialization, potentially impacting Revise's latency.

However it might be positive, as `interp::Interpreter` should allow for simpler code specialization than `@nospecialize(recurse)`. And now that package precompilation is possible, `interp::Interpreter` might actually achieve better latency.

aviatesk:
I tried using Revise on Example.jl with https://github.com/timholy/Revise.jl/pull/911 locally, and it seems like it have slightly improved the latency. I'm not sure how to accurately measure the latency of the First-Time-To-Revise, so it could be just a margin of error.

aviatesk:
This PR is pretty big, but I've been using it locally since yesterday and haven't run into any issues. So, if there are no major objections, I plan to merge it tomorrow.

@timholy Sorry to ping you again, but if you have a moment, it would be great to get your thoughts, even just on the general idea.
That said, I'll take responsibility for handling any issues that might come up from this refactor. So, if you don't have specific feedback, please feel free to ignore this.

Regarding the main concern about the impact on Revise's latency, this PR (and the related Revise PR) actually seems to improve latency slightly. So, I think we're okay to move forward with this direction regarding this aspect.

src/types.jl
```diff
@@ -1,9 +1,33 @@
...
+"""
+abstract type Interpreter end
+
 """
-struct Compiled end
+    RecursiveInterpreter <: Interpreter
+
+`RecursiveInterpreter` is an [`Interpreter`](@ref) that recursively interprets any `:call`
+expressions in the code being interpreted.
+"""
```
serenity4, line 16: I'm not 100% sure that this suggestion is correct, but if it is the case it might help unfamiliar readers such as myself. Feel free to accept or ignore.
```suggestion
`RecursiveInterpreter` is an [`Interpreter`](@ref) that recursively interprets any `:call`
expressions in the code being interpreted.

With this interpreter, code runs in fully interpreted mode; it will never be compiled for execution.
"""
```

src/types.jl
```diff
@@ -1,9 +1,33 @@
...
+`NonRecursiveInterpreter` is an [`Interpreter`](@ref) that evaluates any `:call` expressions
+in the code being interpreted using Julia's normal code execution engine with the native
+compiler.
+
+`JuliaInterpreter.Compiled` is alised to `NonRecursiveInterpreter` for backward compatibility.
+"""
+struct NonRecursiveInterpreter <: Interpreter end
+
+const Compiled = NonRecursiveInterpreter # for backward compatibility
 Base.similar(::Compiled, sz) = Compiled()  # to support similar(stack, 0)
```
serenity4, line 31:
```suggestion
Base.similar(::NonRecursiveInterpreter, sz) = NonRecursiveInterpreter()  # to support similar(stack, 0)
```

serenity4 (approved):
That's a very nice refactor! The type annotations are also very welcome.

I made a few (very minor) suggestions, but I think the design is great. There remain uses of `Compiled` in the codebase, perhaps those should be replaced to `NonRecursiveInterpreter` so we only ever use `Compiled` for defining the alias and as export (but never use that name internally)?

src/types.jl
aviatesk: This additional comment would be very helpful:)

src/construct.jl
```diff
@@ -762,11 +744,50 @@ macro interpret(arg)
...
+        optname isa Symbol || error("Invalid @interpret call: $optname is not a symbol")
+        if optname === :interp
+            return interpret(mod, xs...; interp=esc(optval), kwargs...)
+        else
+            error("Invalid @interpret call: $optname is not a recognized option")
+        end
+    else
+        error("Invalid @interpret call: $opt is not a keyword")
+    end
+end
```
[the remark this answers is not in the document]
aviatesk: Julia can distinguish these method signatures, so it doesn't actually cause a method overwrite warning. I also tried the combined approach, but I feel the current implementation is cleaner, so I'd like to stick with it.

aviatesk:
> There remain uses of `Compiled` in the codebase, perhaps those should be replaced to `NonRecursiveInterpreter` so we only ever use `Compiled` for defining the alias and as export (but never use that name internally)?

I reviewed the places where `Compiled` is being used throughout the codebase and replaced it with `NonRecursiveInterpreter` where appropriate.
I found that in files like `localmethtable.jl` and `optimize.jl`, `Compiled` is used more like a token rather than an actual interpreter. For these situations, it seems better to use a separate token type that isn't a subtype of `Interpreter`.
I plan to do this refactoring later, so for now, I've left the usage of `Compiled` unchanged in those specific places.

src/types.jl
aviatesk: This method is required for the `Compiled` usages where it is used as a special token indicating somethings need to be compiled, so I'd like to leave this method definition as is for now.

aviatesk:
@serenity4 Thanks for your reviews! Very much appreciated.

timholy:
Sorry I missed this before merging! I'm very supportive of this general idea. With the hopes that JuliaLowering will become a thing in the forseeable future, Revise probably won't even need to depend on JuliaInterpreter forever. So things that advance the state of the art in JuliaInterpreter should not be held back.

PR #689, opened 2025-07-22
avoid `args` getting boxed in `maybe_evaluate_builtin`

KristofferC:
😮‍💨 🥊 

```julia 
julia> function mysum(x)
           s = 0
           for v in x
               s += v
           end
           return s
       end;

# master
julia> @time @interpret mysum(fill(1, 100000))
  1.198966 seconds (27.70 M allocations: 527.277 MiB, 0.41% gc time)
100000

# PR
julia> @time @interpret mysum(fill(1, 100000))
  0.958389 seconds (20.80 M allocations: 417.413 MiB, 0.45% gc time)
100000
```

### DIFF
--- a/bin/generate_builtins.jl
+++ b/bin/generate_builtins.jl
@@ -210,7 +210,9 @@ function maybe_evaluate_builtin(interp::Interpreter, frame::Frame, call_expr::Ex
             print(io,
 """
     $head f === tuple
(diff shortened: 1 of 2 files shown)


aviatesk (approved):
We really need to make the compiler able to optimize this pattern of code..

PR #700, opened 2025-09-17
make dynamic-ccall detection more correct

vtjnash:
Relies on https://github.com/JuliaLang/julia/pull/59165 for exact
accuracy, otherwise we're relying on hoping that `eval` on a random
expression is safe and correct (both hopes are quite false).

### DIFF
--- a/src/optimize.jl
+++ b/src/optimize.jl
@@ -2,6 +2,7 @@ const compiled_calls = Dict{Any,Any}()
 
 # Pre-frame-construction lookup
 function lookup_stmt(stmts::Vector{Any}, @nospecialize arg)
+    # this convert a statement into something else, without the slightly interest in correctness:
     if isa(arg, SSAValue)
         arg = stmts[arg.id]
     end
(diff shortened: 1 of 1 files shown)


vtjnash:
merge?

src/optimize.jl
```diff
@@ -2,6 +2,7 @@ const compiled_calls = Dict{Any,Any}()
 
 # Pre-frame-construction lookup
 function lookup_stmt(stmts::Vector{Any}, @nospecialize arg)
+    # this convert a statement into something else, without the slightly interest in correctness:
```
timholy, line 5: 😆 
```suggestion
    # this converts a statement into something else, without the slightest interest in correctness:
```
I remember this one being horrible. It's just enough to squeak by tests.

timholy (commented):
A test that doesn't work currently would be lovely. JuliaInterpreter has s tendency to break each new Julia release, the tests are just about the only thing that prevent the whole thing from burning to the ground.

Honestly, now my preference would be to stop trying to interpret anything with a ccall, but keeping this working for now is also fine.

src/optimize.jl
vtjnash: yes, the comment is mostly as a warning against someone using it for new purposes

vtjnash:
This is mainly about supporting https://github.com/JuliaLang/julia/pull/59165, which removes the need for the unreliable hacks since they become syntactic instead of heuristic operations. Without this, it would start failing all tests once that is merged

PR #715, opened 2026-06-05
Thread world age through interpretation

timholy:
Master carried an inert `Frame.world` field, but dispatch and global lookups still ran at the latest world via `invokelatest` / `get_world_counter()`. This captures a world at frame construction and uses it consistently, so stepping sees a fixed world even if new code is defined mid-session.

`default_world()` is the latest committed world (`get_world_counter()`), matching `invokelatest` semantics so freshly-defined methods are visible on entry. The world is threaded through `FrameCode`, `Frame`, `prepare_call`, `prepare_framecode`, `get_call_framecode`, `optimize!`, `lookup_stmt`, `whichtt`, and `to_function`; every dispatch and global-lookup site now goes through `invoke_in_world(frame.world, ...)`.

Method frames hold their captured world. Toplevel code, however, always runs in the latest world: `step_expr!` refreshes `frame.world` to the current counter for toplevel frames so each statement sees the bindings, methods, and types defined by earlier statements, and `:latestworld` markers advance it after a world-incrementing statement on 1.12.

This ports work I did some time ago in a now-stale `teh/worldage` branch while also narrowing the scope. Much remains to be done for full world-age support, but this is a starting point.


src/interpret.jl
```diff
@@ -191,7 +191,7 @@ function evaluate_foreigncall(interp::Interpreter, frame::Frame, call_expr::Expr
             end...)
         end
     end
-    return Core.eval(moduleof(frame), Expr(head, args...))
+    return invoke_in_world(frame.world, Core.eval, moduleof(frame), Expr(head, args...))
```
simeonschaub: This doesn't do anything, since `Core.eval` will still execute in the latest world, right?

src/interpret.jl
```diff
@@ -226,7 +228,7 @@ function maybe_eval_with_scope(@nospecialize(f), fargs::Vector{Any}, frame::Fram
             newscope = Scope(newscope, scope.values...)
         end
         ex = Expr(:tryfinally, :($f($fargs...)), nothing, newscope)
-        return Some{Any}(Core.eval(moduleof(frame), ex))
+        return Some{Any}(invoke_in_world(frame.world, Core.eval, moduleof(frame), ex))
```
simeonschaub: Same here

simeonschaub (commented):
This looks like a great improvement! Just have a minor question regarding what to do about `Core.eval`

src/interpret.jl
timholy: Nice catch, thanks! And really I didn't even need to make this change since it's a ccall

src/interpret.jl
timholy: This one was more borked, I should have put the world inside the `tryfinally` expression.

timholy:
Test failures are known, see https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/716#pullrequestreview-4436082639

Though maybe we should do what we can to fix the PyCall issue before going too much further. This passes enough tests for me, but I'll let @simeonschaub be the judge.

simeonschaub:
> Though maybe we should do what we can to fix the PyCall issue before going too much further.

Do those affect just 32 bit Windows or 64 bit as well? Either way, I'd be fine with disabling these tests for now though

The nightly failures probably also need to be addressed at some point, but I don't think that needs to block this PR

simeonschaub (approved):

timholy:
For the PyCall thing I'm going to do the same thing as in #716. PR coming.

Issue #354, opened 2019-12-09
TypeError: in cglobal, expected Symbol, got JuliaInterpreter.SSAValue


fredrikekre:
The following code only interprets after running it in compiled mode once:

```julia
using HTTP, JuliaInterpreter
method = "GET"
url = "https://api.github.com/"
headers = Dict("User-Agent" => "Debugger.jl")
```

```julia
julia> @interpret HTTP.request(method, url, headers)
ERROR: TypeError: in cglobal, expected Symbol, got JuliaInterpreter.SSAValue
Stacktrace:
 [1] #evaluate_call_recurse!#48(::Bool, ::typeof(JuliaInterpreter.evaluate_call_recurse!), ::Any, ::Frame, ::Expr) at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/interpret.jl:213
 [2] evaluate_call_recurse! at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/interpret.jl:202 [inlined]
 [3] eval_rhs(::Any, ::Frame, ::Expr) at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/interpret.jl:379
 [4] step_expr!(::Any, ::Frame, ::Any, ::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/interpret.jl:519
 [5] step_expr!(::Any, ::Frame, ::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/interpret.jl:558
 [6] finish!(::Any, ::Frame, ::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/commands.jl:14
 [7] finish_and_return! at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/commands.jl:29 [inlined]
 [8] #evaluate_call_recurse!#48(::Bool, ::typeof(JuliaInterpreter.evaluate_call_recurse!), ::Any, ::Frame, ::Expr) at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/interpret.jl:240
 ... (the last 7 lines are repeated 7 more times)
 [58] evaluate_call_recurse! at /home/fredrik/.julia/packages/JuliaInterpreter/ATirK/src/interpret.jl:202 [inlined]
[...]

PR #716, opened 2026-06-05
Vendor issue #354 test; drop HTTP dependency

timholy:
The HTTP test made a live network request and asserted against `HTTP.Messages.Response`, a path that no longer resolves in HTTP 2.x. Its real purpose was issue #354: an `llvmcall` argument derived from a function argument cannot be interpreted directly and must be routed to compiled execution. `Base.load_state_acquire` was the original trigger but was removed from Base in 1.12. Reproduce the construct self-containedly so the check no longer depends on the network or on an external package, and works on all supported Julia versions.


timholy (commented):
Failures are pre-existing (win32 PyCall bug, nightly)

PR #720, opened 2026-06-10
Re-enable const folding with world-age invalidation

timholy:
`optimize!` folds references to `const` globals into the lowered code, but this was disabled on Julia 1.12+ (where a `const` may be redefined) because a cached `FrameCode` could then serve a stale folded value. Now that frames are gaining the ability to be invalidated, re-enable it.

Each folded binding's `Core.BindingPartition` is recorded in a new `FrameCode.world_deps` field. The partition is an in-place invalidation token: redefining the binding drops its `max_world` below the redefinition world. `framecode_valid_world` checks the frame's world against every recorded partition; `prepare_framecode`/`get_framecode` drop and rebuild a cached `FrameCode` that no longer holds for the requested world.

On 1.12+, folding is restricted in two ways:

- Only method scopes are folded, never module-scoped (toplevel) thunks. Module-scoped thunks only run once, so the performance isn't an issue; moreover, they need ways for definitions to update with world-age, making const-folding a liability rather than a benefit.

- Only *explicit* `const`s are folded, never implicit (`using`-imported) bindings: an implicit name may be locally shadowed -- for example `sin(::Int) = ...` in a module that uses `Base.sin`, where the `:method` introduces a new local `sin` -- so folding the imported value would bind the method to the wrong function. Pre-1.12 this role was served by the now-inert `isbindingresolved`.

It's worth clarifying this leaves one correctness gap: it's possible to set up situations where a toplevel frame that redefines a binding reaches the *old* binding within method frames. It requires a complex sequence that is unlikely to occur in practice (start debugging in a toplevel frame that calls a method, redefines a binding accessed by the method, and then calls the method again). Because I think debugging toplevel frames is rare (I might be the most likely person to do it, because of my work on Revise), I think this omission is defensible. But it will be closed in a follow-up PR, so let's avoid making a release until that lands.

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -371,15 +371,13 @@ end
 maybe_step_through_kwprep!(frame::Frame, istoplevel::Bool=false) =
     maybe_step_through_kwprep!(RecursiveInterpreter(), frame, istoplevel)
 
-@static if isbindingresolved_deprecated
-    function is_empty_namedtuple(stmt)
-        isexpr(stmt, :call) && length(stmt.args) == 1 || return false
-        arg1 = stmt.args[1]
-        isa(arg1, GlobalRef) || return false
-        return arg1.name === :NamedTuple
-    end
-else
-    is_empty_namedtuple(stmt) = isexpr(stmt, :call) && is_quoted_type(stmt.args[1], :NamedTuple) && length(stmt.args) == 1
+# `optimize!` may fold the `NamedTuple` `const` `GlobalRef` into a `QuoteNode` of the type, so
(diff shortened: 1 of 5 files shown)


timholy:
I've been doing some benchmarking and just can't find evidence that this makes an appreciable performance improvement (interpretation cost is dominated by per-statement and per-call overhead, not global lookup). It used to be important, but lookups have gotten a lot faster since the early days of this package when const-folding was added.

I'm therefore going to just close this, since I don't think we benefit from it and it just adds things we'd have to worry about.

PR #724, opened 2026-06-15
Add Aqua.jl quality checks

timholy:
Adds `Aqua.test_all` to the test suite. One real fix: add an explicit inner constructor to `_DispatchableMethod{FrameCode}` to suppress the auto-generated outer constructor, which has an unbound `FrameCode` when `next=nothing` and `frameinstance=Compiled()`.


KristofferC:
Seems like a good idea

timholy:
Yeah, I'm more sold on this one than the ExplicitImports one.

In case it's not obvious, I'm doing a housecleaning before making a new release. Complain if there's some reason to hold off.

PR #725, opened 2026-06-15
Add ExplicitImports.jl as a test dependency

timholy:
This might seem strange for a package that relies on so many internals; I'm fine with closing it without merging, but I thought it would be worth seeing what it looks like.

Fixes implicit imports (CodeTracking, InteractiveUtils, Random, UUIDs), removes four stale explicit imports (convert, isless, get_world_counter, LineInfoNode), removes self-qualified accesses (JuliaInterpreter.X inside the package), and corrects Core.Compiler.Argument to Core.Argument.

The two public-ness checks are suppressed: JuliaInterpreter intentionally uses non-public Core/Base/Compiler internals throughout. Four Core.Compiler.X qualified accesses (Val, getindex, iterate, specialize_method) refer to distinct objects on Julia 1.10 vs 1.12, so all_qualified_accesses_via_owners is suppressed rather than scattering @static VERSION guards. The #Internal submodule uses dynamic includes and is excluded via ignore=.


KristofferC:
:+1:

src/packagedef.jl
```diff
@@ -1,14 +1,14 @@
 using Base.Meta
-import Base: +, -, convert, isless, get_world_counter, mapany, ntupleany, invokelatest, invoke_in_world
-using Core: CodeInfo, SimpleVector, LineInfoNode, GotoNode, GotoIfNot, ReturnNode,
+import Base: +, -, mapany, ntupleany, invokelatest, invoke_in_world
```
simeonschaub:
```suggestion
using Base: +, -, mapany, ntupleany, invokelatest, invoke_in_world
```
at least if it doesn't break anything? If it does, we should make it explicit everywhere that we're overloading base functions

Issue #722, opened 2026-06-12
Remove the defensive `clear_caches()` from entry points


timholy:
`enter_call` and `enter_call_expr` begin with `clear_caches()`, which empties `framedict`/`genframedict` so that every `@interpret` re-lowers, re-optimizes, and rebuilds framecodes from scratch. This has been the blunt-instrument defense against stale caches since #259.

After #721 merges, we don't need it for correctness. The function should still be retained, but it doesn't need to be called automatically anymore.

The remaining design question is whether anything should bound `framedict` growth across redefinitions: Julia never deletes code, so perhaps this package should not either.

Note: benchmarking shows that there isn't much of a performance win to removing it. In many circumstances, building the frames is much cheaper than running them.

aviatesk:
I think we should just remove them, thinking it could obscure a potential lack of world-age handling.
Given the existence of consumers like Revise/JETLS, which externally use JuliaInterpreter subroutines and may not necessarily start interpretation through `enter_call`, it would be better to make those world-age issues easier to detect.

PR #728, opened 2026-06-15
Version 0.11.0

### LINKED ISSUE #617: wrong world age for everything
Looks like this code always uses the wrong `get_world_counter` function instead of the correct `tls_world_age` (which may need to be defined as a ccall for most older versions of julia)

### LINKED ISSUE #666: `@interpret(x == 2)` and `@interpret(x) == 2` lead to different execution world ages

timholy:
This is to prepare for releasing the big world age PRs: #715, #719, #721.

TODOs before merging:

- [x] Decide/merge #724
- [x] Decide/merge #725
- [x] Decide/merge #726
- [x] Decide/merge #727
- [x] Merge #729
- [x] Decide/fix #617
- [x] Fix #666
- [x] Decide/merge #733 ~Decide whether we want to do #722~
- [x] Merge #735
- [x] Tim successfully incorporates this upcoming breaking release into LoweredCodeUtils and Revise (locally on his machine) and convinces himself that it delivers as promised

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "JuliaInterpreter"
 uuid = "aa1ae85d-cabe-5617-a682-6adf51b2e16a"
-version = "0.10.12"
+version = "0.11.0"
 authors = ["Tim Holy <tim.holy@gmail.com>, Kristoffer Carlsson <kcarlsson89@gmail.com>, Shuhei Kadowaki <aviatesk@gmail.com> and contributors"]
 
 [deps]


timholy:
I've verified that I can exploit this to allow Revise to run its own code in a fixed world while interpreting user code in the current world. There will be at least one change needed to LoweredCodeUtils, but once I merge #735 then JI is, from my standpoint, ready for its next release.

PR #729, opened 2026-06-16
Run expanded invokelatest in the latest world

timholy:
When the recursive interpreter expands `Core.invokelatest`/`Core._call_latest`, it strips the wrapper and interprets the target itself. The world the wrapper implies ("latest") is not an argument, so interpreting the target at the enclosing `frame.world` silently dropped it: a binding, method, or type defined since the frame's world was then invisible.

Added `recurse_expanded_builtin_latest` to handle this correctly. The frame's original world is stashed, and then temporarily set to the latest before executing the call.

`invoke_in_world`/`_call_in_world` need no equivalent change: they are not expanded, so the explicit world argument is passed straight to the genuine builtin and honored end-to-end.

### DIFF
--- a/bin/generate_builtins.jl
+++ b/bin/generate_builtins.jl
@@ -54,7 +54,7 @@ const CALL_LATEST = """args = getargs(interp, args, frame)
         for x in args
             push!(new_expr.args, QuoteNode(x))
         end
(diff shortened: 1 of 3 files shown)


timholy (commented):
mac failure is #730

PR #731, opened 2026-06-16
Read freshly-included test functions in the latest world

timholy:
`somefunc` and `f_check` are defined by `include` inside a `mktemp() do ... end` closure and then referenced from that same already-compiled closure. Under Julia 1.12's strict binding world-age semantics this reads the global at a world before its definition, emitting a "binding accessed in a world prior to its definition world" warning (and erroring on future Julia versions). The interpreter is not involved: the `@interpret` macro splices the function value into the caller's code, so the offending read is the test closure's.

Bind a local from the latest world via `invokelatest` right after `include`, so the references below resolve the local rather than re-reading the global.

### DIFF
--- a/test/breakpoints.jl
+++ b/test/breakpoints.jl
@@ -246,6 +246,10 @@ mktemp() do path, io
     close(io)
     breakpoint(path, 3)
     include(path)
+    # `somefunc` is defined by `include` in a world later than this closure's; bind a local in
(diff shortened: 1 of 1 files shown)


timholy:
I strongly prefer #732, so I advise against merging this. But will hold in case there are review comments.

PR #732, opened 2026-06-16
Run interpreted frames in the caller's task world

### LINKED ISSUE #617: wrong world age for everything

timholy:
`default_world()` returned the latest committed world (`Base.get_world_counter()`), so every interpreted call behaved like `invokelatest`: methods and bindings newer than the calling code's own world were visible. Compiled code from the same point sees only its fixed task world and raises a world-age error for a method defined too recently; interpretation must match that.

Resolve and run frames in the caller's task world age (`tls_world_age`, defined via a ccall on Julia before 1.12) and thread that world through generated-function expansion. Toplevel/module frames still refresh to the latest world per statement, so earlier definitions remain visible.

`@interpret` gains a `world=w` option for the legitimate case of reaching a method newer than the caller's world — e.g. one just brought in by `include` within the same function; the breakpoint tests that do this now pass `world=Base.get_world_counter()`.

Fixes #617 Closes #731

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -143,11 +143,11 @@ end
 
 get_source(meth::Method) = Base.uncompressed_ast(meth)
 
-function get_source(g::GeneratedFunctionStub, source::Method, env)
+function get_source(g::GeneratedFunctionStub, source::Method, env, world::UInt)
     b = @static if VERSION < v"1.12.0-DEV.1968"   # julia #57230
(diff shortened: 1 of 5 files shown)


timholy:
@pfitzseb if we merge this, VSCode will likely have to supply the world age when invoking JI since it will now inherit from the invoking task.

Issue #220, opened 2019-03-26
Juno.@enter exits with error when running CSV.read()


zhopan77:
I have a simple script which runs fine in the Juno IDE:

using CSV
using DataFrames

function main()
    fn = "test.csv";
    df = CSV.read(fn);
    display(df);
end

main();

I start debugging with
Juno.@enter main()

Upon executing "df = CSV.read(fn);" the debugger exits with an error message:

> julia> Juno.@enter main()
debug> ERROR: unexpected ccall to %12
Stacktrace:
 [1] resolvefc(::JuliaInterpreter.Frame, ::Any) at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\interpret.jl:132
 [2] #collect_args#33(::Bool, ::Function, ::JuliaInterpreter.Frame, ::Expr) at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\interpret.jl:139
 [3] #collect_args at .\none:0 [inlined]
 [4] evaluate_foreigncall(::JuliaInterpreter.Frame, ::Expr) at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\interpret.jl:153
 [5] eval_rhs(::Any, ::JuliaInterpreter.Frame, ::Expr) at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\interpret.jl:366
 [6] step_expr!(::Any, ::JuliaInterpreter.Frame, ::Any, ::Bool) at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\interpret.jl:497
 [7] step_expr!(::Any, ::JuliaInterpreter.Frame, ::Bool) at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\interpret.jl:536
 [8] finish!(::Any, ::JuliaInterpreter.Frame, ::Bool) at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\commands.jl:14
 [9] finish_and_return! at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\commands.jl:29 [inlined]
 [10] #evaluate_call_recurse!#37(::Bool, ::Function, ::Any, ::JuliaInterpreter.Frame, ::Expr) at C:\Users\zpan\.julia\packages\JuliaInterpreter\8bMSs\src\interpret.jl:235
[...]

Issue #178, opened 2019-03-18
slot not assigned


fredrikekre:
Looks similar to #74:
```
julia> using JuliaInterpreter
       a = [1, 2, 3, 4]
       @breakpoint length(LinearIndices(a))
       frame, bp = @interpret sum(a)
       debug_command(frame, :c)
ERROR: slot _11 with name n not assigned
Stacktrace:
 [1] error(::String, ::Core.SlotNumber, ::String, ::Symbol, ::String) at ./error.jl:42
 [2] lookup_var(::Frame, ::Core.SlotNumber) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/interpret.jl:13
 [3] #collect_args#31(::Bool, ::Function, ::Frame, ::Expr) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/interpret.jl:64
 [4] collect_args(::Frame, ::Expr) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/interpret.jl:136
 [5] #evaluate_call_recurse!#35(::Bool, ::Function, ::Any, ::Frame, ::Expr) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/interpret.jl:204
 [6] evaluate_call_recurse! at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/interpret.jl:198 [inlined]
 [7] eval_rhs(::Any, ::Frame, ::Expr) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/interpret.jl:354
 [8] step_expr!(::Any, ::Frame, ::Any, ::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/interpret.jl:487
 [9] step_expr!(::Any, ::Frame, ::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/interpret.jl:526
 [10] finish!(::Any, ::Frame, ::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/commands.jl:14
 [11] finish_and_return!(::Any, ::Frame, ::Bool) at /home/fredrik/.julia/packages/JuliaInterpreter/RUp4U/src/commands.jl:29
[...]

PR #737, opened 2026-06-18
Support Julia 1.13 (syntactic ccall, print_stackframe)

### LINKED ISSUE #696: `memhash` is to be deleted in 1.13
it appears in some test here
https://github.com/JuliaDebug/JuliaInterpreter.jl/blob/b429971de7dabadf0b9f5d097d882d6edb803ca0/test/interpret.jl#L570

although not in any source.

ref: https://github.com/JuliaLang/julia/issues/59151

timholy:
Adapt to Base changes in Julia 1.13:

- `show_backtrace`: Base's `print_stackframe` gained cycle-tracking arguments (Julia #55841); call the new 8-argument form on 1.13+.
- `resolvefc`: under the 1.13 `:syntacticccall` lowering a foreigncall target is a literal `(name, lib)` tuple expression; pass it through unchanged so the eval'd `:foreigncall` receives the expected form.
- `build_compiled_foreigncall!`: the `:syntacticccall` branch keeps the target tuple symbolic, so record the library binding it resolves in `world_deps` and fold the resolved tuple value into the compiled-call cache key. Otherwise rebinding the library const reused a wrapper that baked the old library.

Tests:

- limits.jl: recalibrate `nstmts` for 1.13's per-iteration statement count.
- debug.jl (#178): the breakpoint target is now reached via local helpers called once, rather than `length`/`sum`; Base routes reductions through `length(::AbstractArray)` on 1.13, which hit a stale `length` breakpoint more than once and leaked it into later tests. Clear breakpoints in the #178 and kwprep testsets, and import InteractiveUtils directly.
- interpret.jl (#220): exercise the ccall-through-binding path with a `@cfunction` pointer instead of `Base.memhash`, removed in 1.13 (#696).

Fixes #696.

This gets tests passing on julia 1.13.0-rc1. There are a number of new binding-access warnings, but these are left for a future PR.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,4 +61,4 @@ jobs:
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v6
         with:
-          file: lcov.info
+          files: lcov.info
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -62,3 +62,4 @@ jobs:
       - uses: codecov/codecov-action@v6
         with:
           files: lcov.info
+          token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,10 @@ jobs:
           - version: '1' # current stable
             os: ubuntu-latest
             arch: x64
-          - version: '1.10' # lowest version supported
+          - version: 'min' # lowest version supported
+            os: ubuntu-latest
+            arch: x64
+          - version: 'pre' # upcoming release, or duplicate of '1'
             os: ubuntu-latest
             arch: x64
           - version: 'nightly' # dev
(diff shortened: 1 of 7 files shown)


KristofferC:
Would it be a good idea to add `pre` to the CI matrix so this gets tested?

timholy:
Just to check, 'pre' duplicates '1' when there is no upcoming release, right?

KristofferC:
Yes, at least that is my recollection

timholy:
Mine too. Also noticed the codecov warning,
```
Warning: Unexpected input(s) 'file', valid inputs are ['base_sha', 'binary', 'codecov_yml_path', 'commit_parent', 'directory', 'disable_file_fixes', 'disable_search', 'disable_safe_directory', 'disable_telem', 'dry_run', 'env_vars', 'exclude', 'fail_ci_if_error', 'files', 'flags', 'force', 'git_service', 'gcov_args', 'gcov_executable', 'gcov_ignore', 'gcov_include', 'handle_no_reports_found', 'job_code', 'name', 'network_filter', 'network_prefix', 'os', 'override_branch', 'override_build', 'override_build_url', 'override_commit', 'override_pr', 'plugins', 'recurse_submodules', 'report_code', 'report_type', 'root_dir', 'run_command', 'skip_validation', 'slug', 'swift_project', 'token', 'url', 'use_legacy_upload_endpoint', 'use_oidc', 'use_pypi', 'verbose', 'version', 'working-directory']
```
and it looks like we haven't uploaded new coverage data since I merged #713, so I fixed that too (hopefully).

I'll merge this once CI is as green as we can expect it (failures on nightly excepted).

PR #740, opened 2026-06-20
Resolve a nested `module` to a fresh local module

### LINKED ISSUE #747: Can't track modules with same name as existing packages
Making a module with the same name as an existing package throws an error.  For example, in "test.jl" I have 
```
module Test
using BenchmarkTools

function test()
    @btime rand()
end

end
```
Now at the REPL if I do
```
> using Revise
> includet("test.jl")
```
I get an error:
```
ERROR: ArgumentError: Package Test does not have BenchmarkTools in its dependencies:
- You may have a partially installed environment. Try `Pkg.instantiate()`
  to ensure all packages in the environment are installed.
- Or, if you have Test checked out for development and have
  added BenchmarkTools as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with Test
```
Doing the same thing without Revise, and just using `include("test.jl")` works without any problems.

This issue goes away if we rename the module Test to something else, like Test2.  I think this is happening because of a name collision with the unit testing package Test.jl.  I'd just generally expect `includet` to behave more like `include` in this instance, or at the very least for the `ERROR` message that gets thrown to be more informative as to the real source of the issue.

timholy:
`ExprSplitter` resolved any `module X` whose name matched an already-loaded package to that package, even when the declaration was nested in an ordinary module. A file `includet`ed into `Main` that defined `module Test` (or any other name shared with a loaded package) therefore had its body evaluated inside the loaded package, e.g. `using SomeDep` failed because the package was not a dependency of stdlib `Test`.

Reuse of a loaded package now happens only when the declaration is evaluated into `Base.__toplevel__` (a package loading itself) or when the name already refers to a genuine submodule of the parent (`parentmodule(found) === parent`, i.e. re-revising a module we created). Otherwise the module is built fresh, so a nested `module X` shadows any loaded package of the same name, matching `include`.

The shared logic for both the `ExprSplitter` trampoline and direct toplevel interpretation now lives in `resolve_module`.

Fixes timholy/Revise.jl#747

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -67,8 +67,9 @@ else
     String(first(methods(getfield(mod, :eval))).file)
 end
 
-# Resolve `module newname ... end` (with source expression `ex`), evaluated into
-# `parentmod`, to the `PkgId` of the module it refers to, or `nothing`.
+# Resolve a top-level `module newname ... end` (with source expression `ex`),
+# evaluated into `parentmod === Base.__toplevel__`, to the `PkgId` of the module it
+# refers to, or `nothing`.
 # `Base.identify_package` is the normal lookup; it returns `nothing` when
 # `newname` is not a declared dependency of the active project (an stdlib in a
 # bare test environment, or a package extension), in which case toplevel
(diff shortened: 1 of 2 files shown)


timholy (commented):
It's a valid question whether this is a bugfix or a breaking change. I lean towards bugfix. There are a couple of non-Revise consumers of ExprSplitter (Pluto being the most notable).

PR #742, opened 2026-06-27
compat: Handle upcoming envout markers

Keno:
JuliaLang/julia#62001 changes subtype envout entries so uncertain static-parameter bindings may be represented as marker SimpleVectors carrying the underlying TypeVar and whether the binding is definitely defined. Treat unconstrained markers as undefined in interpreted frames, and unwrap constrained singleton TypeVars to their concrete value when filling static parameters.

This keeps JuliaInterpreter compatible with the upcoming Base representation while preserving the existing behavior for plain TypeVars and concrete env entries.

This change was prepared with assistance from generative AI. The PR should be checked carefully.


Keno:
I've asked Claude to also fix the other 1.14 regressions. This does expose a 1.14 upstream bug, which I'll go fix, so that test is temporarily disabled on 1.14, but I'll go fix that and I think it's useful to merge this in the meantime.

src/interpret.jl
```diff
@@ -69,6 +69,18 @@ function lookup_expr(interp::Interpreter, frame::Frame, e::Expr)
...
+        # On Julia ≥ 1.14, `cglobal` name lookups lower to `Expr(:foreignglobal, spec)`
+        # (JuliaLang/julia#61709), where `spec` is a (possibly quoted) symbol, string,
+        # or `(name, lib)` tuple. Evaluate it the same way the runtime interpreter
+        # does: resolve the symbol and return the raw `Ptr{Cvoid}` (issue #734).
+        fptr = e.args[1]
+        isa(fptr, QuoteNode) && (fptr = fptr.value)
+        if isexpr(fptr, :tuple)
+            fptr = Core.tuple(Any[lookup_nested(interp, frame, arg) for arg in (fptr::Expr).args]...)
+        end
+        return ccall(:jl_cglobal_auto, Any, (Any,), fptr)
```
topolarity, line 82: This seems to have pulled the interpreter implementation from an old version of the PR

The landed version obeys the syntactic restrictions: https://github.com/JuliaLang/julia/blob/92be8bc0883d013ba96a10060be51354b3df0935/src/interpreter.c#L388-L420

PR #743, opened 2026-06-28
Fix extension module name resolution

timholy:
If two packages have extensions sharing a name, make sure we pick the right extension to load.


timholy:
oops didn't mean to request Copilot, fooled by the UI

PR #747, opened 2026-07-02
interpret: Match the runtime interpreter's syntactic dispatch for `Ex…

Keno:
…pr(:foreignglobal)`

The `Expr(:foreignglobal)` handler added in #742 was modeled on an earlier revision of the upstream lowering PR (JuliaLang/julia#61709); the version that landed restricts the foreign-symbol lookup to syntactic (possibly quoted) symbol, string, and `(name, lib)` tuple arguments, and treats anything else as a value that must evaluate to a pointer, returned as `Ptr{Cvoid}`.

The merged handler instead unwrapped any `QuoteNode` and funneled every argument straight to `jl_cglobal_auto`, so a runtime-valued argument (e.g. `cglobal(p)` with `p::Ptr`) passed the raw `SlotNumber`/`SSAValue` IR node to the runtime and threw a spurious `TypeError` instead of evaluating the operand. Mirror the landed dispatch from `eval_expr` in Julia's src/interpreter.c, including the pointer typecheck on the fallback path.

Addresses the post-merge review on #742.

### DIFF
--- a/src/interpret.jl
+++ b/src/interpret.jl
@@ -70,16 +70,29 @@ function lookup_expr(interp::Interpreter, frame::Frame, e::Expr)
     end
     head === :boundscheck && length(e.args) == 0 && return true
     if head === :foreignglobal
-        # On Julia ≥ 1.14, `cglobal` name lookups lower to `Expr(:foreignglobal, spec)`
-        # (JuliaLang/julia#61709), where `spec` is a (possibly quoted) symbol, string,
-        # or `(name, lib)` tuple. Evaluate it the same way the runtime interpreter
-        # does: resolve the symbol and return the raw `Ptr{Cvoid}` (issue #734).
+        # On Julia ≥ 1.14, `cglobal` lowers to `Expr(:foreignglobal, spec)`
+        # (JuliaLang/julia#61709). Mirror the runtime interpreter's syntactic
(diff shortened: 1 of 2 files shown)


src/interpret.jl
```diff
@@ -70,16 +70,29 @@ function lookup_expr(interp::Interpreter, frame::Frame, e::Expr)
...
-        isa(fptr, QuoteNode) && (fptr = fptr.value)
+        if isa(fptr, QuoteNode) && (isa(fptr.value, String) || isa(fptr.value, Tuple))
+            fptr = fptr.value
+        end
         if isexpr(fptr, :tuple)
-            fptr = Core.tuple(Any[lookup_nested(interp, frame, arg) for arg in (fptr::Expr).args]...)
+            # The `(name, lib)` elements may themselves be `getproperty` chains
+            # (e.g. `Base.Math.libm`), which `lookup_nested` resolves.
+            spec = Core.tuple(Any[lookup_nested(interp, frame, arg) for arg in (fptr::Expr).args]...)
+            return ccall(:jl_cglobal_auto, Any, (Any,), spec)
```
topolarity, line 87: `jl_lookup_foreignsymbol` would be slightly better, but that's my bad for not exporting it for you

(no behavioral difference)

topolarity (approved):
LGTM. Thanks for the fix-up @Keno

PR #750, opened 2026-07-17
AI bug fixing bonanza

KristofferC:
These were created by having Claude Fable and gpt try their best to find bugs in JuliaInterpreter.

Opening this so it can be discussed what to do with these.


KristofferC:
Superseded by the difficulty-tiered split: #751 (easy), #752 (intermediate), #753 (hard). Together they carry all fixes from this PR, each commit paired with its regression test.

PR #751, opened 2026-07-18
Bug fixes: straightforward defects (easy tier)

KristofferC:
First of three PRs splitting #750 by review difficulty. This tier contains the fixes where the defect is visible in the diff itself: wrong variable referenced, missing copy, wrong return shape, a throwing `==` on `missing`, missing nothing-guards — plus the documentation/metadata commits.

Each fix commit carries its own regression test in the logical test file.

- Fix UndefVarError when showing a breakpoint with a (Module, Expr) condition
- Accept `:macrocall` expressions with nothing line info in ExprSplitter
- Make an empty vararg visible to `locals()` and `eval_code`
- Fix undefined variable reference in `extract_args`' dead `:body` branch
- Return `(frame, pc)` from `:s` when a stepped-into call throws and is caught
- Pass `rootistoplevel` when retrying a command after exception unwinding
- Guard `until_line!` against frames without line info
- Make `Variable` equality total and add a consistent hash
- Stop `determine_method_for_expr` from mutating the caller's expression
- Return a copy from `breakpoints()` instead of the live registry
- Docstring corrections, package metadata, concurrency documentation note

Companion PRs: intermediate and hard tiers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/breakpoints.jl
+++ b/src/breakpoints.jl
@@ -6,8 +6,12 @@ const _breakpoints = AbstractBreakpoint[]
     breakpoints()::Vector{AbstractBreakpoint}
 
 Return an array with all breakpoints.
+
+The returned array is a copy of the internal breakpoint registry; mutating it does
+not affect the set of breakpoints. Use [`remove`](@ref), [`enable`](@ref), and
+[`disable`](@ref) to update the breakpoints themselves.
 """
-breakpoints() = _breakpoints
+breakpoints() = copy(_breakpoints)
 
 
 const breakpoint_update_hooks = []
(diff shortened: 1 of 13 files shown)


src/breakpoints.jl
```diff
@@ -7,7 +7,8 @@ const _breakpoints = AbstractBreakpoint[]
 
 Return an array with all breakpoints.
 """
-breakpoints() = _breakpoints
+breakpoints() = copy(_breakpoints)  # a copy: mutating the registry must go through the API so
+                                    # that frame state stays in sync and update hooks fire
```
pfitzseb: Worth a mention in the docstring imo.

pfitzseb (commented):
LGTM.

PR #753, opened 2026-07-18
Bug fixes: subtle semantic defects (hard tier)

KristofferC:
Third of three PRs splitting #750 by review difficulty. These are the subtle state-semantics bugs: the interpreter's model of the active-exception stack and of world age diverging from native execution.

Each fix commit carries its own regression test in the logical test file.

**Exception-state semantics**
- Make `rethrow()` find the exception being handled in caller frames
- Restore the dynamic-scope stack when an exception unwinds (ScopedValue leakage)
- Model the active-exception stack and implement `:pop_exception`
- Pop the exception handler in debugger unwinding on Julia 1.11+

**World age**
- Pin the `applicable` builtin to the frame's world age
- Run the `@interpret` compiled fallback in the caller's world, not via `eval`
- Select `Core.invoke` methods and run compiled invoke fallbacks in the frame's world

Note: the `Core.invoke` world-age commit here is adapted relative to its #750 counterpart — on this branch it applies to the pre-existing tuple-type `invoke` path, since the `invoke(f, ::Method/::CodeInstance)` form support lives in the intermediate-tier PR. The fix logic (world-aware `whichtt` selection, `invoke_in_world` fallbacks) is the same.

Companion PRs: easy and intermediate tiers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bin/generate_builtins.jl
+++ b/bin/generate_builtins.jl
@@ -40,9 +40,8 @@ const RECENTLY_REMOVED = GlobalRef.(Ref(Core), [
     :_apply_pure, :_call_in_world, :_call_latest, :_defaultctors,
 ])
 const kwinvoke = Core.kwfunc(Core.invoke)
-# Builtins whose result depends on the world age (binding resolution): they must run in
-# `frame.world` rather than the interpreter's ambient world. Every entry here takes a `Module`
-# as its first argument, so the dependence is unconditional.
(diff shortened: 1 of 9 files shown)


timholy:
Those look good, though I haven't thought the exception-stack parts through carefully. I assume that each was verified to be a bug with a MWE before making the change?

PR #758, opened 2026-07-20
Preserve module self-bindings during resolution

aviatesk:
Module contexts bind their own name (e.g. `Base.Base === Base`) even though their parent module differs. The nested-module resolution introduced in #740 treated these bindings as unrelated loaded modules: re-interpreting `module X ... end` with `X` itself as the parent context - the core Revise/JET revision pattern - not only returned a fresh child module but also clobbered `X`'s self-binding when the child was evaluated into `X`.

Reuse the existing module when the found binding is the parent context itself, while preserving #740's fresh-module shadowing for unrelated same-name packages (Revise issue timholy/Revise.jl#747).

This knowingly diverges from the `include` semantics for a genuinely nested `module A` inside `module A`, where plain evaluation creates a fresh child `A.A`: that input is syntactically indistinguishable from re-interpretation of `A`'s own definition, so I believe re-entering the parent (the pre-#740 behavior) is the right default for this machinery.

Add a synthetic self-binding regression test.

### DIFF
--- a/src/construct.jl
+++ b/src/construct.jl
@@ -129,12 +129,18 @@ function find_or_create_module(parentmod::Module, ex::Expr)
     if invokelatest(isdefinedglobal, parentmod, newname)
         found = invokelatest(getglobal, parentmod, newname)
         found isa Module || throw(ErrorException("invalid redefinition of constant $(newname)"))
-        # Reuse a loaded package only when it loads itself (`parentmod === Base.__toplevel__`)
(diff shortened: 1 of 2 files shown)


timholy (approved):
Sorry for the breakage. If you've checked this doesn't break Revise's tests (doesn't seem like it should), no concerns whatsoever.

aviatesk:
This is a very subtle issue. Without the JET tests, no one might have noticed it :)
I confirmed that the Revise tests also pass without issue on this JI branch.

PR #759, opened 2026-07-20
reenable global ref lookup optimization

KristofferC:
Now, with worlds propagating, we can invalidate frames. This gives the following changes to the benchmarks (some extra benchmarks are added here):

| benchmark | base | new | ratio |
|---|---:|---:|---:|
| nonconst global read 10_000 | 102.5 ms | 73.3 ms | 0.71x |
| const global read 10_000 | 99.0 ms | 70.9 ms | 0.72x |
| tight loop 10_000 | 97.0 ms | 75.4 ms | 0.78x |
| ccall ptr | 3.18 μs | 2.55 μs | 0.80x |
| global callees 1_000 | 111.7 ms | 89.2 ms | 0.80x |
| ccall library | 5.26 μs | 4.23 μs | 0.81x |
| long function 5_000 | 2.73 ms | 2.30 ms | 0.84x |
| recursive self 1_000 | 2.27 ms | 1.93 ms | 0.85x |
| recursive other 1_000 | 468 μs | 399 μs | 0.85x |
| throw long 1_000 | 36.6 ms | 35.7 ms | 0.98x |
| frame creation cached | 479 ns | 474 ns | 0.99x |
| frame creation uncached | 4.54 μs | 5.71 μs | 1.26x |

Bindings that come from imports are not optimized because the invalidation story for these is kind of messy (see the diff). There is also an existing bug w.r.t in compiled code for ccalls, again see diff and the broken test.

Done with help of Claude Fable.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -66,3 +66,43 @@ function powf(a, b)
     ccall(("powf", Base.Math.libm), Float32, (Float32,Float32), a, b)
 end
 SUITE["ccall library"] = @benchmarkable @interpret powf(2, 3)
+
+# Global reads in a tight loop: `const` globals can be folded at framecode construction,
+# non-`const` globals must be looked up on every execution
+module BenchGlobals
+const CONSTGLOBAL = 1
+NONCONST = 1
+end
+
+function readconst(n)
+    s = 0
+    for i in 1:n
+        s += BenchGlobals.CONSTGLOBAL
+    end
+    return s
+end
+SUITE["const global read 10_000"] = @benchmarkable @interpret readconst(10_000)
+
+function readnonconst(n)
+    s = 0
+    for i in 1:n
+        s += BenchGlobals.NONCONST
+    end
+    return s
+end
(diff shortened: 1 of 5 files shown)


KristofferC:
```
Internal error: during type inference of
var"#SinTable#314"(Core.TypeEgal{Base.IdDict{K, V} where V where K})
Encountered unexpected error in runtime:
TypeError(func=:typeassert, context="", expected=Union{Nothing, Array{Any, 1}, Core.SimpleVector}, got=Array{Core.MethodInstance, 1}(dims=(1,), mem=Memory{Core.MethodInstance}(1, 0x7f339051f320)[MethodInstance (::Memory{Any})(UndefInitializer, Int64) from (::GenericMemory{kind, T, addrspace})(UndefInitializer, Int64) where {T, addrspace, kind}]))
ijl_type_error_rt at /cache/build/builder-amdci4-3/julialang/julia-ci/src/rtutils.c:121:5
ijl_type_error at /cache/build/builder-amdci4-3/julialang/julia-ci/src/rtutils.c:140:5
compute_edges! at ./../usr/share/julia/Compiler/src/typeinfer.jl:826:0 (pc: 55)
```

exciting

PR #763, opened 2026-08-15
Use JuliaLowering for source-faithful stepping

timholy:
The aim here is to use the improved provenance information of JuliaLowering to allow `next`, `step`, etc, follow the actual source. Ultimate goal includes eliminating the painful kwarg-pattern-matcher code.

### DIFF
--- a/src/commands.jl
+++ b/src/commands.jl
@@ -320,6 +320,9 @@ function is_indexed_iterate_call(@nospecialize(stmt))
     return f === Base.indexed_iterate
 end
 
+is_destructured_arg_slotname(name::Symbol) =
+    name === Symbol("") || startswith(String(name), "destructured#")
+
 """
     frame = maybe_step_through_arg_destructuring!(interp::Interpreter, frame::Frame)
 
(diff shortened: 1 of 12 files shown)


src/builtins.jl
```diff
@@ -127,6 +127,8 @@ function maybe_evaluate_builtin(interp::Interpreter, frame::Frame, call_expr::Ex
         return Some{Any}(Core._expr(getargs(interp, args, frame)...))
     elseif @static isdefinedglobal(Core, :_import) && f === Core._import
         return Some{Any}(Core._import(getargs(interp, args, frame)...))
+    elseif f === Core._new_cancel_source
+        return Some{Any}(Core._new_cancel_source(getargs(interp, args, frame)...))
```
simeonschaub: Can we split these out in another PR

src/commands.jl
```diff
@@ -419,15 +422,21 @@ function maybe_step_through_kwprep!(interp::Interpreter, frame::Frame, istopleve
...
     stmt = pc_expr(frame, pc)
-    if isbindingresolved_deprecated && !isa(stmt, Tuple{Symbol,Vararg{Symbol}}) && !is_empty_namedtuple(stmt) && n >= pc+1
-        nextstmt = pc_expr(frame, pc + 1)
-        if isa(nextstmt, Tuple{Symbol,Vararg{Symbol}}) || is_empty_namedtuple(nextstmt)
-            pc += 1
-            stmt = nextstmt
+    if !isa(stmt, Tuple{Symbol,Vararg{Symbol}}) && !is_empty_namedtuple(stmt) && !is_merge_call(stmt)
+        # The NamedTuple-keys tuple may be preceded by side-effect-free loads (binding
+        # resolution, or the `QuoteNode`s of a qualified callee such as `Base.sort`);
+        # scan past them, but stop at anything that executes user code.
```
simeonschaub: Do we need to be careful around resolving GlobalRefs, now that we have properly specified binding semantics? But that might be a bigger discussion

src/commands.jl
```diff
@@ -320,6 +320,9 @@ function is_indexed_iterate_call(@nospecialize(stmt))
     return f === Base.indexed_iterate
 end
 
+is_destructured_arg_slotname(name::Symbol) =
```
mlechu, line 323: Not now, but should eventually do this with provenance instead

mlechu:
discussed IRL: We might want to step through the AST instead, since the heuristics for "next" might be easier than figuring out what IR to skip.

PR #764, opened 2026-08-15
Regenerate builtins.jl for Julia 1.14-DEV

timholy:
Adds dispatch for the new task/cancellation builtins (Core._task, Core._new_cancel_source, Core.cancellation_point!, Core.task_result_type) and Core.bitsizeof. 

Assisted-by: Claude Fable 5 <noreply@anthropic.com>


timholy:
This looks like a nightly bug, not a problem with this PR.

